PR #39, opened 2022-08-03
Package Declarations1

knewbury01:
## Description

Package Declarations1

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - DCL37-C
     - DCL31-C
     - RULE-21-2
     - RULE-5-4
     - RULE-5-1
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

  - [ ] Confirmed

### Author

### Reviewer 


knewbury01:
# Notes DCL37-C

I did not implement the future library directions restrictions for this rule, and described that in the implementation notes, however if it seems like these would be crucial, I can add those

knewbury01:
# Notes DCL31-C

currently not supporting "check for implicit function declarations" portion of the rule, only supporting "Omission of type specifiers" portion

I need a reminder for what to do if our test setup doesnt compile a certain test, but other compilers do
the following:
```
int main(void) {
  for (int i = 0; i < 100; ++i) {
    /* int malloc() assumed */
    char *ptr = (char *)malloc(0x10000000);
    *ptr = 'a';
  }
  return 0;
}
```
compiles with: `gcc -c test.c`
with 
`gcc-11 (Homebrew GCC 11.3.0) 11.3.0`

but ` codeql test run --show-extractor-output --search-path .  c/cert/test/rules/DCL31C/DeclareIdentifiersBeforeUsingThem.qlref ` fails (for clang)

knewbury01:
# RULE-5-1 notes

some perf issue exists for this one currently, related to joining on names I think, but was hoping that external identifiers subclass was a small enough set to avoid this

c/misra/src/rules/RULE-5-4/MacroIdentifiersNotDistinct.ql
```diff
@@ -0,0 +1,29 @@
...
+import codingstandards.c.misra
+
+from Macro m, Macro m2
+where
+  not isExcluded(m, Declarations1Package::macroIdentifiersNotDistinctQuery()) and
+  not m = m2 and
+  (
+    if m.getName().length() >= 64
+    then m.getName().substring(0, 62) = m2.getName().substring(0, 62)
+    else m.getName() = m2.getName()
```
lcartey, line 25: Can you document why this is 63 characters (e.g. due to C99), and note that we don't currently support C90's requirements here?

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
```diff
@@ -0,0 +1,39 @@
...
+  not d = d2 and
+  d.getLocation().getStartLine() >= d2.getLocation().getStartLine()
+}
+
+from ExternalIdentifiers d, ExternalIdentifiers d2
+where
+  not isExcluded(d, Declarations1Package::externalIdentifiersNotDistinctQuery()) and
+  notSame(d, d2) and
+  if d instanceof ExternalIdentifiersLong and d2 instanceof ExternalIdentifiersLong
+  then d.getName().substring(0, 30) = d2.getName().substring(0, 30)
```
lcartey, line 37: Can you document why this is 31 characters (e.g. due to C99), and note that we don't currently support C90's requirements here?

lcartey (changes requested):
I made a first pass review, and added a few comments. I will take a further look tomorrow morning, but it's already looking good. 👍

c/cert/src/rules/DCL31-C/DeclareIdentifiersBeforeUsingThem.ql
```diff
@@ -0,0 +1,21 @@
...
+ *       external/cert/obligation/rule
+ */
+
+import cpp
+import codingstandards.c.cert
+
+from Declaration d
+where
+  not isExcluded(d, Declarations1Package::declareIdentifiersBeforeUsingThemQuery()) and
+  exists(Specifier s | s = d.getASpecifier() and s.toString().regexpMatch("implicit_.*"))
```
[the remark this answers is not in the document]
knewbury01: oh, there are actually no variations, good catch, the default is implicit int, so that is all we need to match!

c/misra/src/rules/RULE-5-4/MacroIdentifiersNotDistinct.ql
knewbury01: is currently noted in [pkg json implementation note](https://github.com/knewbury01/codeql-coding-standards/blob/knewbury01/Declarations1/rule_packages/c/Declarations1.json#L120)s, should it be in both places? can inline in query also but dont want to be redundant

cpp/common/src/codingstandards/cpp/rules/declaredareservedidentifier/DeclaredAReservedIdentifier.qll
```diff
@@ -0,0 +1,37 @@
...
+    (
+      m.(Macro).hasName(name) or
+      m.(Declaration).hasGlobalName(name)
+    ) and
+    (
+      Naming::Cpp14::hasStandardLibraryMacroName(name)
+      or
+      Naming::Cpp14::hasStandardLibraryObjectName(name)
+      or
+      Naming::Cpp14::hasStandardLibraryFunctionName(name)
```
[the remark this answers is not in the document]
knewbury01: possibly, especially if we are only supporting C99 (or as priority) compared to C11 even,

however there are [some](https://github.com/github/codeql-coding-standards/blob/main/c/misra/src/rules/RULE-21-1/DefineAndUndefUsedOnReservedIdentifierOrMacroName.ql#L31) other [queries](https://github.com/github/codeql-coding-standards/blob/main/c/cert/src/rules/STR32-C/NonNullTerminatedToFunctionThatExpectsAString.ql#L32) where we are also doing this (I know, not the best argument to continue down that road) - I suggest if we want to change this that the effort is orthogonal to this PR, as it will affect many queries, and requires a different dedicated effort to creating a parallel unique lib that contains name sets for alternative standards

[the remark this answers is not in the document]
knewbury01: especially since currently I would not know where to scrape these complete sets from, I know we had suggested [this](https://github.com/p12tic/cppreference-doc/blob/master/index-functions-c.xml) previously, but... what standard version is that?

c/misra/src/rules/RULE-5-4/MacroIdentifiersNotDistinct.ql
lcartey: The implementation notes are for developers reviewing results, whereas comments in the query are for ql writers. So I think it would be worth putting this information here, so it's easy to see when you read the query.

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
knewbury01: done!

c/misra/src/rules/RULE-5-4/MacroIdentifiersNotDistinct.ql
knewbury01: done!

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
```diff
@@ -0,0 +1,42 @@
...
+from ExternalIdentifiers d, ExternalIdentifiers d2
+where
+  not isExcluded(d, Declarations1Package::externalIdentifiersNotDistinctQuery()) and
+  notSame(d, d2) and
+  //C99 states the first 31 characters of external identifiers are significant
+  //C90 states the first 6 characters of external identifiers are significant and case is not required to be significant
+  //C90 is not currently considered by this rule
+  if d instanceof ExternalIdentifiersLong and d2 instanceof ExternalIdentifiersLong
+  then d.getName().substring(0, 30) = d2.getName().substring(0, 30)
+  else d.getName() = d2.getName()
```
lcartey, line 41: A general problem with this type of duplicated identifier rule is that there are some circumstances where we get multiple `Declaration`s appear in the database for effectively the same conceptual declaration. Examples include:

 * Compiling the same file twice under different contexts, such as different platform targets (which can result in a different `Declaration` per context).
 * Creating a dynamically linked library and indexing that in the same database as a client of that library (which can result in a `Declaration` each for the library and the client). Note: we typically don't have this issue with statically linked libraries, because we "extract" linker calls to de-duplicate client/library declarations.

I don't think there's a perfect solution to this problem, but the ones I think are most plausible are:

 1. Exclude cases where the names are identical to focus on those cases where the 31 character limit causes confusion. Pros: applicable to both functions and variables, limited false positives. Cons: will not report genuine ODR violation cases.
 2. Restrict the scope to a single link target. i.e. only report issues where we are sure that the two definitions were linked into the same binary. Pros: should eliminate false positives. Cons: we only have this information for `Function` declarations, can exclude genuinely interesting results that occur between translation units which are not linked together.
 3. Restrict the query to only report declarations where `hasDefinition()` is `true`. Pros: will avoid the common case related to dynamic libraries. Cons: false positives still possible in the multiple-compilations-of-the-same-file case.

@knewbury01 I think 1. is probably the best approach for now, but if we adopt that approach we should file an issue to look for ODR violations under Rule 5.1 at a later point.

c/misra/src/rules/RULE-5-4/MacroIdentifiersNotDistinct.ql
```diff
@@ -0,0 +1,33 @@
...
+    //C90 states the first 31 characters of macro identifiers are significant and is not currently considered by this rule
+    //ie an identifier differing on the 32nd character would be indistinct for C90 but distinct for C99
+    //and is currently not reported by this rule
+    if m.getName().length() >= 64
+    then m.getName().substring(0, 62) = m2.getName().substring(0, 62)
+    else m.getName() = m2.getName()
+  ) and
+  //reduce double report since both macros are in alert, arbitrary ordering
+  m.getLocation().getStartLine() >= m2.getLocation().getStartLine()
+select m, "Nondistinct macro identifer used " + m.getName() + " compared to $@.", m2, m2.getName()
```
lcartey, line 33: We shouldn't report macros with ifdef(n) guards around them, e.g.
```c
  #ifndef MY_MACRO
    #define MY_MACRO 1
  #endif
```

c/misra/src/rules/RULE-5-4/MacroIdentifiersNotDistinct.ql
```diff
@@ -0,0 +1,33 @@
...
+where
+  not isExcluded(m, Declarations1Package::macroIdentifiersNotDistinctQuery()) and
+  not m = m2 and
+  (
+    //C99 states the first 63 characters of macro identifiers are significant
+    //C90 states the first 31 characters of macro identifiers are significant and is not currently considered by this rule
+    //ie an identifier differing on the 32nd character would be indistinct for C90 but distinct for C99
+    //and is currently not reported by this rule
+    if m.getName().length() >= 64
+    then m.getName().substring(0, 62) = m2.getName().substring(0, 62)
```
lcartey, line 28: 1. I think you want 63 not 62, as the end is an exclusive offset.
 2. You can use `prefix(63)` instead of `substring`.

cpp/common/src/codingstandards/cpp/rules/declaredareservedidentifier/DeclaredAReservedIdentifier.qll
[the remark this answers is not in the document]
lcartey: Ok, let's add an issue to this repository to investigate using a different data source for the C standard library names.

lcartey (changes requested):
Review completed. Good job overall. Only major feedback was on Rule 5.1, which I think needs some refining to handle performance issues, mixed C/C++ code and edge cases around multiple declarations.

c/cert/src/rules/DCL31-C/DeclareIdentifiersBeforeUsingThem.ql
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+import cpp
+import codingstandards.c.cert
+
+from Declaration d
+where
+  not isExcluded(d, Declarations1Package::declareIdentifiersBeforeUsingThemQuery()) and
+  exists(Specifier s | s = d.getASpecifier() and s.toString() = "implicit_int") and
+  exists(d.getFile().getRelativePath())
```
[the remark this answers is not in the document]
knewbury01: "This line isn't necessary because (a) when running the query on real databases the interpretation phase will exclude results without a relative path and (b) when running in tests, the isExcluded predicate will exclude results without a relative path." 

I thought so too! but for openpilot and rule DCL37-C that predicate makes a difference in the results, without it it gets results in files like: `usr/include/x86_64-linux-gnu/bitss/stdlib-bsearch.h` that I thought it would make more sense to omit than report.

similarly, if I add `#include "stdio.h"` to the test file for that rule and run `codeql test run --show-extractor-output --search-path .  c/common/test/rules/declaredareservedidentifier/DeclaredAReservedIdentifier.ql` I start to get some results in that file `+| ../../includes/standard-library/stdio.h:184:5:184:13 | vasprintf | Reserved identifier 'vasprintf' is declared. |` ... is something in my setup broken?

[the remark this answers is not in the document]
knewbury01: oh, I realized that it is possible to run both 1) `c/cert/src/rules/DCL37-C/DoNotDeclareOrDefineAReservedIdentifier.ql` and 2) `c/common/test/rules/declaredareservedidentifier/DeclaredAReservedIdentifier.ql` in the UI on openpilot, but 2 does not contain the metadata and therefore does not do the filter.  this is what I based including the predicate off of. I guess the `codeql test` is affected in a similar way, still not quite sure what that should actually point to then for a shared query but oh well.

Predicate reverted!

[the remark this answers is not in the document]
knewbury01: also, extra type testcases added and extra query logic also added!

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
```diff
@@ -0,0 +1,42 @@
...
+select d, "External identifer is nondistinct " + d.getName() + " compared to $@.", d2, d.getName()
```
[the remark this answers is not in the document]
knewbury01: the `getName` on either should be same in this case... but good catch! 😅 

agree on the message improvement! change made!

knewbury01:
"Ok, let's add an issue to this repository to investigate using a different data source for the C standard library names."

[done](https://github.com/github/codeql-coding-standards/issues/50)

c/misra/src/rules/RULE-5-4/MacroIdentifiersNotDistinct.ql
knewbury01: another good catch, done in both this and 5-1!

c/misra/src/rules/RULE-5-4/MacroIdentifiersNotDistinct.ql
knewbury01: can you suggest a testcase that makes a false positive from this case? I think since the DB only contains Macros that actually make it into source, the immediate case I came up with does not FP:

testcase:
```
#ifndef CONDITION_MACRO
#define MY_MACRO 1 // COMPLIANT
#else
#define MY_MACRO 2 // COMPLIANT
#endif

#ifndef MY_MACRO
#define MY_MACRO 1 // COMPLIANT
#endif
```
query:
```
from Macro m
select m , m.getName()
```
result:
```
+| test.c:12:1:12:18 | #define MY_MACRO 1 | MY_MACRO |
```
if MY_MACRO only ends up being defined once due to conditional compilation, there is no FP. if its a redefinition, well that is reported but I would say we arent really looking to guess about purposeful redefinition... like:
```
#define MY_MACRO 2 // COMPLIANT

#ifdef MY_MACRO
#define MY_MACRO 1 // NON_COMPLIANT
#endif
```
 unless that is the case you are suggesting to omit?

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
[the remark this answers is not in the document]
knewbury01: change made in both RULE-5-1 and RULE-5-4

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
```diff
@@ -0,0 +1,42 @@
...
+
+predicate notSame(ExternalIdentifiers d, ExternalIdentifiers d2) {
+  not d = d2 and
+  d.getLocation().getStartLine() >= d2.getLocation().getStartLine()
+}
+
+from ExternalIdentifiers d, ExternalIdentifiers d2
+where
+  not isExcluded(d, Declarations1Package::externalIdentifiersNotDistinctQuery()) and
+  notSame(d, d2) and
```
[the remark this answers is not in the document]
knewbury01: hmmmmmm isolating `not d = d2 and
  d.getLocation().getStartLine() >= d2.getLocation().getStartLine()` into a predicate was the only way that I was able to get this query to complete on openpilot on my local setup (4threads allocated 14gb ram).... (waiting >5mins before terminating)

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
```diff
@@ -0,0 +1,42 @@
...
+  //C99 states the first 31 characters of external identifiers are significant
+  //C90 states the first 6 characters of external identifiers are significant and case is not required to be significant
+  //C90 is not currently considered by this rule
+  if d instanceof ExternalIdentifiersLong and d2 instanceof ExternalIdentifiersLong
```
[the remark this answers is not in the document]
knewbury01: done!

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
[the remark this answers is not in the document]
knewbury01: actually now this completes in 187 seconds on my setup, after the other refactor to use a `getSignificantName() ` predicate, so , better, not ideal but improved certainly

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
```diff
@@ -0,0 +1,42 @@
...
+ *       readability
+ *       external/misra/obligation/required
+ */
+
+import cpp
+import codingstandards.c.misra
+import codingstandards.cpp.Linkage
+
+class ExternalIdentifiers extends Declaration {
+  ExternalIdentifiers() { hasExternalLinkage(this) }
```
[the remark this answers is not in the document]
knewbury01: agree, done!

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
knewbury01: for my own learning:
is "Compiling the same file twice under different contexts, such as different platform targets (which can result in a different Declaration per context)." realistic in the same compilation execution? or does this mean conditional compilation, where only some src is seen per compilation? if it is the latter, again, similar to the Macros, does the DB actually have these same conceptual declarations in it?

knewbury01: I have gone with solution 1 for now and opened issue to enhance in future

c/cert/src/rules/DCL31-C/DeclareIdentifiersBeforeUsingThem.ql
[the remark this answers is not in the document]
knewbury01: @lcartey hm actually, [this is some error](https://github.com/github/codeql-coding-standards/runs/7844753286?check_suite_focus=true) from not using the `exists(d.getFile().getRelativePath())` check I think? or at least I thought so previously, if not, I am not quite sure where that comes from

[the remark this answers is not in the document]
lcartey: I think the problem in the test case is a bug in the generation of test cases for shared queries.

If we look in the shared library (`DeclaredAReservedIdentifier.qll`) for this query we can see the following parts related to exclusions:
```
abstract class DeclaredAReservedIdentifierSharedQuery extends Query { }

Query getQuery() { result instanceof DeclaredAReservedIdentifierSharedQuery }

query predicate problems(Element m, string message) {
  not isExcluded(m, getQuery()) and
 ....
```
The abstract class here allows any rule query that imports this library to specify the query indirectly. For example, Rule 21.2 has the following:
```
class DoNotDeclareAReservedIdentifierQuery extends DeclaredAReservedIdentifierSharedQuery {
  DoNotDeclareAReservedIdentifierQuery() {
    this = Declarations1Package::doNotDeclareAReservedIdentifierQuery()
  }
}
```

However, the test case we generate for shared rules do not specify any query id:
https://github.com/github/codeql-coding-standards/blob/c05f9a8d30935fbc51ac5e01667d745bd131a618/c/common/test/rules/declaredareservedidentifier/DeclaredAReservedIdentifier.ql#L1-L2

What this means is `isExcluded` will not hold for any source file/query pair, and will therefore not filter results outside the source location.

I'm not certain what the right fix is here, but it's clearly out of the scope of this PR. For now, I think it would be fine to just keep the standard library results in the result set.

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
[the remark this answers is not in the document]
lcartey: You mentioned this was reduced to a reasonable time by other fixes in this PR, so resolving.

c/cert/src/rules/DCL31-C/DeclareIdentifiersBeforeUsingThem.ql
```diff
@@ -0,0 +1,28 @@
...
+  not isExcluded(d, Declarations1Package::declareIdentifiersBeforeUsingThemQuery()) and
+  d.hasSpecifier("implicit_int") and
+  exists(Type t |
+    (d.(Variable).getType() = t or d.(Function).getType() = t) and
+    // Exclude "short" or "long", as opposed to "short int" or "long int".
+    t instanceof IntType and
+    // Exclude "signed" or "unsigned", as opposed to "signed int" or "unsigned int".
+    not exists(IntegralType it | it = t | it.isExplicitlySigned() or it.isExplicitlyUnsigned())
+  )
+select d, "Declaration is missing a type specifier."
```
lcartey, line 28:
```suggestion
select d, "Declaration " + d.getName() + " is missing a type specifier."
```

rule_packages/c/Declarations1.json
```diff
@@ -0,0 +1,140 @@
...
+          "precision": "very-high",
+          "severity": "warning",
+          "short_name": "ExternalIdentifiersNotDistinct",
+          "tags": [
+            "correctness",
+            "maintainability",
+            "readability"
+          ],
+          "implementation_scope": {
+            "description": "This query checks the first 31 characters of identifiers as significant, as per C99. This query does not consider universal or extended source characters.",
```
lcartey, line 95: I think we should mention here that we do not find contraventions where the names are identical.

rule_packages/c/Declarations1.json
```diff
@@ -0,0 +1,140 @@
...
+            ],
+              "implementation_scope": {
+                "description": "This query checks the first 63 characters of macro identifiers as significant, as per C99. Distinctness of parameters within the same function like macro are checked by compiler and therefore not checked by this rule.",
+                "items": []
+              }
+          },
+          {
+              "description": "Macros with the same name as their parameters are less readable.",
+              "kind": "problem",
+              "name": "Macro identifiers shall be distinct",
```
lcartey, line 127: Mention parameters in the name.

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
lcartey: > "Compiling the same file twice under different contexts, such as different platform targets (which can result in a different Declaration per context)." realistic in the same compilation execution?

This can happen in a single database if the user runs two consecutive compilations on the same machine. For example, a 32 bit build followed by a 64 bit build. In this case I believe we will get a different `Declaration` for each context (for example, a 32-bit version of a function and a 64-bit version of the function).

c/misra/src/rules/RULE-5-4/MacroIdentifiersNotDistinct.ql
lcartey: Yes, I think you're right - it doesn't seem possible to get a false positive with an if(n)def. Thanks for checking!

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
```diff
@@ -0,0 +1,51 @@
...
+ *       external/misra/obligation/required
+ */
+
+import cpp
+import codingstandards.c.misra
+import codingstandards.cpp.Linkage
+
+class ExternalIdentifiers extends Declaration {
+  ExternalIdentifiers() {
+    this.getName().length() > 31 and
```
lcartey, line 21:
```suggestion
    this.getName().length() >= 31 and
```
I think we want to capture the case where one identifier is exactly 31 characters long, and another is more than 31, but both match in the first 31 characters. Suggest adding a test case for that as well.

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
```diff
@@ -0,0 +1,51 @@
...
+  string getNonSignificantName() { result = this.getName().suffix(31) }
+}
+
+from ExternalIdentifiers d, ExternalIdentifiers d2
+where
+  not isExcluded(d, Declarations1Package::externalIdentifiersNotDistinctQuery()) and
+  not d = d2 and
+  d.getLocation().getStartLine() >= d2.getLocation().getStartLine() and
+  d.getSignificantName() = d2.getSignificantName() and
+  not d.getNonSignificantName() = d2.getNonSignificantName()
```
lcartey, line 48: I think it would be better just to use the name here:
```suggestion
  not d.getName() = d2.getName()
```
For two reasons:
 * If one of the matches is exactly 31 characters long, the `suffix` will not return anything, so we will never match.
 * It avoids creating more strings, which can cause performance issues.

lcartey (changes requested):
Sorry, just a couple more things! Thank you!

c/cert/src/rules/DCL31-C/DeclareIdentifiersBeforeUsingThem.ql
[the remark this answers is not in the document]
knewbury01: ok, accepted the library result for this!

c/cert/src/rules/DCL31-C/DeclareIdentifiersBeforeUsingThem.ql
knewbury01: done

rule_packages/c/Declarations1.json
knewbury01: done

rule_packages/c/Declarations1.json
knewbury01: done

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
knewbury01: done

c/misra/src/rules/RULE-5-1/ExternalIdentifiersNotDistinct.ql
knewbury01: done

lcartey (approved):
Looks good, thanks for the changes!

PR #47, opened 2022-08-11
Implement Pointers2 package

kraiouchkine:
## Description

Implements the C Coding Standards Pointers2 package

closes github/coding-standards-team#763

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - ARR39-C
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

  - [ ] Confirmed

### Author

### Reviewer 


c/cert/test/rules/ARR39-C/test.c
```diff
@@ -0,0 +1,30 @@
...
+struct s1 {
+  int v1, v2, v3;
+};
+
+void f2(int p1, int p2) {
+  int *v1;
+  v1 += p1; // NON_COMPLIANT
+  v1 += p2; // COMPLIANT
+}
+
```
jsinglet, line 12: In the rule I think the gist of it is that pointer arithmetic should only be allowed if the arithmetic concerns the thing the pointer points to. In the above example `*v1` doesn't point to anything, so there should be no valid pointer arithmetic allowed if I am understanding correctly.

jsinglet (changes requested):
Looking good @kraiouchkine! I have a question about the test case -- can you take a look and let me know?

c/cert/test/rules/ARR39-C/test.c
kraiouchkine: ARR39-C does not differentiate between results based on the underlying value of the pointer. The rule only generally prohibits performing pointer arithmetic with a scaled integer operand. You are right; the test case contains undefined behavior, such as that pointer arithmetic on an uninitialized variable, but none of it falls under the scope of this rule or affects the tests.

jsinglet: Makes sense, thank you! I had a closer look and I agree.

jsinglet (approved):
Looks good @kraiouchkine! Thanks for this!

PR #65, opened 2022-08-18
`M3-2-1`: Exclude non-object and template variables

### LINKED ISSUE #34: `M3-2-1`: Do not consider variable template instantiations as a redeclaration with a different type
### Affected rules
 - `M3-2-1`

### Description

In CodeQL variable templates are represented by multiple different variables (an uninstantiated copy, plus a copy for each instantiation).  As each of these can have a different type, they currently cause false positives for `M3-2-1`, which is about finding redeclarations of the same object with incompatible types.

The easiest way to exclude templates is to remove versions from template instantiations:
```codeql
  not decl1.isFromTemplateInstantiation(_) and
  not decl2.isFromTemplateInstantiation(_)
```

While we are making modifications to this query, there are a few other changes we should make:
 * Exclude member variables, as they are not "objects".
 * Provide a link to each of the types so that the user can verify that the types are different. This is important because the two types may have the same name, but may be defined differently.
   ```codeql
   select decl1, "The object $@ of type $@ is not compatible with re-declaration $@ of type $@", decl1,
   decl1.getName(), decl1.getType(), decl1.getType().toString(), decl2, decl2.getName(),
   decl2.getType(), decl2.getType().toString()
   ```

lcartey:
## Description

Fixes #34.

   - Reduces false positives by excluding non-object variables (for example, member variables).
   - Reduces false positives by excluding variable templates and template instantiations.
   - Improved the reported error message by including the conflicting type names.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - `M3-2-1`

## Release change checklist

  - [x] Yes
  - [ ] No

  - [ ] Confirmed

### Author

### Reviewer 


lcartey:
@mbaluda I seem to have typo'd the date of the release note in this case to [change_notes/2022-09-18-m3-2-1-templates.md](https://github.com/github/codeql-coding-standards/pull/65/files#diff-f3645c3938cee589f86846e41e8383d374777a7c936fd711ae77b0d16e7dee2c), which I think is fine (it was supposed to be August).

In general the date of the change note is not relevant - the date doesn't appear in the release notes, and the relevant change notes are gathered by querying git to determine which files are new (rather than using the creation date on the file).

PR #67, opened 2022-08-22
Contracts2

mbaluda:
## Description

Rule package Contracts2

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - ENV32-C
     - ENV34-C
     - RULE-21-19
     - RULE-21-20
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

  - [ ] Confirmed

### Author

### Reviewer 

### DIFF
--- a/c/cert/src/rules/ENV32-C/ExitHandlersMustReturnNormally.ql
+++ b/c/cert/src/rules/ENV32-C/ExitHandlersMustReturnNormally.ql
@@ -31,7 +31,7 @@ class RegisteredAtexit extends FunctionAccess {
   }
 }
 
-/*
+/**
  * Nodes of type Function, FunctionCall or FunctionAccess that \
  * are reachable from a redistered atexit handler and
  * can reach an exit function.


c/cert/src/rules/ENV34-C/DoNotStorePointersReturnedByCertainFunctions.ql
```diff
@@ -0,0 +1,43 @@
...
+  // A variable `v` is assigned with `GetEnvFunctionCall` return value
+  exists(Variable v |
+    fc1 = v.getInitializer().getExpr()
+    or
+    exists(Assignment ae | fc1 = ae.getRValue() and v = ae.getLValue().(VariableAccess).getTarget())
+  ) and
+  // A second `GetEnvFunctionCall`
+  fc2 = fc1.getASuccessor+() and
+  // The value is is accessed in `e` afer the second `GetEnvFunctionCall`
+  e = fc2.getASuccessor+() and
```
lcartey, line 40: This approach doesn't take into account definition reachability, which means it will be confused by variable reuse:
```cpp
  char* temp = getEnv("VAR1");
  use(temp);
  temp = getEnv("VAR2");
  use(temp);
```
And also loops:
```cpp
for(std:string envVar : envVars) {
  char* temp = getEnv(envVar);
  use(temp);
}
```
The usual fix for this type of reachability problem is to use `StackVariableReachability`, however, you want to assert that the use is via a second call to the same function, which is not supported.

Furthermore, I think the overall approach risks missing results where the repetition is non-local - i.e. stored in a field or global variable.

Therefore, my suggestion would be to flag _all_ cases where the result is stored in a variable, and then use a `StackVariableReachability` to exclude simple cases where no duplicate calls can occur. You need to be a bit careful here - any call to a function could _indirectly_ call the same function, and I think we should err on the side of reporting results in that case.




If you want to implement this approach, I suggest using `StackVariableReachabilityWithReassignment` to track

lcartey (changes requested):
Have completed a first pass review, with a few suggestions.

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+    "codeQL.cli.executablePath": "/Users/mauro/GitHub/coding-standards/codeql-bundle-20220120/codeql",
+    "codeQL.runningQueries.autoSave": true
+}
```
lcartey, line 4: Accidentally committed file.

c/cert/src/rules/ENV32-C/ExitHandlersMustReturnNormally.ql
```diff
@@ -0,0 +1,75 @@
...
+/*
+ * Nodes of type Function, FunctionCall or FunctionAccess that \
+ * are reachable from a redistered atexit handler and
+ * can reach an exit function.
+ */
+class InterestingNode extends ControlFlowNode {
+  InterestingNode() {
+    exists(Function f |
+      (
+        this.(Function) = f and
```
lcartey, line 43: No need for an explicit cast here:
```suggestion
        this = f and
```

c/cert/src/rules/ENV34-C/DoNotStorePointersReturnedByCertainFunctions.ql
lcartey: Ah, apologies - using dataflow does indeed address the variable reuse case, but not the loop case.

> I added a second version of the query using StackVariableReachabilityWithReassignment to remove false positives due to variable rewrite.

I actually intended something more drastic here - that we flag every assignment of a returned pointer as an alert, and only consider it "safe" if we are convinced that every use of the assigned variable is safe.

The `StackVariableReachabilityWithReassignment` could then be used to determine what uses are "safe", i.e. find def-use pairs which don't go through a direct or indirect call to the problematic APIs. If all uses of the assignment are safe, and the pointer doesn't escape the function after the function ends, then we don't flag it, otherwise we do.

`StackVariableReachabilityWithReassignment` doesn't work with global variables or fields, so we  would have to take a different approach for those. I do think it's pretty dubious to store any returned pointers in global variables or fields, though - it just seems to be asking for trouble.

I don't if overall this approach would have too many false positives. However, I do think the current approach could have too many false negatives (with globals, fields etc.) and we should try to capture those as well.

lcartey (changes requested):
Added a couple more comments.

c/misra/src/rules/RULE-21-19/ValuesReturnedByLocaleSettingUsedAsPtrToConst.ql
```diff
@@ -0,0 +1,80 @@
...
+  not isExcluded(e, Contracts2Package::valuesReturnedByLocaleSettingUsedAsPtrToConstQuery()) and
+  (
+    isNonConstVar(e, c)
+    or
+    isNonConstParam(e, c)
+    or
+    modifiesInternalString(e, c)
+  )
+select e,
+  "The pointer returned by $@ shell only be used as a pointer to const-qualified type as modifying the pointed object leads to unspecified behavior.",
```
lcartey, line 79:
```suggestion
  "The pointer returned by $@ shall only be used as a pointer to const-qualified type as modifying the pointed object leads to unspecified behavior.",
```

cpp/common/src/codingstandards/cpp/rules/constlikereturnvalue/ConstLikeReturnValue.qll
```diff
@@ -0,0 +1,69 @@
...
+ * Call to functions that return pointers to environment objects that should not be modified.
+ */
+
+class NotModifiableCall extends FunctionCall {
+  NotModifiableCall() {
+    this.getTarget().hasGlobalName(["getenv", "setlocale", "localeconv", "asctime", "strerror"])
+  }
+}
+
+/*
```
lcartey, line 25:
```suggestion
/**
```
And reformat to attach the comment to the `ObjectWrite` class.

cpp/common/src/codingstandards/cpp/rules/constlikereturnvalue/ConstLikeReturnValue.qll
```diff
@@ -0,0 +1,69 @@
...
+import codingstandards.cpp.Customizations
+import codingstandards.cpp.Exclusions
+import semmle.code.cpp.dataflow.DataFlow
+import DataFlow::PathGraph
+
+abstract class ConstLikeReturnValueSharedQuery extends Query { }
+
+Query getQuery() { result instanceof ConstLikeReturnValueSharedQuery }
+
+/*
```
lcartey, line 15:
```suggestion
/**
```

cpp/common/src/codingstandards/cpp/rules/invalidatedenvstringpointers/InvalidatedEnvStringPointers.qll
```diff
@@ -0,0 +1,75 @@
...
+import cpp
+import codingstandards.cpp.Customizations
+import codingstandards.cpp.Exclusions
+import semmle.code.cpp.dataflow.DataFlow
+
+abstract class InvalidatedEnvStringPointersSharedQuery extends Query { }
+
+Query getQuery() { result instanceof InvalidatedEnvStringPointersSharedQuery }
+
+/*
```
lcartey, line 15:
```suggestion
/**
```

lcartey (changes requested):
Just a few comment issues to fix, otherwise 👍 . Note that `/**` is required for any comment attached to a predicate or class. `/*` is a free-floating block comment.

lcartey (approved):

PR #69, opened 2022-08-24
use the new non-deprecated Xml instead of XML

erik-krogh:
## Description

I recently renamed XML to Xml throughout github/codeql: https://github.com/github/codeql/pull/10073. 
Now I see that you use the old version here, so I thought I should fix that.   
I don't know if this will work.  

## Change request type

## Release change checklist

  - [ ] Yes
  - [x] No

  - [ ] Confirmed


erik-krogh:
I see from `supported_codeql_configs.json` that CodeQL 2.7.6 is the only supported version.  
Which means that this change isn't relevant.

PR #70, opened 2022-08-25
Rename isPOD to isPod to silence deprecation warnings

erik-krogh:
## Description

This PR targets the `next` branch.

I want to rename a bunch of acronyms to use camelCase/PascalCase to follow our own style guide for QL.  
(See this draft PR: https://github.com/github/codeql/pull/10153).   

When I do that 4 tests in this repo start to fail because of deprecation warnings because if the `idPOD()` predicate. 
I'll let them fail for now, and after I've merged the semmle-code PR I'll fix the deprecation warnings. 

## Change request type

 - [x] No rules added

## Release change checklist

  - [ ] Yes
  - [x] No

  - [ ] Confirmed

### DIFF
--- a/cpp/autosar/src/codingstandards/cpp/HardwareOrProtocolInterface.qll
+++ b/cpp/autosar/src/codingstandards/cpp/HardwareOrProtocolInterface.qll
@@ -39,7 +39,7 @@ class DefinedSizeType extends Type {
 
 class DefinedSizeClass extends Class {
   DefinedSizeClass() {
-    this.isPOD() and
+    this.isPod() and
     forall(Field f | f = this.getAField() | f.getType() instanceof DefinedSizeType)
   }
 }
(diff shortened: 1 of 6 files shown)


lcartey:
I think we would be better changing our use of `isPOD` to `isPod` on the `next` branch, rather than trying put workarounds in here. @jketema what do you think?

jketema:
The policy is that the internal coding standards test may temporarily be broken while we're in the middle for fixing things. So the breakage is no problem. It should not be blocking merging, if it does we have configured something wrongly.

So I agree with @lcartey that we do not need a workaround. That does mean that the CI here will temporarily break until we update to the next release on the `next` branch, but that should also not considered to be blocking.

jketema:
TL;DR: Just rename `isPOD` to `isPod` and let us know when the internal CI checks should be expected to be green again.

jketema:
Also, yes, this should target `next` and not `main` as `next` is the branch we use in internal testing.

erik-krogh:
I've merged the [github/codeql](https://github.com/github/codeql/pull/10153) and [semmle-code](https://github.com/github/semmle-code/pull/43527) PRs that do the rename from `isPOD()` to `isPod()`.   
Now I think we just wait for the next release that includes that rename.

jketema (approved):

jketema:
Merging this now to get the internal CI up and running again. We'll fix the one here when 2.11.0 gets released.

PR #112, opened 2022-10-31
Tests for Language2

knewbury01:
## Description

tests for package Language 2

I think RULE-1-1 is compiler covered, if we are assuming/currently supporting one compiler - it seems its more about edge cases from other compilers. also no concrete `NON_COMPLIANT` examples are given in the rule description.

I also think DIR-2-1 is compiler covered, both because it seems to be an import of A1-4-3 (which is compiler covered) but also because the compiler does seem like the best place to report compiler errors

The DIR-4-2 test will be the exact same as M7-4-1 test , checked and I agree its a plain import and the test for M7-4-1 doesnt have anything c++ specific. I forget what we do when we want to share the test file, rather than copy the content?

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - RULE-1-4
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

  - [ ] Confirmed

### Author

### Reviewer 

### DIFF
--- a/c/misra/test/rules/RULE-1-4/test.c
+++ b/c/misra/test/rules/RULE-1-4/test.c
@@ -1,0 +1,26 @@
+#include <stdalign.h>    //NON_COMPLIANT
+#include <stdatomic.h>   //NON_COMPLIANT
+#include <stdnoreturn.h> //NON_COMPLIANT
+#include <threads.h>     //NON_COMPLIANT
+
+#define MACRO(x) _Generic((x), int : 0, long : 1) // NON_COMPLIANT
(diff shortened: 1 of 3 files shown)


knewbury01:
# Notes RULE-1-4

the annex K bit could be detected case by case but my logic was that - [from this](https://port70.net/~nsz/c/c11/n1570.html#K.3.1.1p1) - if the `__STDC_WANT_LIB_EXT1__` macro meant to guard the inclusion of these extra cases, then we can look for that instead of case by case feature use

jsinglet (approved):

PR #99, opened 2022-10-12
Package Preprocessor6

knewbury01:
## Description

add DIR-4-9

## Change request type

 - [ ] Release or process automation (GitHub workflows, internal scripts)
 - [ ] Internal documentation
 - [ ] External documentation
 - [x] Query files (`.ql`, `.qll`, `.qls` or unit tests)
 - [ ] External scripts (analysis report or other code shipped as part of a release)

## Rules with added or modified queries

 - [ ] No rules added
 - [x] Queries have been added for the following rules:
     - DIR-4-9
 - [ ] Queries have been modified for the following rules:
     - _rule number here_

## Release change checklist

A change note ([development_handbook.md#change-notes](https://github.com/github/codeql-coding-standards/blob/main/docs/development_handbook.md#change-notes)) is required for any pull request which modifies:

 - The structure or layout of the release artifacts.
 - The evaluation performance (memory, execution time) of an existing query.
 - The results of an existing query in any circumstance.

If you are only adding new rule queries, a change note is not required.

_**Author:**_ Is a change note required? 
  - [ ] Yes
  - [x] No

_**Reviewer:**_ Confirm that either a change note is not required or the change note is required and has been added.
  - [ ] Confirmed

## Query development review checklist

For PRs that add new queries or modify existing queries, the following checklist should be completed by both the author and reviewer:

### Author

 - [x] Have all the relevant rule package description files been checked in?
 - [x] Have you verified that the metadata properties of each new query is set appropriately?
 - [x] Do all the unit tests contain both "COMPLIANT" and "NON_COMPLIANT" cases?
 - [x] Are the alert messages properly formatted and consistent with the [style guide](https://github.com/github/codeql-coding-standards/blob/main/docs/development_handbook.md#query-style-guide)?
 - [x] Have you run the queries on OpenPilot and verified that the performance and results are acceptable?<br />_As a rule of thumb, predicates specific to the query should take no more than 1 minute, and for simple queries be under 10 seconds. If this is not the case, this should be highlighted and agreed in the code review process._
 - [x] Does the query have an appropriate level of in-query comments/documentation?
 - [x] Have you considered/identified possible edge cases?
 - [x] Does the query not reinvent features in the standard library?
 - [x] Can the query be simplified further (not golfed!)

### Reviewer 

 - [ ] Have all the relevant rule package description files been checked in?
 - [ ] Have you verified that the metadata properties of each new query is set appropriately?
[...]

PR #113, opened 2022-11-01
Package Contracts5

mbaluda:
## Description

- [x] ERR32-C
- [x] ERR33-C

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - ERR32-C
     - ERR33-C
     - ERR30-C

## Release change checklist

  - [ ] Yes
  - [x] No

  - [x] Confirmed

### Author

### Reviewer 


c/cert/src/rules/ERR33-C/DetectAndHandleStandardLibraryErrors.ql
```diff
@@ -0,0 +1,207 @@
...
+import codingstandards.c.cert
+import semmle.code.cpp.commons.NULL
+import codingstandards.cpp.ReadErrorsAndEOF
+import semmle.code.cpp.dataflow.DataFlow
+
+/**
+ * Classifies error returning function calls based on the
+ * type and value of the required checked
+ */
+class ExpectedErrReturn extends FunctionCall {
```
knewbury01, line 23: recently on my Preprocessor6 package Remco [gave me some feedback that abstract classes](https://github.com/github/codeql-coding-standards/pull/99/files#diff-026dd0360dc3233a8e85504b5c38c638d74caa50800b07a43335f8eaf399fc31R8) were a neat way to do some case handling, I think considering either using abstract classes or at least more types for the specific cases covered here might make this more readable

c/cert/src/rules/ERR33-C/DetectAndHandleStandardLibraryErrors.ql
```diff
@@ -0,0 +1,207 @@
...
+    op.getAnOperand() = err.getErrValue() and
+    // special case for function `realloc` where the returned pointer
+    // should not be invalidated
+    not (
+      err.getTarget().hasName("realloc") and
+      op.getAnOperand().(VariableAccess).getTarget() =
+        err.getArgument(0).(VariableAccess).getTarget()
+    )
+  ) and
+  // EXCEPTIONS
```
knewbury01, line 181: similarly , might be good to consider a few more predicate/classes whichever for the below bits for readability sake

rule_packages/c/Contracts5.json
```diff
@@ -0,0 +1,38 @@
+{
+  "CERT-C": {
+    "ERR32-C": {
```
knewbury01, line 3: Im having a moment here, but doesnt including this in the package json usually cause CI failure without the rest of the files? or did we change that a long time ago?

rule_packages/c/Contracts5.json
```diff
@@ -0,0 +1,38 @@
...
+      ],
+      "title": "Do not rely on indeterminate values of errno"
+    },
+    "ERR33-C": {
+      "properties": {
+        "obligation": "rule"
+      },
+      "queries": [
+        {
+          "description": "Detect and handle standard library errors.",
```
knewbury01, line 26: I would consider moving the "Undetected failures can lead to unexpected or undefined behavior" part of the alert msg to here instead

c/cert/src/rules/ERR33-C/DetectAndHandleStandardLibraryErrors.ql
```diff
@@ -0,0 +1,207 @@
...
+ * @tags external/cert/id/err33-c
+ *       correctness
+ *       external/cert/obligation/rule
+ */
+
+import cpp
+import codingstandards.c.cert
+import semmle.code.cpp.commons.NULL
+import codingstandards.cpp.ReadErrorsAndEOF
+import semmle.code.cpp.dataflow.DataFlow
```
knewbury01, line 17: the `import codingstandards.cpp.ReadErrorsAndEOF` includes `import semmle.code.cpp.dataflow.DataFlow`, although I guess I am not sure on if its better to explicitly import here also, or rm the redundant line

c/cert/test/rules/ERR33-C/test.c
```diff
@@ -0,0 +1,137 @@
+#include <inttypes.h>
```
knewbury01, line 1: was just kind of skimming results in openpilot, saw this line `float scale = getenv("SCALE") != NULL ? std::stof(getenv("SCALE")) : 1.0;` come up in qt_window.cc:15:53, is this a FP?

knewbury01 (commented):
Hey @mbaluda ! I've made some initial comments, will need to come back for a more deep look at implementation after you've had some thoughts on these comments :)

you always have the most "case heavy" rules it seems!

c/cert/src/rules/ERR32-C/DoNotRelyOnIndeterminateValuesOfErrno.ql
```diff
@@ -0,0 +1,50 @@
...
+where
+  not isExcluded(errno, Contracts5Package::doNotRelyOnIndeterminateValuesOfErrnoQuery()) and
+  // errno read in the handler
+  exists(SignalCallingHandler sc |
+    sc.getHandler() = h and
+    (
+      sc.calls*(errno.getEnclosingFunction())
+      or
+      // errno is read after the handle
+      errno.(ControlFlowNode).getAPredecessor+() = sc.getHandler()
```
knewbury01, line 47: I think this is the same as `errno.getAPredecessor+() = sc.getHandler()` bc `ErrnoRead extends Expr`, `Expr extends StmtParent`, `StmtParent extends ControlFlowNode`

c/cert/test/rules/ERR32-C/test.c
```diff
@@ -0,0 +1,42 @@
+#include <errno.h>
```
knewbury01, line 1: do we want to add the other COMPLIANT case, as like a self documenting type thing?
```
void handlerX(int signum) {
  pfv old_handler = signal(signum, SIG_DFL);
  if (old_handler == SIG_ERR) {
    _Exit(0); // COMPLIANT
  }
}
```

c/cert/test/rules/ERR32-C/test.c
```diff
@@ -0,0 +1,42 @@
...
+
+void handler3(int signum) { pfv old_handler = signal(signum, SIG_DFL); }
+
+pfv helper4(int signum) {
+  perror(""); // NON_COMPLIANT
+  return signal(signum, SIG_DFL);
+}
+void handler4(int signum) { pfv old_handler = helper4(signum); }
+
+int main(void) {
```
knewbury01, line 31: is this also NON_COMPLIANT?
```
int main(void) {
  pfv old_handler = signal(SIGINT, handler1);
  if (old_handler == SIG_ERR) {
    perror(""); // NON_COMPLIANT
  }

  old_handler = signal(SIGINT, handler2);

  old_handler = signal(SIGINT, handler3);

  old_handler = signal(SIGINT, handler4);

  old_handler = signal(SIGINT, handlerX);

  FILE *fp = fopen("something", "r");
   if( fp == NULL ) {
      perror("Error: "); // NON_COMPLIANT?
   }
}
```
either way this one generates alerts for results for every matching call to signal above, do we want to consider just the closest above? (although with no results on openpilot this may actually not be an alert explosion issue at all, your call!)

c/cert/test/rules/ERR32-C/test.c
```diff
@@ -0,0 +1,42 @@
+#include <errno.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/wait.h>
+
+typedef void (*pfv)(int);
+
+void handler1(int signum) {
```
knewbury01, line 9: is this COMPLIANT? (currently shows up as NON_COMPLIANT) - picking on the "the corresponding signal handler obtained a SIG_ERR return from a call to the signal function." wording of the rule. what if errno is accessed on not that type of exit status from the signal function called in the signal handler...?
```
void handler1(int signum) {
  pfv old_handler = signal(signum, SIG_DFL);
  if (old_handler == SIG_ERR) {
    perror(""); // NON_COMPLIANT
  } else {
    perror(""); // COMPLIANT?
  }
}
```

knewbury01 (commented):
ERR32-C reviewed! I was trying to be quite picky on edge cases but I am not totally sure if my testcase suggestions are realistic. otherwise looking good, these rules are very cool !

will submit a separate (re)review for ERR33-C

c/cert/test/rules/ERR32-C/test.c
knewbury01: er, sorry just to clarify I mean this one:
```
FILE *fp = fopen("something", "r");
   if( fp == NULL ) {
      perror("Error: "); // NON_COMPLIANT?
   }
```
fopen will "re-determinate" the errno? or is it possible that the errno will still have been last used for something related to a signal event... if the signal was seen after the fopen call ...

c/cert/src/rules/ERR33-C/DetectAndHandleStandardLibraryErrors.ql
```diff
@@ -0,0 +1,275 @@
...
+ */
+abstract class ExpectedErrReturn extends FunctionCall {
+  Expr errValue;
+  ComparisonOperation errOperator;
+
+  Expr getErrValue() { result = errValue }
+
+  ComparisonOperation getErrOperator() { result = errOperator }
+}
+
```
knewbury01, line 31: this is fairly easy to read as is but maybe add a quick ql dock comment on each with the sample syntax it covers

knewbury01 (commented):
@mbaluda both queries reviewed, 
ERR33-C looks good to me !
2 minor questions on ERR32-C, but after those tiny things I think it will be good to go

c/cert/test/rules/ERR32-C/test.c
```diff
@@ -0,0 +1,64 @@
...
+
+  old_handler = signal(SIGINT, handler4);
+
+  old_handler = signal(SIGINT, handler5);
+
+  FILE *fp = fopen("something", "r");
+  if (fp == NULL) {
+    perror("Error: "); // NON_COMPLIANT
+  }
+}
```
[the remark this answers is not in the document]
knewbury01: er, not quite, its not that "signal(SIGABRT, ... equivalent to calls to abort()" 

the thing I am unsure of is: is accessing errno fine in `handler6` in the snippet above? I think that it gets reported as NON_COMPLIANT but I think it is COMPLIANT. 

the reason I think its fine is that the rule states "the value of errno is referred to after a signal occurred **other than as the result of calling the abort or raise**", well we only get to `handler6` when abort/raise is called. so the value of errno is reliable then. I think we want to omit handlers registered to handle `SIGABRT`

I could still be wrong, lmk if we should sync to see if my logic makes sense or not! thanks!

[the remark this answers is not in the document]
knewbury01: as per discussion with @mbaluda I have [opened a question](https://wiki.sei.cmu.edu/confluence/display/c/ERR32-C.+Do+not+rely+on+indeterminate+values+of+errno?focusedCommentId=392527955#comment-392527955) on the rule description page to determine if my case is relevant or not

we have concluded that this case can be added as an enhancement later if it turns out that there truly will be false positives in this unique case

knewbury01 (approved):
I believe we have combed this PR over very thoroughly!

good work @mbaluda :)

PR #157, opened 2023-01-05
Implement C Declarations6 package

knewbury01:
and refactor Identifiers to include new predicate

## Description

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - RULE-17-3
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

  - [ ] Confirmed

### Author

### Reviewer 


knewbury01:
# Notes RULE-17-3

The description also talks a bit about function prototypes (which are inherently not implicit declarations) however I think checking that in this rule would be too strict of an interpretation, and function prototype usage is already covered in RULE-8-2 anyways.

knewbury01:
# Notes RULE-5-9

I omitted consideration of the exception case, "Inline functions with internal linkage may be defined in more than one translation unit if they are in the same header file" - this isnt really relevant to CodeQL bc the database only includes one `Declaration` in this case either way, so re-adding the duplicate detection across multiple translation units then re-removing it makes no sense. 

Plus if we think about it, what would the action item be for: "a declaration has the same name - in the same file - just across multiple translation units"... make sure no file could be used in multiple translation units? that wouldnt make sense to me at least.

knewbury01:
# Notes RULE-8-10

I have opted for the most strict implementation of this rule so far. The rule title says the most strict fix, but the rationale describes how the undefined behaviour actually only occurs in a specific subset of cases when an inline function omits the static storage class. I can take another look if it currently actually seems too strict.

c/misra/src/rules/RULE-5-9/IdentifiersWithInternalLinkageNotUnique.ql
```diff
@@ -0,0 +1,32 @@
...
+  d1.isStatic() and
+  d1.isTopLevel() and
+  not d1 = d2 and
+  d1.getName() = d2.getName() and
+  // Apply an ordering based on location to enforce that (d1, d2) = (d2, d1) and we only report (d1, d2).
+  (
+    d1.getFile().getAbsolutePath() < d2.getFile().getAbsolutePath()
+    or
+    d1.getFile().getAbsolutePath() = d2.getFile().getAbsolutePath() and
+    d1.getLocation().getStartLine() < d2.getLocation().getStartLine()
```
kraiouchkine, line 30: I think you could implement the deduplication with a simple comparison of the Location rather than having to check both path and line numbers. The implementation does work as-is though.

rule_packages/c/Declarations6.json
```diff
@@ -0,0 +1,143 @@
...
+          "precision": "very-high",
+          "severity": "error",
+          "short_name": "IdentifiersWithInternalLinkageNotUnique",
+          "tags": [
+            "maintainability",
+            "readability"
+          ]
+        }
+      ],
+      "title": "Identifiers that define objects or functions with internal linkage should be unique"
```
kraiouchkine, line 80: I would document the omission of detection of non-unique (multiple translation units) inline functions defined in a header file from the query's implementation in implementation_scope.

kraiouchkine (changes requested):

kraiouchkine:
Completely agree on the rationale for both of your implementation notes 👍  `RULE-5-9` is extremely noisy as-is though; I'm still wondering if there is a more useful interpretation of the rule.

LGTM other than missing implementation notes and a suggestion for preventing permutations in `RULE-5-9`.

c/misra/src/rules/RULE-5-9/IdentifiersWithInternalLinkageNotUnique.ql
knewbury01: I actually copied it from [here](https://github.com/github/codeql-coding-standards/blob/main/cpp/autosar/src/rules/M8-4-2/NonIdenticalIdentifierUsedForTheParameterInReDeclarationOfAFunction.ql#L35), now that Ive used it a few times though it could be extracted to a predicate... I just was thinking that the general form to accommodate the different times its been used would be with `Locatable`s, in which case that sounds terrible to be evaluated without the context of the query... also though I've used this in like 3 queries now so is that enough to amortize?  lmk what you think!

kraiouchkine (approved):

PR #165, opened 2023-01-16
Implement C Declarations7 package

knewbury01:
## Description

Decl7 package

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - RULE-8-12
     - RULE-18-8
     - DCL36-C (wont implement decision - pls verify if agreed - notes on justification below
     - DCL39-C (import DCL55-CPP)
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


knewbury01:
# Notes omit DCL36-C

compiler checked by clang (and according to wiki, gcc) **but not** MVSC

verified with:
testcase non compliant example from the [wiki](https://wiki.sei.cmu.edu/confluence/display/c/DCL36-C.+Do+not+declare+an+identifier+with+conflicting+linkage+classifications)

as well as this testcase built upon the table depicted in the wiki (except if the "no linkage" header actually said "no specifier" [which I think is what is it is meant to say - to be confirmed](https://wiki.sei.cmu.edu/confluence/display/c/DCL36-C.+Do+not+declare+an+identifier+with+conflicting+linkage+classifications?focusedCommentId=393642249#comment-393642249))
```
////TABLE - specifiers

static int i1; //internal linkage
static int i1; //COMPLIANT - internal linkage


int i2; //no linkage
static int i2; //NON_COMPLIANT - undefined - compiler detected

extern int i3; //external linkage
static int i3; ///NON_COMPLIANT - undefined - compiler detected


//-----
static int i11; //internal linkage
int i11; //NON_COMPLIANT - undefined - compiler detected

int i22; //no specifier
int i22; //COMPLIANT - no specifier


extern int i33; //external linkage
int i33; //NON_COMPLIANT - undefined according to table - but not compiler detected - so... COMPLIANT?


//-----
static int i111; //internal linkage
extern int i111; //COMPLIANT - internal linkage

int i222; //no specifier
extern int i222; //COMPLIANT - external linkage

extern int i333; //external linkage
extern int i333; //COMPLIANT - external linkage
```

cpp/common/src/codingstandards/cpp/rules/informationleakageacrossboundaries/InformationLeakageAcrossBoundaries.qll
```diff
@@ -0,0 +1,17 @@
+/**
+ * Provides a library which includes a `problems` predicate for reporting....
```
jsinglet, line 2: Perhaps add a description here?

c/cert/src/rules/DCL39-C/InformationLeakageAcrossTrustBoundariesC.ql
```diff
@@ -0,0 +1,22 @@
...
+ *       security
+ *       external/cert/obligation/rule
+ */
+
+import cpp
+import codingstandards.c.cert
+import codingstandards.cpp.rules.informationleakageacrossboundaries.InformationLeakageAcrossBoundaries
+
+class InformationLeakageAcrossTrustBoundariesCQuery extends InformationLeakageAcrossBoundariesSharedQuery {
+  InformationLeakageAcrossTrustBoundariesCQuery() {
```
jsinglet, line 19: I know this is an import BUT: as written I don't think this query is addressing the problem in the rule -- specifically, it doesn't appear to handle the cases it describes -- namely the edge cases in structure padding with it is non-homogeneous. 

Perhaps we should list out the limitations of this query in the implementation details?

```c

struct test {
  int a;
  char b;
  int c;
};
 
/* Safely copy bytes to user space */
extern int copy_to_user(void *dest, void *src, size_t size);
 
void do_stuff(void *usr_buf) {
  struct test arg = {.a = 1, .b = 2, .c = 3};
  copy_to_user(usr_buf, &arg, sizeof(arg));
}
```

jsinglet (changes requested):
Looking good @knewbury01! Some minor comments and a question about how you want to handle edge cases not being handled.

cpp/common/src/codingstandards/cpp/rules/informationleakageacrossboundaries/InformationLeakageAcrossBoundaries.qll
knewbury01: ah , good catch! I always miss those... 

done!

c/cert/src/rules/DCL39-C/InformationLeakageAcrossTrustBoundariesC.ql
knewbury01: I agree an implementation note would be a good idea!

if I use the same testcase (the first NCE) in the [DCL55-CPP](https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL55-CPP.+Avoid+information+leakage+when+passing+a+class+object+across+a+trust+boundary) rule on the unmodified query that also does not get reported

since the related guidelines for DCL55-CPP points to DCL39-C and all of the examples are the same between them, and I believe the conditions are the same , I didnt spend too too much time digging into exactly what the CPP version was able to handle . but yeah now looking [closer I see ](https://github.com/github/codeql-coding-standards/blob/main/cpp/common/src/codingstandards/cpp/trustboundary/UninitializedField.qll#L22) it only handles cases where the fields werent initialized by an initializer, moreso focuses on missing initialization than padding.

knewbury01: (added the note!)

jsinglet: Thanks @knewbury01 -- I think that's enough for now. As an additional step I wanted to suggest to open a issue to track this as an enhancement. I think that there is a large enough gap that we should probably keep an eye on it.

jsinglet (approved):
LGTM!

c/cert/src/rules/DCL39-C/InformationLeakageAcrossTrustBoundariesC.ql
knewbury01: added an enhancement issue as well!
https://github.com/github/codeql-coding-standards/issues/175

PR #176, opened 2023-01-27
Implement InvalidMemory1 package

kraiouchkine:
## Description

This pull request implements rules in the `InvalidMemory1` package.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
    - EXP33-C: `DoNotReadUninitializedMemory.ql` (closes github/coding-standards-team#743)
    - EXP34-C: `DoNotDereferenceNullPointers.ql` (closes github/coding-standards-team#744)
    - MEM30-C: `DoNotAccessFreedMemory.ql` (closes github/coding-standards-team#746)
    - RULE-9-1: `ObjectWithAutoStorageDurationReadBeforeInit.ql` (closes github/coding-standards-team#747)
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


c/cert/test/rules/MEM30-C/test.c
```diff
@@ -0,0 +1,26 @@
+#include <stdlib.h>
```
knewbury01, line 1: more of a question - but right now we definitely cover "dereferences after free" but what about the part of the rule that says "using it as an operand of an arithmetic operation, type casting it, and using it as the right-hand side of an assignment" 

do we want to also cover some of these cases? (I dont quite understand why stuff like pointer arithmetic (without dereference) would be bad per say - if it doesnt access the freed memory - but maybe a case of confusing intention?) 
```
  void test_freed(char *input) {
  char *buf = (char *)malloc(strlen(input) + 1);
  free(buf);

  char c0 = *buf; // NON_COMPLIANT

  char *newbuf;
  newbuf = buf; // NON_COMPLIANT? - RHS of an assignment
  char c1 = *newbuf;

  *buf = 'a'; // this is reported but then should the deref 2 lines down also be?
  // or is that getting into more of access unintialized memory...
  buf++;
  char c2 = *buf; // NON_COMPLIANT?

  double *d;
  *d = 0.9;
  free(d);
  char *p = (char *)&d; // NON_COMPLIANT? - typecast

  char *buf2 = (char *)malloc(strlen(input) + 1);
  free(buf2);
  free(buf2);       // NON_COMPLIANT
  char c02 = *buf2; // NON_COMPLIANT
}
```

c/cert/src/rules/MEM30-C/DoNotAccessFreedMemory.ql
```diff
@@ -0,0 +1,63 @@
...
+class UseAfterFreeReachability extends StackVariableReachability {
+  UseAfterFreeReachability() { this = "UseAfterFree" }
+
+  override predicate isSource(ControlFlowNode node, StackVariable v) { isFreeExpr(node, v) }
+
+  override predicate isSink(ControlFlowNode node, StackVariable v) { isDerefExpr(node, v) }
+
+  override predicate isBarrier(ControlFlowNode node, StackVariable v) {
+    definitionBarrier(v, node) or
+    isFreeExpr(node, v)
```
knewbury01, line 53: also a more for my understanding - why is a second free a barrier?

c/cert/src/rules/MEM30-C/DoNotAccessFreedMemory.ql
```diff
@@ -0,0 +1,63 @@
+/**
```
knewbury01, line 1: oh! if this query is the same as the [useafterfree](https://github.com/github/codeql/blob/main/cpp/ql/src/Critical/UseAfterFree.ql#L16) from standard library, we probably want to cite that somehow , so that if the upstream changes we can propagate fixes or if they are running the full coding standards+out of box queries then they can figure out why this would show up 2x

maybe something like [this](https://github.com/github/codeql-coding-standards/blob/main/c/cert/src/rules/FIO32-C/DoNotPerformFileOperationsOnDevices.ql#L21).- maybe with a CodeQL lib version number...?

knewbury01 (commented):
agree with all the import decisions

just a couple really minor things/q's on MEM30-C, other than that looks good!

c/cert/test/rules/MEM30-C/test.c
knewbury01: also, I wrote this then saw that its pretty much an import - maybe we can just add the extra testcases/or an implementation note, if you think the rule is making sense with the list of things to not do

kraiouchkine: Thanks for noticing this! I did not extend the implementation to cover those additional cases as I interpreted "dereference" as "access". Looking back, you're right, and my interpretation is wrong. Here is an excerpt from a C99 Rationale document (https://www.open-std.org/jtc1/sc22/wg14/www/C99RationaleV5.10.pdf):

> An invalid pointer might be created in several ways. An arbitrary value can be assigned (via a cast) to a pointer variable. (This could even create a valid pointer, depending on the value.) A pointer to an object becomes invalid if the memory containing the object is deallocated or moved by realloc. Pointer arithmetic can produce pointers outside the range of an array.

> Regardless how an invalid pointer is created, any use of it yields undefined behavior. Even
assignment, comparison with a null pointer constant, or comparison with itself, might on some
systems result in an exception.

> Consider a hypothetical segmented architecture on which pointers comprise a segment descriptor
and an offset. Suppose that segments are relatively small so that large arrays are allocated in
multiple segments. While the segments are valid (allocated, mapped to real memory), the
hardware, operating system, or C implementation can make these multiple segments behave like
a single object: pointer arithmetic and relational operators use the defined mapping to impose
30 the proper order on the elements of the array. Once the memory is deallocated, the mapping is
no longer guaranteed to exist. Use of the segment descriptor might now cause an exception, or
the hardware addressing logic might return meaningless data. 

I will modify the query and test to model those additional test-cases.

c/cert/src/rules/MEM30-C/DoNotAccessFreedMemory.ql
kraiouchkine: Changes made.

c/cert/test/rules/MEM30-C/test.c
kraiouchkine: Changes made.

c/cert/src/rules/MEM30-C/DoNotAccessFreedMemory.ql
kraiouchkine: not entirely sure what edge/test-case the standard library is trying to account for with that, but I left it in given that (as far as I know) queries that make it into the standard library are checked against and tuned to minimise FPs across a ton of real-world codebases. it also makes sense by looking at double frees as a separate issue that could be lead to way more undefined behaviour than just invalid pointers and be trickier to confirm without FPs (think more complicated CFGs and guardws, allocator behaviour, etc.).

c/cert/src/rules/MEM30-C/DoNotAccessFreedMemory.ql
kraiouchkine: Changes made.

c/cert/src/rules/MEM30-C/DoNotAccessFreedMemory.ql
knewbury01: hm makes sense! thanks for giving it a thought

c/cert/test/rules/MEM30-C/test.c
knewbury01: lgtm!

knewbury01 (approved):
thanks for making those changes, they look good to me!

PR #181, opened 2023-02-03
Sorting out false positives for A0-1-2 and A0-1-4

jeongsoolee09:
## Description

Strengthen query by getting rid of false positive cases for both Autosar A0-1-2: Unused Return Value and A0-1-4: Unused Parameter.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - Autosar A0-1-2
     - Autosar A0-1-4

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


cpp/common/src/codingstandards/cpp/rules/unusedparameter/UnusedParameter.qll
```diff
@@ -32,10 +32,13 @@ query predicate lambdaExprParamHasAccess(Parameter parameter) {
...
+  (
+    not isMaybeUnusedParameter(p) and
+    f = p.getFunction() and
+    // Virtual functions are covered by a different rule
+    not f.isVirtual()
+  ) and
   message = "Unused parameter '" + p.getName() + "' for function $@." and
-  fName = f.getQualifiedName()
+  // fName = f.getQualifiedName()
+  fName = "TODO."
```
jeongsoolee09, line 43: @rvermeulen If the function is referred to by a lambda expression, then `f.getQualifiedName()` becomes a goal that cannot be satisfied, hence the entire row gets excluded. What would be a nice alternative to the formula?

jeongsoolee09:
Query performance for A0-1-2 is not good (33.8s):
```
$ codeql database analyze --format=sarif-latest --output=a0-1-2.sarif --search-path=./ -- ../codeql-coding-standards-release-engineering/data/commaai-openpilot-72d1744d830bc249d8761a1d843a98fb0ced49fe-cpp/ cpp/autosar/src/rules/A0-1-2/UnusedReturnValue.ql

Running queries.
Compiling query plan for .../codeql-coding-standards/cpp/autosar/src/rules/A0-1-2/UnusedReturnValue.ql.
Resolving database upgrade for .../codeql-coding-standards/cpp/autosar/src/rules/A0-1-2/UnusedReturnValue.ql
[1/1 comp 4.1s] Compiled .../codeql-coding-standards/cpp/autosar/src/rules/A0-1-2/UnusedReturnValue.ql.
UnusedReturnValue.ql: [1/1 eval 33.8s] Results written to autosar-cpp-coding-standards/rules/
Shutting down query evaluator.
Interpreting results.
```

cpp/autosar/src/rules/A0-1-2/UnusedReturnValue.ql
```diff
@@ -39,8 +40,16 @@ where
...
-  // Rule allows disabling this rule where a static_cast<void> is applied
-  not fc.getExplicitlyConverted().(StaticCast).getActualType() instanceof VoidType
+  // Rule allows disabling this rule where a static_cast<void> or a C-style cast to void is applied
+  not (
+    fc.getExplicitlyConverted().(StaticCast).getActualType() instanceof VoidType
+    or
+    exists(CStyleCast cast |
+      not cast.isCompilerGenerated() and
+      cast.getExpr() = fc and
+      cast.getActualType() instanceof VoidType
```
lcartey, line 52: This case should replicate the first case, e.g.
```
or
fc.getExplicitlyConverted().(CStyleCast).getActualType() instanceof VoidType
```
As I don't think there's any intended semantic difference between the two case, so they should be identical in structure.

Note: `getExplicitlyConverted()` is the right concept for this query, because we want casts that are explicit in the code. An implicit conversion is not sufficient as "sanitizer" for this query. `not cast.isCompilerGenerated() and cast.getExpr() = fc` mostly does what you want in terms of excluding implicit conversions, but has some strange behaviour in the case that there are both explicit and implicit conversions, or parenthesis (which are also treated as conversions). In these cases, the conversions "stack", so `cast.getExpr()` may return another conversion - for example, consider `(foo) (f())`.

`getExplicitlyConverted()` ensures we get the explicit conversion associated with the function call return value.

An alternative to repeating the cases would be to unify them:
```
exists(Cast c |
  c instanceof StaticCast
  or
  c instanceof CStyleCast
  |
  c= fc.getExplicitlyConverted() and
  c.getActualType() instanceof VoidType
)
 ```

File: cpp/common/src/codingstandards/cpp/Operator.qll [hunk 3]
```diff
@@ -124,7 +124,7 @@ class AssignmentOperator extends MemberFunction {
     // operator op, where op is =, +=, -=, *=, /=, %=, ^=, &=, |=, >>=
     exists(string op |
       "operator" + op = this.getName() and
-      op in ["=", "+=", "-=", "*=", "/=", "%=", "^=", "&=", "|=", ">>="]
+      op in ["=", "+=", "-=", "*=", "/=", "%=", "^=", "&=", "|=", ">>=", "<<="]
```
lcartey, line 127: Good spot, but can you also update the comment a few lines above?

cpp/common/src/codingstandards/cpp/deadcode/UnusedParameters.qll
```diff
@@ -12,6 +12,7 @@ import cpp
  */
 class UsableParameter extends Parameter {
   UsableParameter() {
+    /* Regular Function */
```
lcartey, line 15: I think this comment is redundant

lcartey (changes requested):
Looks good! Just a few comments on some stylistic details, and some subtleties around casts/conversions.

lcartey (approved):

lcartey:
@jeongsoolee09 I just approved this, but noticed we're missing a change note. Can you please raise a separate PR to add a change note as per the developer handbook:
https://github.com/github/codeql-coding-standards/blob/main/docs/development_handbook.md#change-notes
Thanks!

PR #189, opened 2023-02-14
Implement C package: Declarations8

knewbury01:
## Description

Decl8

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - DCL30-C
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


knewbury01:
# Notes DCL30-C

I chose to implement a specific subset of cases for this rule, as opposed to the general title/description of the rule. The examples in the wiki only describe the cases that I chose to cover and the tool comparison part of the wiki/logistics details described confirm that handling more general cases is a more difficult problem

### This query was very close to several pre-existing queries...

1) the C++ rule [M7-5-1](https://github.com/github/codeql-coding-standards/blob/main/cpp/autosar/src/rules/M7-5-1/FunctionReturnAutomaticVarCondition.ql) covers one specific part of the cases that I chose to cover (however C does not have references so I opted to write one query to unify the cases and omit the unnecessary reference bits)
2) the out of the box queries - [ReturnStackAllocatedMemory](https://github.com/github/codeql/blob/main/cpp/ql/src/Likely%20Bugs/Memory%20Management/ReturnStackAllocatedMemory.ql) and [StackAddressEscapes](https://github.com/github/codeql/blob/main/cpp/ql/src/Likely%20Bugs/Memory%20Management/StackAddressEscapes.ql) also cover pretty similar scenarios, however as I said, I chose to unify the results into one query and additionally `StackAddressEscapes` has the same FP that my version does so I think maybe the results are not too different?

### possible expansions

if it seems necessary for this query to cover more of the ways that storage durations can conflict or lead to undefined behaviour , possibly [this query](https://github.com/github/codeql-coding-standards/blob/main/c/cert/src/rules/CON34-C/AppropriateThreadObjectStorageDurations.ql) for thread storage durations could be a shared query split for the thread case specifically

lcartey:
There is actually an even closer shared query we've written: [DoNotCopyAddressOfAutoStorageObjectToOtherObject.qll](https://github.com/github/codeql-coding-standards/blob/main/cpp/common/src/codingstandards/cpp/rules/donotcopyaddressofautostorageobjecttootherobject/DoNotCopyAddressOfAutoStorageObjectToOtherObject.qll)

This is used in:
 * `M7-5-2` `AssignmentOfEscapingAutoStorage.ql`
 * `Rule 18.6` `AutomaticStorageObjectAddressCopiedToOtherObject.ql`

I think the query you've written covers essentially the same topics, so I think I would recommend replacing your query with an import query for the same library. The shared query could be extended if it doesn't cover all the cases (return statements?).

knewbury01:
> There is actually an even closer shared query we've written: [DoNotCopyAddressOfAutoStorageObjectToOtherObject.qll](https://github.com/github/codeql-coding-standards/blob/main/cpp/common/src/codingstandards/cpp/rules/donotcopyaddressofautostorageobjecttootherobject/DoNotCopyAddressOfAutoStorageObjectToOtherObject.qll)
> 
> This is used in:
> 
> * `M7-5-2` `AssignmentOfEscapingAutoStorage.ql`
> * `Rule 18.6` `AutomaticStorageObjectAddressCopiedToOtherObject.ql`
> 
> I think the query you've written covers essentially the same topics, so I think I would recommend replacing your query with an import query for the same library. The shared query could be extended if it doesn't cover all the cases (return statements?).

@lcartey, thanks for the review! I took a look,

I think DCL30-C == M7-5-1 + M7-5-2 (this one is actually using the out of the box equivalent for stack addresses that I found, but is the refined for our purposes version, so thank you for the much better find!)

I have now split and use the shared lib for the M7-5-2 part

however now that I look at the `M7-5-1` implementation , I still think that maybe my version of it makes more sense for C (omitting the references stuff and using dataflow instead of not), unless you think that both need an update?

c/cert/src/rules/DCL30-C/AppropriateStorageDurationsFunctionReturn.ql
```diff
@@ -0,0 +1,50 @@
...
+ *       external/cert/obligation/rule
+ */
+
+import cpp
+import codingstandards.c.cert
+import codingstandards.cpp.FunctionParameter
+import semmle.code.cpp.dataflow.DataFlow
+
+class Source extends StackVariable {
+  Source() { not this instanceof FunctionParameter }
```
lcartey, line 20: I don't think `FunctionParameter` is needed here - you want to exclude all parameters, not just those from definitions or instantiated templates.
```suggestion
  Source() { not this instanceof Parameter }
```

c/cert/src/rules/DCL30-C/AppropriateStorageDurationsFunctionReturn.ql
```diff
@@ -0,0 +1,50 @@
...
+}
+
+abstract class Sink extends DataFlow::Node { }
+
+class FunctionSink extends Sink {
```
lcartey, line 25: Combine `FunctionSink` and `Sink` - you don't need the abstract class, as you only have one sub-class.

c/cert/src/rules/DCL30-C/AppropriateStorageDurationsFunctionReturn.ql
```diff
@@ -0,0 +1,50 @@
...
+  FunctionSink() {
+    //output parameter
+    exists(FunctionParameter f |
```
lcartey, line 28: Again, we don't need `FunctionParameter` here.
```suggestion
    exists(Parameter f |
```

lcartey (approved):
Just some minor stylistic comments, otherwise looks good to me!

c/cert/src/rules/DCL30-C/AppropriateStorageDurationsFunctionReturn.ql
knewbury01: agree, addressed!

c/cert/src/rules/DCL30-C/AppropriateStorageDurationsFunctionReturn.ql
knewbury01: right, I think that was leftover maybe from the previous unified query, good catch!

c/cert/src/rules/DCL30-C/AppropriateStorageDurationsFunctionReturn.ql
knewbury01: agree, addressed!

PR #348, opened 2023-08-23
Fix for A5-1-3 false positives

rak3-sh:
## Description

A5-1-3's original implementation raised alerts for cases where lambdas had an explicit argument list. However, it raised a false positive in case the lambda had an argument list but was defined inside a macro.

```cpp
#define PARAM_MACRO [](int i) { i; };

int test_lambda_in_macro()
{
  PARAM_MACRO // False positive here.
  return 0;
}
```
As a fix, A5-1-3 is modified to consider only those lambdas which have zero arguments, as non-zero arguments in a lambda function would require the user to specify an explicit argument list. Adding this check, removed all of the false positives that were seen in the project for this rule.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - A5-1-3

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 

### DIFF
--- a/cpp/autosar/src/rules/A5-1-3/LambdaExpressionWithoutParameterList.ql
+++ b/cpp/autosar/src/rules/A5-1-3/LambdaExpressionWithoutParameterList.ql
@@ -21,6 +21,10 @@ where
   not isExcluded(lambda, LambdasPackage::lambdaExpressionWithoutParameterListQuery()) and
   lambdaFunction = lambda.getLambdaFunction() and
   not lambdaFunction.isAffectedByMacro() and
+  // If it has a parameter, then it will have an
+  // explicit parameter list. Therefore, proceed to check only if the lambda
+  // does not have any parameters.
+  not exists (lambdaFunction.getAParameter()) and
   // The extractor doesn't store the syntactic information whether the parameter list
   // is enclosed in parenthesis. Therefore we cannot determine if the parameter list is
   // explicitly specified when the parameter list is empty.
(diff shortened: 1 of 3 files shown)


cpp/autosar/src/rules/A5-1-3/LambdaExpressionWithoutParameterList.ql
```diff
@@ -21,6 +21,10 @@ where
   not isExcluded(lambda, LambdasPackage::lambdaExpressionWithoutParameterListQuery()) and
   lambdaFunction = lambda.getLambdaFunction() and
   not lambdaFunction.isAffectedByMacro() and
+  // If it has a parameter, then it will have an
+  // explicit parameter list. Therefore, proceed to check only if the lambda
+  // does not have any parameters.
+  not exists (lambdaFunction.getAParameter()) and
```
lcartey, line 27: Can you please provide an updated test case? For example, add a macro that expands to a lambda with parameters:
```cpp
#define PARAM_MACRO [](int i) { i; };
```
Then expand that in the test function:
```cpp
PARAM_MACRO // COMPLIANT
```

lcartey (changes requested):
Thanks @rak3-sh, this looks like a good improvement. Just requires a small change to the test case, and an update to the change note.

cpp/autosar/src/rules/A5-1-3/LambdaExpressionWithoutParameterList.ql
rak3-sh: Thanks for the suggestion. I've added this in: [a7dd290](https://github.com/github/codeql-coding-standards/pull/348/commits/a7dd290a650f9997c97763680bbbb8fa3574e0db)

lcartey (approved):

lcartey:
@rak3-sh you need to format the test file using `clang-format` - you can see the formatting failure here:
https://github.com/github/codeql-coding-standards/actions/runs/6019907111/job/16339180294?pr=348

rak3-sh:
Formatted using clang-format in https://github.com/github/codeql-coding-standards/pull/348/commits/4297f0a82e461f2076f3db0834d8e79eadb7b8b3

Issue #359, opened 2023-09-28
`A7-3-1`: `HiddenInheritedNonOverridableMemberFunction` is very slow


anthony-williams-woven:
When running the CodeQL check suite against my repository this rule takes a lot longer than any other rules: it takes 49 minutes whereas the next slowest rule takes 13 minutes.

This repository has ~80k source lines.

This is being run in an Ubuntu 20.04 docker container on Ubuntu 22.04

Issue #360, opened 2023-09-28
`A7-3-1`: `HiddenInheritedNonOverridableMemberFunction` reports false positives


anthony-williams-woven:
### Affected rules
 - A7-3-1/HiddenInheritedNonOverridableMemberFunction

### Description

A7-3-1 reports violations on `operator=` in derived classes. For assignment of a derived class to do anything meaningful, it *must* provide an `operator=`, and cannot rely on the base class definition. Any report of a violation on `operator=` is thus a false positive.

### Example
```cpp
class base_class {};

class derived : private base_class {
 public:
  derived& operator=(const derived&) noexcept = default;
};
```

lcartey:
I think this is a general problem with special member functions, which should be excluded from this rule. This is because special member functions are never inherited.

knewbury01:
this wasnt previously auto closed simply because the syntax of the linked issue in the PR was not correct, needs to use the full keyword syntax for each issue (ie closes X and closes Y not closes X and Y)

PR #365, opened 2023-09-29
`A7-3-1`: Improve performance, reduce false positives

lcartey:
## Description

Fixes https://github.com/github/codeql-coding-standards/issues/359 and https://github.com/github/codeql-coding-standards/issues/360.

This pull request addresses a number of issues with `HiddenInheritedNonOverridableMemberFunction.ql`:
   - Addresses potential performance issues and fix false positives related to an incorrect implementation of ordering between `using` declarations and function declarations in a class. We previously required the using declaration to appear first, but order does not matter.
   - Addresses potential performance issues related to matching functions by name.
   - Reduce duplication by reporting `Function`s not `FunctionDeclarationEntry`s in both the hiding and hidden declarations.
   - Exclude special member functions, which cannot be inherited.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - `A7-3-1`

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


cpp/autosar/test/rules/A7-3-1/test.cpp
```diff
@@ -60,3 +60,25 @@ void f1() {
...
   l3.f2(l4); // calls C1:f2(S1)
 }
+
+class C5 : public C1 {
+public:
+  void f1(double); // COMPLIANT
+  using C1::f1;    // order of using and f1 declaration is not relevant
+
+  void f2(double) override; // COMPLIANT
+  using C1::f2;             // order of using and f1 declaration is not relevant
```
rvermeulen, line 70:
```suggestion
  using C1::f2;             // order of using and f2 declaration is not relevant
```

rvermeulen (commented):
Do we want to apply the same optimization for cpp/autosar/src/rules/A7-3-1/HiddenInheritedOverridableMemberFunction.ql?

lcartey:
> Do we want to apply the same optimization for cpp/autosar/src/rules/A7-3-1/HiddenInheritedOverridableMemberFunction.ql?

I updated `HiddenInheritedOverridableMemberFunction.ql` to apply the same logic around `using` declarations - e.g. not to be sensitive to the line location. I did not observe a need for the other optimizations.

cpp/autosar/src/rules/A7-3-1/HiddenInheritedNonOverridableMemberFunction.ql
```diff
@@ -15,24 +15,43 @@
...
-    ude.getDeclaration() = hiddenDecl.getDeclaration() and
-    ude.getEnclosingElement() = overridingDecl.getDeclaration().getDeclaringType() and
-    ude.getLocation().getStartLine() < overridingDecl.getLocation().getStartLine()
+    ude.getDeclaration() = hiddenDecl and
+    ude.getEnclosingElement() = overridingDecl.getDeclaringType()
   ) and
   // Exclude compiler generated member functions which include things like copy constructor that hide base class
   // copy constructors.
-  not overridingDecl.getDeclaration().isCompilerGenerated()
+  not overridingDecl.isCompilerGenerated() and
```
rvermeulen, line 52: The OpenPilot results show we are finding issues in macros like `QObject`.
The suggestion removes 38 of those results.

```suggestion
  not overridingDecl.isCompilerGenerated() and
  not overridingDecl.isInMacroExpansion() and
```

rvermeulen (changes requested):
A suggestion to remove results that cannot be acted upon by developers.

cpp/autosar/src/rules/A7-3-1/HiddenInheritedNonOverridableMemberFunction.ql
lcartey: I don't think we should generally exclude results for this rule because they were generated by a macro. If I use a macro to generate a member function that hides base class member functions it still contravenes the rule, and addressable by adding a "using ...".

For OpenPilot, QT has a bit of an unusual meta-object system, implemented through the `Q_OBJECT` macro. This macro generates common member functions for QT objects. Reviewing the results for this rule we seemed to be flagging 3 generated functions: `qt_static_metacall`, `tr` and `trUtf8`. `tr` and `trUtf8` seem to be genuine contraventions of this rule - as the macro is generating an implementation of a non-virtual member function that hides the base class implementation.

`qt_static_metacall` is interesting because it's actually a private (static) member function. I don't think we should be reporting _private_ member functions from the base class, because they would not be visible in the subclass anyway, so no hiding is going on. I've therefore updated the query to exclude this case.

rvermeulen (approved):
We keep contraventions in macro expansions and rely on deviations to exclude those that are caused by patterns outside of the control of the developer (e.g., Qt framework)

Issue #230, opened 2023-03-03
`A0-1-1`: `auto` local variables in uninstantiated templates are reported as unused


lcartey:
### Affected rules
 - `A0-1-1`

### Description

Where an `auto` variable is declared in a template function or class, the copy of the variable in the uninstantiated template may be considered "unused", even though it is considered used in at least one instantiation.

### Example

```cpp
#include <vector>
template <typename T> void test_range_based_for_loop_template() {
  std::vector<A> values_;
  for (auto &elem : values_) { // COMPLIANT - should not report either elem or
                               //             the compiler generated (__range)
                               //             variable in the uninstantiated
                               //             template
    elem;
  }
}

// Instantiate the template
void test_template() { test_range_based_for_loop_template<A>(); }
```

lcartey:
Updated with a refinement of our understanding of the problem.

lcartey:
I believe I've determined the cause of this issue - I've updated the issue body to reflect this.

Issue #402, opened 2023-10-06
`A0-1-1`: Exclusions need to apply to the definition, not the variable


lcartey:
### Affected rules
 - `A0-1-1`

### Description

We need to update the exclusion mechanism to consider the `SsaDefinition` not the variable, otherwise the result cannot be deviated.

lcartey:
This was fixed as part of https://github.com/github/codeql-coding-standards/pull/414.

Issue #372, opened 2023-10-02
`A7-1-1`: Exclude compiler generated statements for range based for loops


lcartey:
### Affected rules
 - `A7-1-1`
 - `A0-1-1`

### Description

Similar to:
https://github.com/github/codeql-coding-standards/issues/309

### Example
```cpp
  for (const bool value : values) {
    do_stuff(value);
  }
```

lcartey:
As part of this work we should do a systematic analysis of the queries using QL for QL to identify queries reporting `Variable`s as their primary location where the `isCompilerGenerated` is not included as a condition on the `where` clause.

rvermeulen:
A0-1-1 is already address by @lcartey in b8a527cacc3.
Cannot replicate the issue for A7-1-1 because the range based variables are already filtered out.
I will explicitly exclude compiler generated variables, but we unfortunately cannot add a test case for this report.

PR #414, opened 2023-10-25
`A0-1-1`: Ignore incomplete or compiler generated variables

lcartey:
## Description

Fixes https://github.com/github/codeql-coding-standards/issues/230 and https://github.com/github/codeql-coding-standards/issues/402.

Also addresses https://github.com/github/codeql-coding-standards/issues/372 but for `A0-1-1` only.

This PR makes the following adjustments:

 * Excludes assignments to variables that occur in uninstantiated templates. Uninstantiated templates provide a necessarily incomplete model of the code - with certain types, call targets and so forth only being resolved once the template is instantiated. It is therefore not possible, or necessary, to determine whether a variable is unused.
 * Excludes compiler-generated variables, such as those generated for range-based for loops (e.g. `(__range)`). With the range-based for loops these compiler-generated variables are generally not "unused", because the compiler will also generate a use for each variable. However, if they appeared in uninstantiated templates then the model may be incomplete. I would therefore expect most range-based for loop results to be removed by the first change, but I think it's prudent to exclude all compiler-generated variables to avoid false positives.
 * Enable deviations on the assignment as well as the variable itself.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - `A0-1-1`

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


cpp/autosar/src/rules/A0-1-1/UselessAssignment.ql
```diff
@@ -20,6 +20,7 @@ import codingstandards.cpp.deadcode.UselessAssignments
 
 from SsaDefinition ultimateDef, InterestingStackVariable v
 where
+  not isExcluded(ultimateDef, DeadCodePackage::uselessAssignmentQuery()) and
   not isExcluded(v, DeadCodePackage::uselessAssignmentQuery()) and
```
rvermeulen, line 24:
```suggestion
  not isExcluded(v, DeadCodePackage::uselessAssignmentQuery()) and
```

Let remove the exclusion on the variable itself to prevent confusion around the cases

```
int x; // suppress all useless assignments of x

x = 0;

if (...) {
  x = 1;
} else {
  x = 2;
  x = 3;
}
```

and

```
int x = 0; // suppress all or only the current useless assignement

if (...) {
  x = 1;
} else {
  x = 2;
  x = 3;
}
```

rvermeulen, line 24:
```suggestion
  not isExcluded(ultimateDef, DeadCodePackage::uselessAssignmentQuery()) and
```

Lets remove the exclusion on the variable itself to prevent confusion around the cases

```
int x; // suppress all useless assignments of x

x = 0;

if (...) {
  x = 1;
} else {
  x = 2;
  x = 3;
}
```

and

```
int x = 0; // suppress all or only the current useless assignement

if (...) {
  x = 1;
} else {
  x = 2;
  x = 3;
}
```

rvermeulen (dismissed):
See comment on exclusion on variable.

lcartey:
@rvermeulen ready for re-review.

change_notes/2023-10-25-a0-1-1.md
```diff
@@ -0,0 +1,4 @@
+ * `A0-1-1` - address a number of false positive issues:
+   * Exclude compiler-generated variables, such as those generated for range-based for loops.
+   * Exclude variables in uninstantiated templates, for which we have no precise data on uses.
+   * Deviations can now be applied to the useless assignment as well as the variable itself.
```
rvermeulen, line 4:
```suggestion
   * Deviations can now be applied to the useless assignment as well as the variable itself.
```
With the removal of `isExcluded` on `v` it can only be applied to the useless assignment.

rvermeulen (changes requested):
Looks good. Just a final check on the change note stating that a deviation can be applied to both the assignment and the variable, which I believe is no longer the case.

change_notes/2023-10-25-a0-1-1.md
lcartey: Done.

rvermeulen (approved):

PR #425, opened 2023-11-08
`M6-5-5`: Improve detection of modification by reference

### LINKED ISSUE #380: `M6-5-5`: Consider `const` when determining whether an escaped reference modifies
### Affected rules
 - `M6-5-5`

### Description

`M6-5-5` identifies cases where a loop control variable is modified in the _condition_ of a loop. It identifies two kinds of modifications: direct modification, and indirect modification - where a reference to the loop counter variable escapes the local scope, and could be modified. We currently assume that any reference that escapes the scope could be modified, but we should factor in whether the escaping function is const/non-const.

Specifically, we need to use `isAddressOfAccessNonConst()` instead of `isAddressOfAccess()` here:
https://github.com/github/codeql-coding-standards/blob/47c8a57dbe8197bad4d6ebc9b3968ad998ba8352/cpp/common/src/codingstandards/cpp/Loops.qll#L263

### Example

```cpp
void loop(std::vector v) {
  std::vector<int>::iterator first = v.begin();
  std::vector<int>::iterator last = v.end();
  for (; first != last; first++) { // call to operator!= passes a reference to first
  }
}
```

lcartey:
## Description

Fixes #380.

This PR makes the following changes:
 - No longer consider the taking of a const reference as a modification.
 - Consider taking a non-const reference of a potential loop counter as modifying that loop counter.
 - Update tests to reflect changes.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - `M6-5-5`

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


cpp/common/src/codingstandards/cpp/Loops.qll
```diff
@@ -106,6 +106,15 @@ Variable getALoopCounter(ForStmt fs) {
     )
     or
     updateOp = result.getAnAssignedValue()
+    or
+    // updateOp is an access whose address is taken in a non-const way
+    exists(FunctionCall fc, VariableAccess va |
+      fc = updateOp and
+      fc instanceof FunctionCall and
```
rvermeulen, line 113: Superfluous condition
```suggestion

```

change_notes/2023-11-05-m6-5-5-const-refs.md
```diff
@@ -0,0 +1,3 @@
+ * `M6-5-5`
```
rvermeulen, line 1:
```suggestion
 - `M6-5-5`:
```

Since this is indirectly changing queries I don't think we want to list them.

rvermeulen (changes requested):
Minor request changes. Otherwise good to go.

rvermeulen:
The alert message of `LoopControlVariableModifiedInLoopExpression.ql`is not adhering to our style guide. Not sure if you want to address that in this PR.

rvermeulen (approved):
:shipit:

PR #497, opened 2024-01-24
Exceptions: do not rethrow from functions with incompatible exception specifications.

### LINKED ISSUE #159: `A15-4-2`: Only report the first `noexcept` function that throws
### Affected rules
 - `A15-4-2`
 - `ERR55-CPP`

### Description

`noexcept` functions frequently call other `noexcept` functions. If a `noexcept` function throws an exception, then we currently report that as a violation in every `noexcept` function that calls that function. This creates multiple alerts for one issues. We should instead report the original instance and not the others.

### Example
```cpp
void f1() noexcept {
  throw foo(); // report this case
}

void f2() noexcept {
  f1(); // do not report this case
}
```

lcartey:
## Description

Per `[except.spec]`, by default if an exception specification is violated, the exception is not rethrown and `std::terminate` is called. I have therefore modified the `ExceptionFlow.qll` library to no longer propagate exceptions thrown by `noexcept` or dynamic specification functions where the exception type isn't listed. This should significant reduce the number of false positives reported by this rule.

Note: this has impacted a number of different queries that used the base exception flow library.

Fixes #159.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
   - `A15-4-2`, `ERR55-CPP` - reduce false positives for `noexcept` functions which call other `noexcept` function which may throw.
   - `A15-2-2` - reduce false positives for constructors which call `noexcept` functions.
   - `A15-4-5` - reduce false positives for checked exceptions that are thrown from `noexcept` functions called by the original function.
   - `DCL57-CPP` - do not report exceptions thrown from `noexcept` functions called by deallocation functions or destructors.
   - `A15-5-1`, `M15-3-1` - do not report exceptions thrown from `noexcept` functions called by special functions.

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


cpp/common/src/codingstandards/cpp/exceptions/ExceptionFlow.qll
```diff
@@ -312,7 +313,21 @@ class ReThrowExprThrowingExpr extends ReThrowExpr, ThrowingExpr {
...
 /** An expression which calls a function which may throw an exception. */
 class FunctionCallThrowingExpr extends FunctionCall, ThrowingExpr {
-  override ExceptionType getAnExceptionType() { result = getAFunctionThrownType(getTarget(), _) }
+  override ExceptionType getAnExceptionType() {
+    exists(Function target |
+      target = getTarget() and
+      result = getAFunctionThrownType(target, _) and
+      // [expect.spec] states that throwing an exception type that is prohibited
+      // by the specification will result in the program terminating. We therefore
+      // do not propagate such exceptions to the call sites for the function.
```
rvermeulen, line 322:
```suggestion
      // [expect.spec] states that throwing an exception type that is prohibited
      // by the specification will result in the program terminating, unless
      // a custom `unexpected_handler` is registered that throws an exception type
      // which is compatible with the specification.
      // We assume a default `std::unexpected` handler and therefore
      // do not propagate such exceptions to the call sites for the function.
      //
```

cpp/cert/test/rules/ERR55-CPP/test_no_except.cpp
```diff
@@ -1,9 +1,34 @@
...
+}
+
+void test_indirect_throw_5() noexcept {
+  noexceptIndirectThrowA(); // COMPLIANT - noexceptIndirectThrowA would call
+                            // std::terminate() if ExceptionA is thrown
+}
+
+void test_indirect_throw_6() noexcept {
+  indirectThrowA(); // NON_COMPLIANT
 }
```
rvermeulen, line 34:
```suggestion
}

void test_indirect_throw_6() noexcept {
    dynamicIndirectThrowA(); // NON_COMPLIANT[FALSE_NEGATIVE] - dynamicIndirectThrowA indirectly throws std::bad_exception
}
```

cpp/autosar/test/rules/A15-4-2/test.cpp
```diff
@@ -21,4 +23,13 @@ void test_indirect_throw_3() noexcept(false) {
 
 void test_indirect_throw_4() {
   throwA(); // COMPLIANT - function marked as noexcept(false)
+}
+
+void test_indirect_throw_5() noexcept {
+  noexceptIndirectThrowA(); // COMPLIANT - noexceptIndirectThrowA would call
```
rvermeulen, line 29: A similar FN exists as for EXP55-CPP when a `std::bad_exception` can still be thrown.

rvermeulen (changes requested):
Added a FN test case and suggested a comment change to capture an assumption. Not sure if we want to over approximate and rethrow `std::bad_exception` if that is part of the dynamic exception specification of a function that throws.

cpp/cert/test/rules/ERR55-CPP/test_no_except.cpp
lcartey: Ah, this is actually not a false negative. `dynamicIndirectThrowA` will not throw a `std::bad_exception` unless:

  * [`std::set_unexpected`](http://en.cppreference.com/w/cpp/error/set_unexpected) has been called with a custom unexpected error handler (note: this handler functionality was deprecated in C++11 and removed in C++17).
  * The custom handler re-throws an exception, and the exception it rethrows is not compatible with the dynamic exception specification.

This is shown in this example here: https://en.cppreference.com/w/cpp/error/bad_exception. Given this relates to deprecated functionality in our supported C++ standards (C++14 onwards), I don't think we need to implement this edge case or add a test for it. Let me know if you disagree.

Note: the [ERR55-CPP](https://wiki.sei.cmu.edu/confluence/display/cplusplus/ERR55-CPP.+Honor+exception+specifications) CERT rule has an incorrect statement about this behaviour:

> If a function declared with a dynamic-exception-specification throws an exception of a type that would not match the exception-specification, the function std::unexpected() is called **The behavior of this function can be overridden but, by default, causes an exception of std::bad_exception to be thrown**

That is not true - the default behaviour is for it to call `std::terminate`. You have to set the `unexpected` handler to trigger any of this behaviour.

cpp/common/src/codingstandards/cpp/exceptions/ExceptionFlow.qll
lcartey: Added, although I reworded it slightly for clarity.

cpp/cert/test/rules/ERR55-CPP/test_no_except.cpp
```diff
@@ -1,9 +1,34 @@
...
-  throw "test";
+void test_throw() noexcept(true) {
+  throw ExceptionA(); // NON_COMPLIANT - function marked as noexcept(true)
 }
 
-void test_noexcept_true() noexcept(true) { // NON_COMPLIANT
-  throw "test";
+void throwA() { throw ExceptionA(); }
+void indirectThrowA() { throwA(); }
+void noexceptIndirectThrowA() noexcept { throwA(); } // NON_COMPLIANT
```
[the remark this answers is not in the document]
lcartey: See below for rationale on not adding this case.

cpp/autosar/test/rules/A15-4-2/test.cpp
lcartey: See below for rationale on not adding this case.

lcartey:
Thanks for the review! As `std::bad_exception` is never thrown with the default handler (only via custom handlers), and the handler mechanism is deprecated in C++14, I have opted not to add the additional test cases - let me know if you disagree with that. I have clarified the comment in `ExceptionFlow.qll` to explain how it works and why we exclude it.

cpp/cert/test/rules/ERR55-CPP/test_no_except.cpp
rvermeulen: Agreed, it is only a FP in the case of a custom unexpected handler that rethrows an unspecified exception type. Good catch on the mistake in CERT!

rvermeulen (approved):
Given that the corner case is no longer possible in new standards it is time to :shipit: without.

Issue #406, opened 2023-10-06
`A13-3-1`: Ignore false overloads with the same parameter type


lcartey:
### Affected rules
 - `A13-3-1`

### Description

`getAnOverload()` produces incorrect results when files are extracted multiple times in different contexts. To avoid producing false positives, we should exclude cases where the "overload" has the same parameter type (i.e. an r-value reference type) as the original function being overloaded.

 ### Example

```cpp
template <typename... T> void suppress_unused_warning(T &&...) {}
```

knewbury01:
was not sure if this had to do with instantiations of the template in the database looking like separate functions, 

however, cannot replicate with:
```
template <typename T> void suppress_unused_warning(T &&...) {}

int test_406() {
  suppress_unused_warning(1);
  suppress_unused_warning("test");
}
```

this might have to do with the implicit copy/move constructors... which technically the examples in the rule from the standard doc only show alerts when there are explicit overloads (not involving the implicit copy/move ctors) , so I have chosen to omit those for this

PR #514, opened 2024-01-30
A13-3-1: exclude functions with different number parameters

### LINKED ISSUE #399: `A13-3-1`: Only report functions with the same number of parameters
### Affected rules
 - `A13-3-1`

### Description

The rule title for this rule states:
> A function that contains “forwarding reference” as its argument shall not be overloaded.

With the rationale saying:
> Overloading functions with “forwarding reference” argument may lead to developer’s confusion on which function will be called.

Although the rule title suggests overloads with different numbers of parameters should be prohibited, there is no confusion in this case, and therefore I believe it's safe to exclude them.

### Example

```cpp
class foo {
public:
  foo() {} // error here
  template<typename T>
  foo(T&&){}
};
```

knewbury01:
## Description

fixes #399 

note in the testcase added, it is still an expected behaviour for the rule to alert on the overloading that is occurring between the function with the forward ref and the implicit copy and move constructors. this is because the testcase does not fit the exception condition where those are required to be constrained.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - A13-3-1

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


change_notes/2024-01-30-exclusion-a13-3-1.md
```diff
@@ -0,0 +1,2 @@
+`A13-3-1`: `cpp/autosar/function-that-contains-forwarding-reference-as-its-argument-overloaded`
```
rvermeulen, line 1:
```suggestion
`A13-3-1` - `FunctionThatContainsForwardingReferenceAsItsArgumentOverloaded.ql`
```

Applying the style described in the developer handbook.

cpp/autosar/src/rules/A13-3-1/FunctionThatContainsForwardingReferenceAsItsArgumentOverloaded.ql
```diff
@@ -26,5 +26,7 @@ where
   not isExcluded(f,
     OperatorsPackage::functionThatContainsForwardingReferenceAsItsArgumentOverloadedQuery()) and
   not f.isDeleted() and
-  f = c.getAnOverload()
+  f = c.getAnOverload() and
+  //allow for overloading with different number of parameters
```
rvermeulen, line 30:
```suggestion
  // allow for overloading with different number of parameters, because there is no
  // confusion on what function will be called.
```

Adding reason for exception not described in the standard.

cpp/autosar/test/rules/A13-3-1/FunctionThatContainsForwardingReferenceAsItsArgumentOverloaded.expected
```diff
@@ -3,3 +3,5 @@
 | test.cpp:50:3:50:3 | A | Function overloads a $@ with a forwarding reference parameter. | test.cpp:47:3:47:3 | A | function |
 | test.cpp:63:8:63:8 | B | Function overloads a $@ with a forwarding reference parameter. | test.cpp:66:3:66:3 | B | function |
 | test.cpp:63:8:63:8 | B | Function overloads a $@ with a forwarding reference parameter. | test.cpp:66:3:66:3 | B | function |
+| test.cpp:71:7:71:7 | C | Function overloads a $@ with a forwarding reference parameter. | test.cpp:74:25:74:25 | C | function |
```
rvermeulen, line 6: Why does the compliant example add rows?

rvermeulen (changes requested):
Style adjustment and question on the new rows in the expected file.

cpp/autosar/test/rules/A13-3-1/FunctionThatContainsForwardingReferenceAsItsArgumentOverloaded.expected
rvermeulen: We want to exclude compiler generated functions such as the move and copy constructors that will match in parameter count.

knewbury01: I currently disagree, but may be misunderstanding the rule description

on page 239 of the standard, in the examples for this rule - there is an explicit example (a constructor in class A) where the only reason that the explicit declared constructor is COMPLIANT is via exception, where the exception is due to the the template param constraints, and otherwise would be overloading a compiler generated move constructor

I think the testcase added still describes a situation where it could be unclear which function is called, which is the purpose of the rule, however willing to do a reconsider !

rvermeulen: Per discussion we want to alert on the override with the forward reference parameters, instead of the copy/move ctors, because it doesn't apply SFINAE for which we have an exception. This would align with the example in the standard.

change_notes/2024-01-30-exclusion-a13-3-1.md
```diff
@@ -1,2 +1,2 @@
-`A13-3-1`: `cpp/autosar/function-that-contains-forwarding-reference-as-its-argument-overloaded`
+`A13-3-1`: `FunctionThatContainsForwardingReferenceAsItsArgumentOverloaded.ql`
```
rvermeulen, line 1:
```suggestion
`A13-3-1` - `FunctionThatContainsForwardingReferenceAsItsArgumentOverloaded.ql`:
```

cpp/autosar/src/rules/A13-3-1/FunctionThatContainsForwardingReferenceAsItsArgumentOverloaded.ql
```diff
@@ -21,12 +21,26 @@ class Candidate extends TemplateFunction {
...
+  then (
+    msg = "implicit constructor" and
+    overloaded = f and
+    overload = c
+  ) else (
+    msg = "function" and
+    overloaded = c and
+    overload = f
+  )
+select overload, "Function overloads a $@ with a forwarding reference parameter.", overloaded, msg
```
rvermeulen, line 46: The alert message requires some further explanation depending on the case.
1. "Function with forward reference parameter overloads implict constructor"
2. "Function overloads a ..."

rvermeulen: I was also expecting a change in the expected file due to this change in alert location.

rvermeulen (changes requested):
Minor format and alert message change request. Also was expecting a different expected file due to changes in the alert message.

cpp/autosar/src/rules/A13-3-1/FunctionThatContainsForwardingReferenceAsItsArgumentOverloaded.ql
knewbury01: yep!

cpp/autosar/test/rules/A13-3-1/test.cpp
```diff
@@ -61,9 +62,17 @@ A b(a);
...
 };
 
-int main() {}
\ No newline at end of file
+int main() {}
+
+class C {
+public:
+  C() {} // COMPLIANT by exception
+  template <typename T> C(T &&) {}
```
rvermeulen, line 77:
```suggestion
  C() {} 
  template <typename T> C(T &&) {} // COMPLIANT by exception
```

The alert is now on the move constructor so we need to adjust the COMPLIANT annotation.

cpp/autosar/src/rules/A13-3-1/FunctionThatContainsForwardingReferenceAsItsArgumentOverloaded.ql
```diff
@@ -21,10 +21,30 @@ class Candidate extends TemplateFunction {
...
+  f = c.getAnOverload() and
+  // allow for overloading with different number of parameters, because there is no
+  // confusion on what function will be called.
+  f.getNumberOfParameters() = c.getNumberOfParameters() and
+  //build a dynamic select statement that guarantees to read that the overloading function is the explicit one
+  if
+    (f instanceof CopyConstructor or f instanceof MoveConstructor) and
+    f.isCompilerGenerated()
+  then (
+    msg = "implicit constructor" and
```
rvermeulen, line 40: To provide the developer with better information we should state which construct it is. Now we will get two of the same alerts on the same location.

rvermeulen (changes requested):
Request to add more information into the alert. Would it be easy to turn the current FPs into TP and FN using a heuristic?

knewbury01:
> Request to add more information into the alert. Would it be easy to turn the current FPs into TP and FN using a heuristic?

I spent quite a bit of time trying to figure it out, no luck so far.

I have opened [another issue](https://github.com/github/codeql-coding-standards/issues/524) (and think there actually already was [one other](https://github.com/github/codeql-coding-standards/issues/406) to continue to investigate this) and have asked internally on how I can detect the `exclude_if` guard
so I think this PR sufficiently improves the alert message to be more understandable for now,
and further improvement can continue if it turns out modelling that guard is possible

@rvermeulen thanks for the check!

rvermeulen (approved):
Ready to merge!

PR #517, opened 2024-01-31
A16-0-1: exclusions for handling else and elif

### LINKED ISSUE #386: `A16-0-1`: Fix handling of `#else` and `#elif` in valid preprocessor conditionals
### Affected rules
 - `A16-0-1`

### Description

We erroneously identify otherwise acceptable conditionals with `#else` or `#elif` preprocessor directives as invalid, as the `#else` and `#elif` are not considered de facto valid.

This should be addressed by:
 * Always considering `#else` acceptable, and relying on reporting any issues in the `#..if` directive.
 * Considering `#elif` an acceptable directive within a preprocessor conditional, but still considering the `#elif` on its own merits.

### Example

```cpp
#ifdef MACRO_ENABLED
#include "test1.h"
#else
#include "test2.h"
#endif
```

knewbury01:
## Description

closes #386. 

The implementation here follows the exact recommendations in the issue. which are - 

Exclude all `#else` preprocessor directives from consideration (if a wrapper ifdef/ifndef is invalid then we only want to report at that location). We were also invalidly considering an `#elif` as a non permitted item in a valid `ifdef/ifndef` - so that is now also excluded.

there are still a couple of cases where behaviour may not be ideal according to the rule example. but this is unavoidable based on what is in the database when conditional compilation occurs, I have included them in the testcase simply as documentation of what occurs. here I add some annotations about what would be more ideal:
```
#ifdef OBJECTLIKE_MACRO_NO // NON_COMPLIANT - likely ideally should not be reported based on knowing what we have in the database, but we currently dont check what is the "closest" wrapper macro (ie this is reported bc of the int x = 1;, not the elif nor the int x = 0; ). although according to the rule description it would make sense to report this.
int x = 0;                 // not present
#elif OBJECTLIKE_MACRO > 0 // NON_COMPLIANT
int x = 1;  // present
#endif                     // COMPLIANT

#ifdef OBJECTLIKE_MACRO // NON_COMPLIANT
int x1 = 0;             // present
#elif OBJECTLIKE_MACRO >                                                       \
    -1 // COMPLIANT - by technicality of conditional compilation - according to the rule description this should be reported but we simply cannot
int x1 = 1; // not present
#endif // COMPLIANT
```

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - A16-0-1

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


change_notes/2024-01-31-exclusion-a16-0-1.md
```diff
@@ -0,0 +1,2 @@
+`A16-0-1`: `cpp/autosar/pre-processor-shall-only-be-used-for-certain-directives-patterns`
```
rvermeulen, line 1:
```suggestion
`A16-0-1` - `PreProcessorShallOnlyBeUsedForCertainDirectivesPatterns.ql`:
```

cpp/autosar/src/rules/A16-0-1/PreProcessorShallOnlyBeUsedForCertainDirectivesPatterns.ql
```diff
@@ -21,13 +21,26 @@ import cpp
...
 import codingstandards.cpp.FunctionLikeMacro
 
+class PermittedInnerDirectiveType extends PreprocessorDirective {
+  PermittedInnerDirectiveType() {
+    //permissive listing for directives that can be used in a valid wrapper
+    this instanceof MacroWrapper or
+    this instanceof PreprocessorEndif or
+    this instanceof Include or
+    this instanceof PermittedMacro or
+    this instanceof PreprocessorElif or
```
rvermeulen, line 31: To consider `elseif` separately we need to change the `getAGuard` predicate to slice up the range into appropriate ranges. That is from `if` to `elseif` and from `elseif` to `elseif` or `endif`.

We currently aren't capturing the case:

```
#ifdef MACRO_ENABLED_NON  // COMPLIANT
#include <string>         // COMPLIANT
#elif MACRO_ENABLED_OTHER // COMPLIANT
int x = 1;                // NON_COMPLIANT
#endif     
```

rvermeulen (changes requested):
Format change request and a possible corner case on how we deal with `elseif`

cpp/autosar/src/rules/A16-0-1/PreProcessorShallOnlyBeUsedForCertainDirectivesPatterns.ql
knewbury01: this one is complicated by the fact that the database will or will not have an element in it based on the outcome of the preprocessor conditional

I built a full enumeration of the testcases that cover all scenarios

however if I add this to the testcase, it has defs and undefs that make it into the alerts
unsure if it should still be added for explanation sake though?

```

// case 1 - first present only
#define MACRO_ENABLED_NON 1
#ifdef MACRO_ENABLED_NON  // COMPLIANT
#include <string>         //present
#elif MACRO_ENABLED_OTHER // NON_COMPLIANT[FALSE_NEGATIVE]
int x = 1;  // not present
#endif

// case 2 - second present only
#undef MACRO_ENABLED_NON
#define MACRO_ENABLED_OTHER 1
#ifdef MACRO_ENABLED_NON  // COMPLIANT
#include <string>         //not present
#elif MACRO_ENABLED_OTHER // NON_COMPLIANT
int x = 1;  // present
#endif

// case 3 - neither present
#undef MACRO_ENABLED_NON
#undef MACRO_ENABLED_OTHER
#ifdef MACRO_ENABLED_NON  // COMPLIANT
#include <string>         //not present
#elif MACRO_ENABLED_OTHER // NON_COMPLIANT[FALSE_NEGATIVE]
int x = 1;  // not present
#endif

// case 4 - both look present but the second still not bc the condition is not
// required to be evaluated
#define MACRO_ENABLED_NON 1
#define MACRO_ENABLED_OTHER 1
#ifdef MACRO_ENABLED_NON  // COMPLIANT
#include <string>         //present
#elif MACRO_ENABLED_OTHER // NON_COMPLIANT[FALSE_NEGATIVE]
int x = 1;  // not present
#endif
```

I have made an adjustment to the guard range logic though to avoid duplication where possible
so let me know if that now makes more sense

knewbury01: additionally now I am a little unsure of how to label this testcase
```
#ifdef OBJECTLIKE_MACRO_NO // COMPLIANT
int x0 = 0;                // not present
#elif OBJECTLIKE_MACRO > 0 // NON_COMPLIANT
int x0 = 1; // present
#endif                     // COMPLIANT
```
according to the examples in the spec `#ifdef OBJECTLIKE_MACRO_NO // COMPLIANT` should maybe be `#ifdef OBJECTLIKE_MACRO_NO // NON_COMPLIANT[FALSE_NEGATIVE] - known due to database not containing elements`

rvermeulen: I think it would be good to have the test cases in for documentation purposes.
To prevent the `#define` and `#undef` to be part of the results set you can use an option file to set the preprocessor defines like you would on a compiler invocation. You do have to change the cases because you can't define and undefine on the command line.

Yes, I agree that `OBJECTLIKE_MACRO_NO` is `// NON_COMPLIANT[FALSE_NEGATIVE] - known due to database not containing elements`

cpp/autosar/src/rules/A16-0-1/PreProcessorShallOnlyBeUsedForCertainDirectivesPatterns.ql
```diff
@@ -40,10 +53,18 @@ pragma[noinline]
...
-  exists(PreprocessorEndif end | pb.getEndIf() = end |
-    isPreprocFileAndLine(pb, filepath, startLine) and
-    isPreprocFileAndLine(end, filepath, endLine)
-  )
+  //the range of an if with an elif ends at the elif to avoid reporting things twice
+  if exists(PreprocessorElif elif | elif.getIf() = pb)
+  then
+    exists(PreprocessorElif end | end.getIf() = pb |
+      isPreprocFileAndLine(pb, filepath, startLine) and
+      isPreprocFileAndLine(end, filepath, endLine)
```
rvermeulen, line 61: I think we want the least range by looking at the closest `elseif` in case of multiple ones. Something like (not tested):

```suggestion
    isPreprocFileAndLine(pb, filepath, startLine) and
    endLine = min(int elifEndLine, PreProcessorElif end | isPreprocFileAndLine(end, filepath, elifEndLine) | elifEndLine)
```

cpp/autosar/src/rules/A16-0-1/PreProcessorShallOnlyBeUsedForCertainDirectivesPatterns.ql
knewbury01: awesome, thanks, didnt think of/know(?) about the command line flags, excellent suggestion!

cpp/autosar/src/rules/A16-0-1/PreProcessorShallOnlyBeUsedForCertainDirectivesPatterns.ql
knewbury01: great catch , this didnt quite associate the end item to the one above, but ended up simplifying the whole thing immensely!

cpp/autosar/src/rules/A16-0-1/PreProcessorShallOnlyBeUsedForCertainDirectivesPatterns.ql
```diff
@@ -40,10 +53,13 @@ pragma[noinline]
...
   PreprocessorBranch pb, string filepath, int startLine, int endLine
 ) {
-  exists(PreprocessorEndif end | pb.getEndIf() = end |
-    isPreprocFileAndLine(pb, filepath, startLine) and
-    isPreprocFileAndLine(end, filepath, endLine)
-  )
+  isPreprocFileAndLine(pb, filepath, startLine) and
+  endLine =
+    min(int elifEndLine, PreprocessorDirective end |
+      pb.getNext() = end and isPreprocFileAndLine(end, filepath, elifEndLine)
```
rvermeulen, line 59: I wasn't aware of `getNext`. This seems we don't need the `min` construct in that case.

rvermeulen (changes requested):
Final question on the range detection. With `getNext` I don't think we need the `min` construct. It is also better because of nested `elsif` constructs.

cpp/autosar/src/rules/A16-0-1/PreProcessorShallOnlyBeUsedForCertainDirectivesPatterns.ql
knewbury01: agree, nice catch

rvermeulen (approved):
Great stuff, lets :shipit:

PR #519, opened 2024-01-31
Fix FP reported in #467

### LINKED ISSUE #467: `M9-3-3`: Do not flag uninstantiated templates
### Affected rules
 - `M9-3-3`

### Description

Uninstantiated templates may not have resolved all the calls, which prevents this analysis from determining whether non-const functions are further called.

### Example

```cpp

template<typename T>
class Wrapper {
public:
  T& Top() { return elements.back(); } // Likely NON_COMPLIANT, but cannot be determined until instantiation.
private:
  vector<T> elements;
};

```

rvermeulen:
## Description

Fix #467 

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - M9-3-3

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


cpp/autosar/test/rules/M9-3-3/test.cpp
```diff
@@ -161,3 +161,29 @@ class Z22 : Z1 {
...
+private:
+  T data[128];
+  unsigned int size;
+};
+
+template <class T, template <class...> class U> class Stack {
+public:
+  T &Top() {
+    return this->data.back();
+  } // Likely NON_COMPLIANT, but cannot be determined until instantiation.
```
knewbury01, line 178: lets change this message to best describe the expected behaviour (align with the [style guide](https://github.com/github/codeql-coding-standards/blob/main/docs/development_handbook.md#style-guide)), 

in this case with the instantiation, it is a COMPLIANT example due to (after instantiation) being able to see that it calls a non const using member variable.

though this location represents 2 member functions, and one (the instantiated one) is compliant due to the logic of the rule
and the non instantiated one is compliant due to not having enough info to know if it obeys the rule - 

so maybe... label COMPLIANT - instantiated compliant truly and uninstantiated compliant via exception
or whatever seems most clear to describe the expected behaviour

knewbury01 (changes requested):
minor change requested to make the testcase slightly more clear 

(since the representations in the db are not the most clear around templates - ie this only shows up once the template is instantiated, but its the uninstantiated version of the function that was falsley triggering it 😅 )

cpp/autosar/test/rules/M9-3-3/test.cpp
rvermeulen: I would argue that the style guide is inconclusive about cases that are neither compliant nor non-compliant. Perhaps we should call it `COMPLIANT[FALSE_NEGATIVE|TRUE_NEGATIVE]` to highlight a nonstandard exception with the explanation that the template dictates the interface and not its instantiations?

knewbury01: I would be happy with either a `COMPLIANT - exclude consideration of functions in uninstantiated templates` or `COMPLIANT[FALSE_NEGATIVE|TRUE_NEGATIVE] - insert explanation` with the explanation you mention

its true that the latter doesnt completely follow the style guide, so I am leaning towards the former but would find either to communicate what is going on well enough (at least imo labelling it both fn/tn indicates that we are purposely ignoring it)

cpp/autosar/test/rules/M9-3-3/test.cpp
```diff
@@ -162,6 +162,35 @@ class Z22 : Z1 {
...
+  IntVectorStack s;
+
+  int i = s.Top();
+}
+
 class Z3 {
   void f(int) = delete; // COMPLIANT
-};
\ No newline at end of file
+};
```
knewbury01, line 195: needs a reformat, then it will be all good

knewbury01 (changes requested):

knewbury01 (approved):
nice work!

PR #520, opened 2024-02-01
Fix FP reported in #466

### LINKED ISSUE #466: `A7-1-2`: Consider `constexpr` function behaviour
### Affected rules
 - `A7-1-2`

### Description

A `constexpr` function only returns a compile-time constant when provided with compile-time constants as arguments. The query currently incorrectly considers `constexpr` functions to always return constants.

### Example

This function should only be considerd
```cpp
constexpr int add(int x, int y) {
  return x + y;
}

void test(int p) {
  int x = add(1,2); // NON_COMPLIANT - could be marked constexpr
  int y = add(1,p); // COMPLIANT - cannot be marked constexpr
```

rvermeulen:
## Description

Fixes #466 
## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - A7-1-2

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


cpp/autosar/src/rules/A7-1-2/VariableMissingConstexpr.ql
```diff
@@ -33,6 +33,15 @@ predicate isTypeZeroInitializable(Type t) {
   t.getUnderlyingType() instanceof ArrayType
 }
 
+predicate isCompileTimeEvaluated(Call call) {
+  call.getTarget().isConstexpr() and
+  forall(Expr arg | arg = call.getAnArgument() |
```
knewbury01, line 38: I think this logic needs another consideration

literals flowing into a var might not be sufficient

consider:
```
int a = 0;
  if (p> 0){
   a= 1;  
  }else{
    a =p+1; 
  }
    
  int x = add(a, 2); // cannot be made constexpr but currently being reported
```

if the literal comes from a constexpr var itself though then it is fine - like the following (actually also fine if a is not declared `constexpr` - but in this case it also could be):
```
  constexpr int a = 0;    
  int x = add(a, 2); // NON_COMPLIANT
```

knewbury01 (changes requested):
some logic requiring reconsideration I believe

cpp/autosar/src/rules/A7-1-2/VariableMissingConstexpr.ql
rvermeulen: Good catch. I definitely need to consider more corner cases.

cpp/autosar/test/rules/A7-1-2/test.cpp
```diff
@@ -204,4 +204,44 @@ class ExcludedCases {
...
+  int w = add1(x, 2); // NON_COMPLIANT
+  int a = add1(x);    // NON_COMPLIANT
+  int b = add1(1);    // NON_COMPLIANT
+  int c = add1(1, z); // COMPLIANT
+  int d = add1(1, z); // COMPLIANT
+  int e = add1(z);    // COMPLIANT
+  int f = add2(1);    // NON_COMPLIANT
+  int g = add2(1, 2); // NON_COMPLIANT
+  int h = add2(x, 2); // NON_COMPLIANT
+  int i = add2(x, 2); // NON_COMPLIANT
```
knewbury01, line 238: so I'll confess, I dont know the exact details of the constant propagation type logic that most compilers will use to checkup on if this can be `constexpr`... so I just added that specifier to all of the NON_COMPLIANT cases and the ones that disallow from the example compiling are : line 237,238,230,228 (maybe a few others)

additionally the readability of this section would benefit from some overhead comments that describe what each grouping exercises

knewbury01 (changes requested):

cpp/autosar/test/rules/A7-1-2/test.cpp
rvermeulen: I've added clarifications to what is supposed to be tested. The compiler errors are caused because the functions receive a compile time constant but the variable couldn't be marked `constexpr` because the intermediate variable holding the compile time constant wasn't a `constexpr`. 

Because of the resulting compilation errors I`ve opted to not present all the result to prevent confusion. It does mean that a change can result in multiple new results, but that is the trade off that has to be made.

knewbury01 (approved):
nice! 

I was naively thinking getting those extra testcases fairly accurate was as simple as - if another assignment (to anything) exists control flow upwards cut those out, but I definitely think that would also have not been as precise at what you implemented

very cool, think it looks good

PR #530, opened 2024-02-13
A18-0-1: improve query logic

### LINKED ISSUE #7: `A18-0-1`: Consider more than just the file name when identifying use of C standard libraries
### Affected rules
 - `A18-0-1`

### Description

The query for this rule reports any use of headers with _file names_ the same as a prohibited C standard library header. This can cause false positives if the included file is not from a C standard library implementation but just happens to have the same name as a C standard library header.

There's no certain way to determine whether an include is of a C Standard Library header file, because the files themselves are not universally distinguishable, so we will need to consider some heuristics for identification. 

As an initial idea, we could only report cases where:

 * The `#include` specifies no file path (e.g. `filename = i.getIncludeText().substring(1, i.getIncludeText().length() - 2)`)
 * And where the target file is not within the source location (e.g. `not exists(i.getIncludedFile().getRelativePath())`

### Example
```
$ cat lib/example.h 
#ifndef LIB_EXAMPLE_H_
#define LIB_EXAMPLE_H_

#endif

$ cat test.cpp 
#include "lib/example.h"  // A18-0-1 reported here
#include <iostream>
```

knewbury01:
## Description

closes #7 

implements the exact logic suggested in the issue. cannot think of any other heuristics to improve the query but think that these improvements should cover any additional cases that can be thought of at this time

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - A18-0-1

## Release change checklist

  - [ ] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 

### DIFF
--- a/change_notes/2024-02-12-improve-a18-0-1.md
+++ b/change_notes/2024-02-12-improve-a18-0-1.md
@@ -1,0 +1,2 @@
+`A18-0-1` - `CLibraryFacilitiesNotAccessedThroughCPPLibraryHeaders.ql`:
+    - Improve query logic to only match on exact standard library names (exclude local files with same names. Now excludes sys/header.h type headers as well from the results as those are not C standard libraries).
--- a/cpp/autosar/src/rules/A18-0-1/CLibraryFacilitiesNotAccessedThroughCPPLibraryHeaders.ql
+++ b/cpp/autosar/src/rules/A18-0-1/CLibraryFacilitiesNotAccessedThroughCPPLibraryHeaders.ql
@@ -28,12 +28,13 @@ where
    *    not use any of 'signal.h's facilities, for example.
    */
 
-  filename = i.getIncludedFile().getBaseName() and
+  filename = i.getIncludeText().substring(1, i.getIncludeText().length() - 1) and
   filename in [
       "assert.h", "ctype.h", "errno.h", "fenv.h", "float.h", "inttypes.h", "limits.h", "locale.h",
       "math.h", "setjmp.h", "signal.h", "stdarg.h", "stddef.h", "stdint.h", "stdio.h", "stdlib.h",
       "string.h", "time.h", "uchar.h", "wchar.h", "wctype.h"
-    ]
+    ] and
+  not exists(i.getIncludedFile().getRelativePath())
 select i,
   "C library \"" + filename + "\" is included instead of the corresponding C++ library <c" +
     filename.prefix(filename.length() - 2) + ">."
--- a/cpp/autosar/test/rules/A18-0-1/lib/assert.h
+++ b/cpp/autosar/test/rules/A18-0-1/lib/assert.h
@@ -1,0 +1,4 @@
+#ifndef LIB_EXAMPLE_H_
+#define LIB_EXAMPLE_H_
+
+#endif
--- a/cpp/autosar/test/rules/A18-0-1/test.cpp
+++ b/cpp/autosar/test/rules/A18-0-1/test.cpp
@@ -40,3 +40,5 @@
 #include <cuchar>    // COMPLIANT
 #include <cwchar>    // COMPLIANT
 #include <cwctype>   // COMPLIANT
+
+#include "lib/assert.h" // COMPLIANT


cpp/autosar/src/rules/A18-0-1/CLibraryFacilitiesNotAccessedThroughCPPLibraryHeaders.ql
```diff
@@ -28,12 +28,13 @@ where
...
 
-  filename = i.getIncludedFile().getBaseName() and
+  filename = i.getIncludeText().substring(1, i.getIncludeText().length() - 1) and
   filename in [
       "assert.h", "ctype.h", "errno.h", "fenv.h", "float.h", "inttypes.h", "limits.h", "locale.h",
       "math.h", "setjmp.h", "signal.h", "stdarg.h", "stddef.h", "stdint.h", "stdio.h", "stdlib.h",
       "string.h", "time.h", "uchar.h", "wchar.h", "wctype.h"
-    ]
+    ] and
+  not exists(i.getIncludedFile().getRelativePath())
```
rvermeulen, line 37: Excluding non-source files doesn't have a meaningful effect on the results in OpenPilot, but it would lead to FN in projects that include the C library standard headers. What are your thoughts on removing this heuristic?

change_notes/2024-02-12-improve-a18-0-1.md
```diff
@@ -0,0 +1,2 @@
+`A18-0-1` - `CLibraryFacilitiesNotAccessedThroughCPPLibraryHeaders.ql`:
```
rvermeulen, line 1:
```suggestion
- `A18-0-1` - `CLibraryFacilitiesNotAccessedThroughCPPLibraryHeaders.ql`:
```

rvermeulen (changes requested):
Minor format change request and question on not in source heuristic.

cpp/autosar/src/rules/A18-0-1/CLibraryFacilitiesNotAccessedThroughCPPLibraryHeaders.ql
knewbury01: I think this really depends on what is common practice, and I dont feel like I have enough experience of what is the most common case...

If a C library standard header is checked into a project, where is the most likely location that it will be located? and should we take that into consideration?

for example the `#include "lib/assert.h"` testcase that I added, assumes that this is not a std lib, by looking at the full path in the filename
but that wouldnt work for if its flat (now added a testcase for that)

I think we actually do want to permit both of the (new) testcases because we cannot assume if these are user made or not (or at least treat them the same regardless of where they are)

rvermeulen: I would assume they will still use `#include <assert.h>` or `#include "assert.h"` and not `#include "lib/assert.h"` when the stdlib is part of the source tree. So the location is not important and we can keep the path heuristic.

I'm not sure if more complicated heuristics for `#include "assert.h"` are worth it. For example, sampling definitions, because you are likely required to look at transitive includes given how these standard libraries are implemented and then you have to consider whether the definition was not introduced from the cpp header file (e.g. `include <cassert>`).

To conclude, I think the path heuristic is sufficient and works for both cases where the standard library is part of the source tree and when it is not.

knewbury01: asked for confirmation, but your assumptions make sense and I agree,

have reverted the one relative path heuristic and only use the name path portion

change_notes/2024-02-12-improve-a18-0-1.md
```diff
@@ -1,2 +1,2 @@
-`A18-0-1` - `CLibraryFacilitiesNotAccessedThroughCPPLibraryHeaders.ql`:
+- `A18-0-1` - `CLibraryFacilitiesNotAccessedThroughCPPLibraryHeaders.ql`:
     - Fix issue #7 - improve query logic to only match on exact standard library names (exclude local files with same names. Now excludes sys/header.h type headers as well from the results as those are not C standard libraries).
```
rvermeulen, line 2:
```suggestion
- `A18-0-1` - `CLibraryFacilitiesNotAccessedThroughCPPLibraryHeaders.ql`:
    - Fix issue #7 - improve query logic to only match on exact standard library names (e.g., now excludes sys/header.h type headers from the results as those are not C standard libraries).
```

rvermeulen (changes requested):
Update request for the changenote to reflect we no longer exclude local files. Otherwise all looks good 👍

rvermeulen (approved):
🚢

PR #532, opened 2024-02-13
Fix FP reported in 424

### LINKED ISSUE #424: `A15-4-4`: Conservatively assuming throwing behaviour for externally defined functions
### Affected rules
 - `A15-4-4`
### Description

The exceptions library makes a conservative assumption about which functions will throw which exceptions. However, for this query that leads to false positives where an externally defined function is known to throw in certain circumstances, but this is not captured in the throw specification for the function.

### Example
This function is erroneously identified as a candidate for `noexcept` even though `append` could throw an `out_of_range` error.
```cpp
std::string concat(const std::string& s1, const std::string& s2) {
  std::string s3;
  s3.reserve(s1.size() +s2.size());
  s3.append(s1.c_str(), s1.size())
  s3.append(s2.c_str(), s2.size()));
  return s3;
}```

### PR DESCRIPTION
**Pull request by rvermeulen**
## Description

Resolves #424 

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - _rule number here_

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 

### DIFF
--- a/cpp/common/src/codingstandards/cpp/exceptions/ExceptionFlowCustomizations.qll
+++ b/cpp/common/src/codingstandards/cpp/exceptions/ExceptionFlowCustomizations.qll
@@ -1,8 +1,49 @@
+/*
+ * A library customize models that model the flow of exceptions through the program.
+ */
+
 import cpp
 private import codingstandards.cpp.exceptions.ExceptionFlow
 
 /** A `ThrowingExpr` which is the origin of a exceptions in the program. */
 abstract class OriginThrowingExpr extends ThrowingExpr { }
+
+/**
+ * A `FunctionCall` to an external function without an exception specification that *
+ *  may throw an exception.
+ */
+abstract class ExternalUnderspecifiedFunctionCallThrowingExpr extends FunctionCall, ThrowingExpr { }
+
+/**
+ * An extensible predicate that describes functions that when called may throw an exception.
+ */
+extensible predicate throwingFunctionModel(
+  string functionNamespaceQualifier, string functionTypeQualifier, string functionName,
+  string exceptionNamespaceQualifier, string exceptionType
+);
+
+/**
+ * A `FunctionCall` that may throw an exception of type `ExceptionType` as provded by
+ * the extensible predicate `throwingFunctionModel`.
+ */
+private class ExternalFunctionCallThrowingExpr extends FunctionCall, ThrowingExpr {
+  ExceptionType exceptionType;
+
+  ExternalFunctionCallThrowingExpr() {
+    exists(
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: cpp/common/src/codingstandards/cpp/exceptions/ExceptionFlowCustomizations.qll [hunk 1]
```diff
@@ -0,0 +1,113 @@
...
+private import codingstandards.cpp.exceptions.ExceptionFlow
+
+/** A `ThrowingExpr` which is the origin of a exceptions in the program. */
+abstract class OriginThrowingExpr extends ThrowingExpr { }
+
+/**
+ * A `FunctionCall` to an external function without an exception specification that *
+ *  may throw an exception.
+ */
+abstract class ExternalUnderspecifiedFunctionCallThrowingExpr extends FunctionCall, ThrowingExpr { }
```
knewbury01, line 15: I dont quite see the purpose of this addition (was this meant to be left in here, since you are actually using the extensible models approach instead?)

cpp/common/src/ext/stdc++.model.yml
```diff
@@ -0,0 +1,7 @@
+extensions:
+  - addsTo:
+      pack: codeql/common-cpp-coding-standards
+      extensible: throwingFunctionModel
+    data:
+      - ["std", "basic_string", "append", "std", "out_of_range"]
```
knewbury01, line 6: just to check, do we want to add anything else here?
cant think of a good systematic way to scrape these out of the cpp docs
but for example: [replace?](https://en.cppreference.com/w/cpp/string/basic_string/replace)
but there doesnt seem to be a top down list, [exception types say that they may be thrown by some members](https://en.cppreference.com/w/cpp/error/out_of_range) but not a full list of which...

knewbury01 (commented):
just a couple of small questions

cpp/common/src/codingstandards/cpp/exceptions/ExceptionFlowCustomizations.qll
rvermeulen: This is here so it can be extended from QL as well if needed. Model as data is the preferred route, but has limitations due to its expressiveness.

cpp/common/src/ext/stdc++.model.yml
rvermeulen: For this PR I wanted to limit it to just these cases.

PR #538, opened 2024-02-16
Resolve 371

### LINKED ISSUE #371: `A5-1-1`: Exclude literals generated for template instantiations
### Affected rules
 - `A5-1-1`

### Description

Template instantiations with literal arguments can cause the creation of generated `Literal`s in the program in two ways:

 1. If the template argument was a `constexpr`, we may generate a `Literal` expression to represent the compiler-calculated value.
 2. If the template argument is used as an expression, we will generate a `Literal` at that location to hold the value of the compiler-calculated value.

### Example

```cpp
#include <limits>
constexpr std::size_t dynamic_extent = std::numeric_limits<std::size_t>::max();

template <typename _Type, std::size_t _Extent = dynamic_extent> class span {
  static constexpr std::size_t extent = _Extent;
};

void test_extent() { span<int> s; }
```

rvermeulen:
## Description

- Resolve #371.
- Exclude literals used to instantiate a template class.
- Reformat the alert message to adhere to the style guide.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - A5-1-1

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


cpp/common/src/codingstandards/cpp/Literals.qll
```diff
@@ -28,3 +28,22 @@ class Utf16StringLiteral extends StringLiteral {
 class Utf32StringLiteral extends StringLiteral {
   Utf32StringLiteral() { this.getValueText().regexpMatch("(?s)\\s*U\".*") }
 }
+
+/**
+ * A literal resulting from the use of a constexpr
+ * variable, or macro expansion.
+ */
+class CompileTimeComputedIntegralLiteral extends Literal {
```
knewbury01, line 36: hm was looking at a (I think) similarly [concepted CodeQL query](https://github.com/github/codeql/blob/main/cpp/ql/src/Best%20Practices/Magic%20Constants/MagicConstants.qll#L75C45-L75C59) - should we be caring about floating point types also?

cpp/common/src/codingstandards/cpp/Literals.qll
```diff
@@ -28,3 +28,22 @@ class Utf16StringLiteral extends StringLiteral {
...
+  CompileTimeComputedIntegralLiteral() {
+    this.getUnspecifiedType() instanceof IntegralType and
+    not this.getUnspecifiedType() instanceof BoolType and
```
knewbury01, line 39: oh , this also deprecates/duplicates some of the individual exclusions that were used in the query - like 
```
  // Exclude character literals - I think this one can now be removed based on your regex? double check me on that!
  not l instanceof CharLiteral and 
  // Exclude `nullptr`
  not l.getType() instanceof NullPointerType and
  // Exclude boolean `true` and `false` - this can definitely now be removed
  not l.getType() instanceof BoolType and
```

cpp/common/src/codingstandards/cpp/Literals.qll
```diff
@@ -28,3 +28,22 @@ class Utf16StringLiteral extends StringLiteral {
...
+    not this.getUnspecifiedType() instanceof CharType and
```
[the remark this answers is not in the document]
rvermeulen: Good catch. I will add those character types.

cpp/common/src/codingstandards/cpp/Literals.qll
rvermeulen: Not at this point. Floating points cannot be used as [template non type parameters](https://en.cppreference.com/w/cpp/language/template_parameters) for C++ 14. This has been added for C++ 20 so I will add it as a comment.

cpp/common/src/codingstandards/cpp/Literals.qll
rvermeulen: No, we are stating that it should not be a compile time computed value, which doesn't include char literals and booleans. So if also don't want those, we need to keep the above conditions.

cpp/common/src/codingstandards/cpp/Literals.qll
rvermeulen: The other use cases, where we have `#define delta 0.1f` is already covered by the query excluding macro calls.

cpp/common/src/codingstandards/cpp/Literals.qll
knewbury01: er Im not saying to omit them, I am saying [they are duplicated checks](https://github.com/github/codeql-coding-standards/pull/538/files#diff-6bbf6c525fda0e915ad2ae7af53c73ad2da5f2febc18cbf59349808dc138eaa0R42) (the snip I pasted was from the query file, but I did not link the line number, my bad)

rvermeulen: Not sure I'm following. The `CompileTimeComputedIntegralLiteral` class is intended to find cases where a `constexpr int foo = 1;` variable or macro `#define FOO 1` is used in a template instantiation. While there is some duplication in the query and the library predicate, it is still needed if the predicate is used outside the query at some point.

What are you suggesting to do with the duplicate checks?

knewbury01: nvm, I was blanking on thinking that they were redundant in the query, but they are not! I was thinking bc `CompileTimeComputedIntegralLiteral` is never a bool, that the literal in the query was never going to be a bool but that is not accurate logic bc not all literals (bools included) are `CompileTimeComputedIntegralLiteral` to start! my bad

knewbury01 (approved):
lgtm!

PR #545, opened 2024-02-26
M5-14-1: exclusion unevaluated contexts

### LINKED ISSUE #375: `M5-14-1`: Exclude unevaluated contexts
### Affected rules
 - `M5-14-1`

### Description

Exclude results in unevaluated contexts, because in practice they will never cause side-effects, and the rationale behind the rule is to report cases which may or may not cause side effects.

### Example

```cpp

```

knewbury01:
## Description

fixes #375 

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - M5-14-1

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 

### DIFF
--- a/cpp/autosar/src/rules/M5-14-1/RightHandOperandOfALogicalAndOperatorsContainSideEffects.ql
+++ b/cpp/autosar/src/rules/M5-14-1/RightHandOperandOfALogicalAndOperatorsContainSideEffects.ql
@@ -18,6 +18,7 @@ import cpp
 import codingstandards.cpp.autosar
 import codingstandards.cpp.SideEffect
 import codingstandards.cpp.sideeffect.DefaultEffects
+import codingstandards.cpp.Expr
 
 from BinaryLogicalOperation op, Expr rhs
 where
@@ -25,6 +26,5 @@ where
     SideEffects1Package::rightHandOperandOfALogicalAndOperatorsContainSideEffectsQuery()) and
   rhs = op.getRightOperand() and
   hasSideEffect(rhs) and
-  not rhs.(NoExceptExpr).getExpr().isUnevaluated() and
-  not rhs.(SizeofExprOperator).getExprOperand().isUnevaluated()
+  not rhs instanceof UnevaluatedExprExtension
 select op, "The $@ may have a side effect that is not always evaluated.", rhs, "right-hand operand"
--- a/cpp/autosar/src/rules/M5-14-1/RightHandOperandOfALogicalAndOperatorsContainSideEffects.ql
+++ b/cpp/autosar/src/rules/M5-14-1/RightHandOperandOfALogicalAndOperatorsContainSideEffects.ql
@@ -19,10 +19,39 @@ import codingstandards.cpp.autosar
 import codingstandards.cpp.SideEffect
 import codingstandards.cpp.sideeffect.DefaultEffects
 
+/**
+ * an operator that does not evaluate its operand
+ * `decltype` also has a non evaluated operand but cannot be used in a `BinaryLogicalOperation`
+ */
+class UnevaluatedOperand extends Expr {
+  Expr operator;
+  UnevaluatedOperand() {
+    exists(SizeofExprOperator op | op.getExprOperand() = this |
+      not this.getUnderlyingType().(ArrayType).hasArraySize()
+      and operator = op
+    )
+    or
+    exists(NoExceptExpr e | e.getExpr() = this
+    and operator = e)
+    or 
+    exists(TypeidOperator t | t.getExpr() = this
+    and operator = t)
+    or 
+    exists(FunctionCall declval | declval.getTarget().hasQualifiedName("std", "declval")
+    and declval.getAChild() = this
+    and operator = declval)
+  }
+
+  Expr getOp(){
+    result = operator
+  }
+}
+
 from BinaryLogicalOperation op, Expr rhs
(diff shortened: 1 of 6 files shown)


cpp/autosar/src/rules/M5-14-1/RightHandOperandOfALogicalAndOperatorsContainSideEffects.ql
```diff
@@ -19,10 +19,44 @@ import codingstandards.cpp.autosar
 import codingstandards.cpp.SideEffect
 import codingstandards.cpp.sideeffect.DefaultEffects
 
+/**
+ * an operator that does not evaluate its operand
+ * `decltype` also has a non evaluated operand but cannot be used in a `BinaryLogicalOperation`
+ */
+class UnevaluatedOperand extends Expr {
```
rvermeulen, line 26: There is an opportunity for reuse using [isUnevaluated](https://codeql.github.com/codeql-standard-libraries/cpp/semmle/code/cpp/exprs/Expr.qll/predicate.Expr$Expr$isUnevaluated.0.html). 

Perhaps we can create an extended predicate in our `Expr` module that adds the `decval` case in addition to the cases covered by `isUnevaluated`.

cpp/autosar/test/rules/M5-14-1/test.cpp
```diff
@@ -23,4 +23,14 @@ void f3(bool b) {
...
   if (b || f2()) { // COMPLIANT, f2 has local side-effects
   }
+}
+
+int g1 = 0;
+int f4() { return g1++; }
+int f5() { return 1; }
+
+void f6() {
+  if (noexcept(f5()) &&noexcept(
```
rvermeulen, line 33: Can we cover the other cases as well to ensure we properly handle those?

rvermeulen (changes requested):
Some minor change request. There is some room for reuse of an existing standard library predicate that allows use to reduce our code. 

Improvement of coverage in our unit test.

Other than that all looks good!

cpp/autosar/test/rules/M5-14-1/test.cpp
knewbury01: great suggestion, this let me know that actually using decval that way is also not possible! good catch!

cpp/autosar/src/rules/M5-14-1/RightHandOperandOfALogicalAndOperatorsContainSideEffects.ql
knewbury01: great catch! cant believe I missed this, oh I actually think that case is not relevant now, cannot create a testcase that would use that in the conditional (think its not possible)

cpp/autosar/src/rules/M5-14-1/RightHandOperandOfALogicalAndOperatorsContainSideEffects.ql
```diff
@@ -19,44 +19,12 @@ import codingstandards.cpp.autosar
...
-
 from BinaryLogicalOperation op, Expr rhs
 where
   not isExcluded(op,
     SideEffects1Package::rightHandOperandOfALogicalAndOperatorsContainSideEffectsQuery()) and
   rhs = op.getRightOperand() and
   hasSideEffect(rhs) and
-  not exists(UnevaluatedOperand un | un.getOp() = rhs)
+  not rhs.(NoExceptExpr).getExpr().isUnevaluated() and
+  not rhs.(SizeofExprOperator).getExprOperand().isUnevaluated()
```
rvermeulen, line 29: There is no need to check whether the expressions of `NoExceptExpr` and `SizeofExprOperator` are unevaluated. That is always the case.

I think it is best to reinstate `UnevalutedOperand` and replace all but the `declval` case with `isUnevaluated` on the child of `rhs`.

We can then exclude the cases by asking if `not  rhs instanceof UnevaluatedOperand`

rvermeulen (changes requested):
I think we should include `isUnevaluated` as part of the `UnevaluatedOperand` and place that in the `Expr` module for reuse.

cpp/autosar/src/rules/M5-14-1/RightHandOperandOfALogicalAndOperatorsContainSideEffects.ql
knewbury01: agree! good suggestions

cpp/common/src/codingstandards/cpp/Expr.qll
```diff
@@ -189,3 +189,17 @@ module MisraExpr {
     CValue() { isCValue(this) }
   }
 }
+
+/**
+ * an operator that does not evaluate its operand
+ */
+class UnevaluatedExprExtension extends Expr {
+  UnevaluatedExprExtension() {
+    this.getChild(_).isUnevaluated()
```
rvermeulen, line 198:
```suggestion
    this.getAChild().isUnevaluated()
```

rvermeulen (changes requested):
Last bit of nitpicking. Otherwise good to ship!

rvermeulen (approved):
:shipit:

PR #546, opened 2024-02-27
A2-10-1: add functions and types to identifier consideration

### LINKED ISSUE #118: `A2-10-1`: Report type and function identifier hiding
### Affected rules
 - `A2-10-1`

### Description

`A2-10-1` states "An identifier declared in an inner scope shall not hide an identifier declared in an outer scope.". The existing query looks for user defined variables which cause hiding, but functions and types can also be hidden (and cause hiding). 

### Example
```cpp
int a;
namespace b {
        void a() { } // NON_COMPLIANT
        int c() { return a; }
}
```

knewbury01:
## Description

fixes #118 

the query seems to maybe be slower than before, will need this confirmed

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - A2-10-1
     - RULE-5-3

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


rvermeulen (changes requested):
The generalization of identifiers needs some further consideration, because of the increased complexity and incomplete exclusions part of the first iteration.

Possible FN are class members where a derived class member hides a base class member. It is possibly because the outer/inner scope definition for this rule doesn't mention it, but it successor in Misra C++ 23 does include that explicitly in the definition of the outer/inner scope.

Our `hides` logic now has to consider functions and their overloads. The query itself uses the strict version that requires a different scope which excludes this as problem. Not sure if we want to correct the non-strict version. 

Given that the Misra C/C++ 23 version of this rule, rule 6.4.1 and 6.4.2, focuses solely on variables and functions, I wonder if we want to do the same here and exclude types.

cpp/common/src/codingstandards/cpp/Scope.qll
```diff
@@ -204,18 +212,9 @@ class TranslationUnit extends SourceFile {
...
-}
-
-/** Holds if `v2` may hide `v1`. */
-private predicate hides_candidateStrict(UserVariable v1, UserVariable v2) {
-  not v1 = v2 and
-  v2 = getPotentialScopeOfVariableStrict(v1) and
+  v2 = getPotentialScopeOfDeclarationStrict(v1) and
   v1.getName() = v2.getName() and
   // Member variables cannot hide other variables nor be hidden because the can be referenced through their qualified name.
   not (v1.isMember() or v2.isMember()) and
```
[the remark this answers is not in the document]
knewbury01: fixed the first one,

for the second one, having a bit of trouble - this wasnt actually an FP , it was an FN
I dont understand lambdas' scopes...
for some reason the body block of a function is not showing up as the ancestor scope of a lambda:
```
void f4() {
  int a1, b;
  auto lambda1 = [a1]() {
    int b = 10; // COMPLIANT - exception - non captured variable b
  };

  auto lambda2 = [b]() {
    int b = 10; // NON_COMPLIANT[FALSE_NEGATIVE] - not an exception - captured variable b
  };
}
```
`s.getAnAncestor+()` ([from here](https://github.com/github/codeql-coding-standards/blob/main/cpp/common/src/codingstandards/cpp/Scope.qll#L148)) quick eval'd

shows  - 
```

{ ... } | decltype([...](...){...})
-- | --
43 | { ... } | operator()
44 | { ... } | (unnamed constructor)
45 | { ... } | decltype([...](...){...})


```
none of which are the f4 function body or anything else upwards of that

knewbury01:
> The generalization of identifiers needs some further consideration, because of the increased complexity and incomplete exclusions part of the first iteration.
> 
> Possible FN are class members where a derived class member hides a base class member. It is possibly because the outer/inner scope definition for this rule doesn't mention it, but it successor in Misra C++ 23 does include that explicitly in the definition of the outer/inner scope.

this is covered in  [A10-2-1](https://github.com/github/codeql-coding-standards/blob/main/cpp/autosar/src/rules/A10-2-1/NonVirtualPublicOrProtectedFunctionsRedefined.ql) - so we should be ok to not worry about this exact case? 

> 
> Our `hides` logic now has to consider functions and their overloads. The query itself uses the strict version that requires a different scope which excludes this as problem. Not sure if we want to correct the non-strict version.
> 

currently the only use of `hides` is [here](https://github.com/github/codeql-coding-standards/blob/main/cpp/cert/src/rules/DCL53-CPP/LocalConstructorInitializedObjectHidesIdentifier.ql#L23), where that wasnt relevant, but I omitted the overloads nonetheless, assuming that overloads are always intentional (not hiding), which I think sounds reasonable?

> Given that the Misra C/C++ 23 version of this rule, rule 6.4.1 and 6.4.2, focuses solely on variables and functions, I wonder if we want to do the same here and exclude types.

done, great catch (didnt own the 23 version til this comment)

cpp/common/src/codingstandards/cpp/Scope.qll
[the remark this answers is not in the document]
knewbury01: nevermind, was able to write something to cover both cases!

cpp/common/src/codingstandards/cpp/Scope.qll
```diff
@@ -57,10 +57,18 @@ private Element getParentScope(Element e) {
...
+
+/** A construct which is defined by the user, rather than being from a third party or compiler generated. */
+class UserDeclaration extends Declaration {
+  UserDeclaration() {
     exists(getFile().getRelativePath()) and
-    not isCompilerGenerated() and
+    not this.(Variable).isCompilerGenerated() and
+    not this.(Function).isCompilerGenerated() and
     not this.(Parameter).getFunction().isCompilerGenerated() and
+    // will falsely conflict
```
rvermeulen, line 70:
```suggestion
    // Class template instantiations are compiler generated instances that share the same parent scope. This will result in a cross-product on class template instantiations because they have the same name and same parent scope. We therefore exclude these from consideration like we do with other compiler generated identifiers of interest.
```

cpp/common/src/codingstandards/cpp/Scope.qll
```diff
@@ -257,20 +267,22 @@ private Stmt getEnclosingStmt(LocalScopeVariable v) {
...
+predicate hides(UserDeclaration v1, UserDeclaration v2) {
   hides_candidate(v1, v2) and
   // Confirm that there's no closer candidate variable which `v2` hides
-  not exists(UserVariable mid |
+  not exists(UserDeclaration mid |
     hides_candidate(v1, mid) and
     hides_candidate(mid, v2)
-  )
+  ) and
+  //ignore intentional overloads
```
rvermeulen, line 277:
```suggestion
  // Unlike `hidesStrict`, that requires a different scope, `hides` considers declarations in the same scope. This will include function overloads based on their name. To remove overloads from consideration, we exclude them.
```

cpp/common/src/codingstandards/cpp/rules/identifierhidden/IdentifierHidden.qll
```diff
@@ -11,13 +11,36 @@ abstract class IdentifierHiddenSharedQuery extends Query { }
 
 Query getQuery() { result instanceof IdentifierHiddenSharedQuery }
 
-query predicate problems(UserVariable v2, string message, UserVariable v1, string varName) {
+/**
+ * There is a lambda that contains a declaration
+ * that hides something that is captured
+ * and the lambda exists in the function where this lamda is enclosed
+ */
+predicate hiddenInLambda(UserDeclaration v2, UserDeclaration v1) {
```
rvermeulen, line 19: To make it consistent with `hidesStrict` we should swap the parameters.
`v2` hides `v1`.

```suggestion
predicate hiddenInLambda(UserDeclaration v1, UserDeclaration v2) {
```

cpp/common/src/codingstandards/cpp/rules/identifierhidden/IdentifierHidden.qll
```diff
@@ -11,13 +11,36 @@ abstract class IdentifierHiddenSharedQuery extends Query { }
...
+  exists(Scope s, Closure le |
+    s.getADeclaration() = v2 and
+    s.getAnAncestor() = le and
+    le.getEnclosingFunction().getBasicBlock().(Scope) = v1.getParentScope() and
```
rvermeulen, line 23: Why are we looking at the basic block of the function? I believe this works because it coincides with the `BlockStmt` of the function.
Would it be better to look at the transitive parent scope of the closure or does that not have a parent scope that works?

cpp/common/src/codingstandards/cpp/rules/identifierhidden/IdentifierHidden.qll
rvermeulen: I think overall using of more verbose variables help with understanding the predicate.

```ql
predicate hiddenInLambda(UserDeclaration outerDecl, UserDeclaration innerDecl) {
  exists(Scope scopeOfInnerDecl, Closure closureWithInnerDecl|
    scopeOfInnerDecl.getADeclaration() = innerDecl and
    scopeOfInnerDecl.getAnAncestor() = closureWithInnerDecl and
    closureWithInnerDecl.getEnclosingFunction().getBasicBlock().(Scope) = outerDecl.getParentScope() and
    exists(LambdaCapture cap, Variable v |
      v.getAnAccess() = cap.getInitializer().(VariableAccess) and
      v = outerDecl and
      closureWithInnerDecl.getLambdaExpression().getACapture() = cap
    ) and
    innerDecl.getName() = outerDecl.getName()
  )
}
```

rvermeulen: Since we are limiting to variable in the `exists` it would be a performance benefit to change the parameters to `UserVariable` to potentially reduce the number of starting tuples.

cpp/common/src/codingstandards/cpp/rules/identifierhidden/IdentifierHidden.qll
```diff
@@ -11,13 +11,36 @@ abstract class IdentifierHiddenSharedQuery extends Query { }
 
 Query getQuery() { result instanceof IdentifierHiddenSharedQuery }
 
-query predicate problems(UserVariable v2, string message, UserVariable v1, string varName) {
+/**
+ * There is a lambda that contains a declaration
+ * that hides something that is captured
+ * and the lambda exists in the function where this lamda is enclosed
```
rvermeulen, line 17:
```suggestion
 * Holds if declaration `v2`, declared in a lambda, hides a declaration `v1` captured by the lambda.
```

Not that this is based on the current version. If the suggestions below are considered, we need to adjust this accordingly.

cpp/common/src/codingstandards/cpp/rules/identifierhidden/IdentifierHidden.qll
```diff
@@ -11,13 +11,36 @@ abstract class IdentifierHiddenSharedQuery extends Query { }
...
+}
+
+query predicate problems(UserDeclaration v2, string message, UserDeclaration v1, string varName) {
   not isExcluded(v1, getQuery()) and
   not isExcluded(v2, getQuery()) and
   //ignore template variables for this rule
   not v1 instanceof TemplateVariable and
   not v2 instanceof TemplateVariable and
-  hidesStrict(v1, v2) and
+  //ignore types for this rule
```
rvermeulen, line 39:
```suggestion
  //ignore types for this rule
```
Can we also document why for future reference?

cpp/common/src/codingstandards/cpp/Scope.qll
```diff
@@ -287,3 +299,18 @@ predicate hasClassScope(Declaration decl) { exists(decl.getDeclaringType()) }
 
 /** Holds if `decl` has block scope. */
 predicate hasBlockScope(Declaration decl) { exists(BlockStmt b | b.getADeclaration() = decl) }
+
+/**
+ * identifiers in nested (named/nonglobal) namespaces are exceptions to hiding due to being able access via fully qualified ids
+ */
+predicate excludedViaNestedNamespaces(UserDeclaration v2, UserDeclaration v1) {
```
rvermeulen, line 306:
```suggestion
predicate excludedViaNestedNamespaces(UserDeclaration v1, UserDeclaration v2) {
```
Can we use the same order as `hides`, where `v2` hides `v1`.

Better names would be `outerDecl` and `innerDecl` where `innerDecl` hides the `outerDecl`.

rvermeulen (changes requested):
I left a few considerations for the lambda case and proposal for more reorder the parameters for consistency.

cpp/common/src/codingstandards/cpp/rules/identifierhidden/IdentifierHidden.qll
knewbury01: the scope check here is to say - the innerDecl is in a scope inside the lambda, its not for the capture (but this is hopefully more clear now with the var name improvements 😅 )

not sure I understand the case - "The situation becomes more tricky for variables that accessible without a capture." 
if the variable is not captured, then there is no hiding, reference to the compliant exception example in the standard

cpp/common/src/codingstandards/cpp/rules/identifierhidden/IdentifierHidden.qll
knewbury01: "Would it be better to look at the transitive parent scope..."
yeah that sounds more ideal but I wasnt able to get that to work 😞 

quick eval for `le.getParentScope()` has no result

tldr the scopes in lambdas werent as I naively expected, but if you see a better way to do this I could totally be missing something obvious!

cpp/common/src/codingstandards/cpp/rules/identifierhidden/IdentifierHidden.qll
rvermeulen: A lambda, can for example refer to a global variable without capturing that.

```
int foo = 0;

int bar() {
    auto lambda_with_shadowing = []() {
        int foo = 1;
        return foo + foo;
    };

    auto lambda_without_shadowing = []() {
        return foo + foo;
    };

    return lambda_with_shadowing();
}
```

I would argue that `lambda_with_shadowing` is non compliant.

knewbury01: aha, you are correct ,

I have added cases for the description [here](https://en.cppreference.com/w/cpp/language/lambda) when a lambda can access a variable without explicitly capturing it

cpp/common/src/codingstandards/cpp/rules/identifierhidden/IdentifierHidden.qll
knewbury01: [your addition](https://github.com/knewbury01/codeql-coding-standards/pull/2) to this model deprecated this comment!

cpp/common/src/codingstandards/cpp/Scope.qll
```diff
@@ -122,31 +132,31 @@ class GeneratedBlockStmt extends BlockStmt {
...
       not result = v and
       v.getName() <= result.getName()
     )
   )
 }
 
-/** Gets a variable that is in the potential scope of variable `v`. */
-private UserVariable getOuterScopesOfVariable_candidate(UserVariable v) {
+/** Gets a Declarationthat is in the potential scope of Declaration `v`. */
+private UserDeclaration getOuterScopesOfDeclaration_candidate(UserDeclaration v) {
```
rvermeulen, line 153:
```suggestion
private UserDeclaration getPotentialScopeOfDeclarationStrict_candidate(UserDeclaration v) {
```

I missed this being renamed, but we should keep the previous name.

cpp/common/src/codingstandards/cpp/Scope.qll
```diff
@@ -161,20 +171,20 @@ predicate inSameTranslationUnit(File f1, File f2) {
...
 
 /**
- * Gets a user variable which occurs in the "potential scope" of variable `v`.
+ * Gets a user Declaration which occurs in the "outer scope" of Declaration `v`.
  */
 cached
-UserVariable getPotentialScopeOfVariable(UserVariable v) {
-  result = getPotentialScopeOfVariable_candidate(v) and
+UserDeclaration getPotentialScopeOfDeclarationStrict(UserDeclaration v) {
+  result = getOuterScopesOfDeclaration_candidate(v) and
```
rvermeulen, line 178:
```suggestion
  result = getPotentialScopeOfDeclarationStrict_candidate(v) and
```

cpp/common/src/codingstandards/cpp/Scope.qll
rvermeulen: I noticed the previous name was also outer scope ..., not sure why.

rvermeulen:
With addition of other declarations the performance of this query has significantly decreased. My current hypothesis is that the number of declarations in the global namespace explodes the size of the `getOuterScopesOfDeclaration_candidate` relation that requires some further scrutinizing.

cpp/common/src/codingstandards/cpp/Scope.qll
knewbury01: hmm yep, good catch, also not sure why!

rvermeulen (approved):
LGTM.

PR #548, opened 2024-03-01
Fix 152: Adding query for depencence on operator precedence

### LINKED ISSUE #152: `M5-0-2`: Missing support for inadequate use of parenthesis
### Affected rules
 - `M5-0-2`

### Description

Our current query identifies gratuitous parentheses i.e. parentheses that are not strictly required. However, the rule is actually "Limited dependence should be placed on C++ operator precedence rules in expressions." In other words, we should be identifying cases where a lack of parentheses causes ambiguity.

rvermeulen:
## Description

Resolve #152.

Adding query to find operands of binary operations and ternary operators that should be parenthesized.
The query does not implement the case where associativity may compute a different result due to integer promotions. 

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - M5-0-2
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 

### DIFF
--- a/cpp/autosar/src/rules/M5-0-2/InsufficientUseOfParentheses.ql
+++ b/cpp/autosar/src/rules/M5-0-2/InsufficientUseOfParentheses.ql
@@ -20,9 +20,10 @@ import semmle.code.cpp.commons.Assertions
 
 class InsufficientlyParenthesizedExpr extends Expr {
   InsufficientlyParenthesizedExpr() {
-    // Exclude assertions because if the child is the expression being asserted it
-    // is not necessary to add parenthesis.
-    not any(Assertion a).getAsserted() = this and
+    // Exclude expressions affected by macros, including assertions, because
+    // it is unclear that the expression must be parenthesized since it seems
+    // to be the top-level expression instead of an operand of a binary or ternary operation.
+    not this.isAffectedByMacro() and
     (
       exists(BinaryOperation root, BinaryOperation child | child = this |
         root.getAnOperand() = child and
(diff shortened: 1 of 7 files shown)


cpp/autosar/src/rules/M5-0-2/InsufficientUseOfParentheses.ql
kraiouchkine, line 1: I think the standard is unclear about the requirements for this rule, since as written, it applies to all "operands of binary or ternary operators." However, this query produces results in OpenPilot for the following cases that I think are either covered by Rule 5-2-1 or too noisy:
1) Member access operators, such as:
  - `ctversion[0] != rtversion[0]`
  - `assert(pvd->cnt < std::size(pvd->v))` (might be from macro expansion, but I have not checked)
  - `assert(s->font_sans_regular >= 0);`
2) Possibly relational operators in binary expressions (but not ternary expressions), such as:
  - `platform == 3 && encoding == 1`
  - `static_cast<uint8_t>(c) >= 0x20 && static_cast<uint8_t>(c) <= 0x7f`
  - `8 * sizeof(long) > 1 * PyLong_SHIFT`
  - But keep reporting cases such as `p[1] = (p[0] == tc[0] ? 0 : 255);` (currently reported)

Case 1 seems like something this rule should not report. The main justification for excluding case 2 is duplication of reporting already present in M5-2-1 and the noise it produces, but it seems technically correct to me.

kraiouchkine (changes requested):

cpp/autosar/src/rules/M5-0-2/InsufficientUseOfParentheses.ql
kraiouchkine: Additionally, I think macro expansion should be excluded in certain cases, such as:
- `assert(port >= 0)`
- With the macro `#define CMSG_LEN(len)   (CMSG_ALIGN (sizeof (struct cmsghdr)) + (len))`,  `CMSG_LEN(0)` in `recv_fds = (cmsg->cmsg_len - CMSG_LEN(0));` in `ipc.cc` in OpenPilot is currently reported.

kraiouchkine (approved):

PR #561, opened 2024-03-22
CTR55-CPP: address 374

### LINKED ISSUE #374: `CTR55-CPP`: Consider `iter != last` to be a bounds check
### Affected rules
 - `CTR55-CPP`

### Description

This query has a fairly restricted view of what a valid bounds check consists of. Consider expanding to support `iter != last` as a valid bound check, if `last` is derived from `end()`. We should also consider the case of utility methods on iterators, and whether they are correct at the call site.

### Example
```cpp
for (; iter != end; ++iter) {
}
```

knewbury01:
## Description

fix #374 

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - CTR55-CPP
     - CTR52-CPP (just a refactor)
## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [x] Confirmed

### Author

### Reviewer 

### DIFF
--- a/cpp/cert/src/rules/CTR52-CPP/GuaranteeGenericCppLibraryFunctionsDoNotOverflow.ql
+++ b/cpp/cert/src/rules/CTR52-CPP/GuaranteeGenericCppLibraryFunctionsDoNotOverflow.ql
@@ -80,7 +80,7 @@ where
     iteratorCreationCall = outputContainer.getAnIteratorFunctionCall() and
     iteratorCreationCall = c.getOutputIteratorSource()
   |
-    size_compare_bounds_checked(iteratorCreationCall, c)
+    sizeCompareBoundsChecked(iteratorCreationCall, c)
     or
     // Container created with sufficient size for the input
     exists(ContainerAccessWithoutRangeCheck::ContainerConstructorCall outputIteratorConstructor |
--- a/cpp/cert/src/rules/CTR52-CPP/GuaranteeGenericCppLibraryFunctionsDoNotOverflow.ql
+++ b/cpp/cert/src/rules/CTR52-CPP/GuaranteeGenericCppLibraryFunctionsDoNotOverflow.ql
@@ -80,16 +80,7 @@ where
     iteratorCreationCall = outputContainer.getAnIteratorFunctionCall() and
     iteratorCreationCall = c.getOutputIteratorSource()
   |
-    // Guarded by a bounds check that ensures our destination is larger than "some" value
-    exists(
-      GuardCondition guard, ContainerAccessWithoutRangeCheck::ContainerSizeCall sizeCall,
-      boolean branch
-    |
-      globalValueNumber(sizeCall.getQualifier()) =
-        globalValueNumber(iteratorCreationCall.getQualifier()) and
-      guard.controls(c.getBasicBlock(), branch) and
-      relOpWithSwapAndNegate(guard, sizeCall, _, Greater(), _, branch)
-    )
+    size_compare_bounds_checked(iteratorCreationCall, c)
     or
     // Container created with sufficient size for the input
     exists(ContainerAccessWithoutRangeCheck::ContainerConstructorCall outputIteratorConstructor |
--- a/cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
+++ b/cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
@@ -35,10 +35,9 @@ predicate sizeCheckedAbove(ContainerIteratorAccess it, IteratorSource source) {
  */
 predicate validEndBoundCheck(ContainerIteratorAccess it, IteratorSource source) {
   exists(
-    STLContainer c, BasicBlock b, GuardCondition l, ContainerIteratorAccess otherAccess,
-    IteratorSource end
+    ContainerAccessWithoutRangeCheck::ContainerEndCall end, BasicBlock b, GuardCondition l,
+    ContainerIteratorAccess otherAccess
   |
-    end = c.getAnIteratorEndFunctionCall() and
     //guard controls the access
     l.controls(b, _) and
     b.contains(it) and
--- a/cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
+++ b/cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
@@ -14,14 +14,14 @@
 import cpp
 import codingstandards.cpp.cert
 import codingstandards.cpp.Iterators
+import semmle.code.cpp.controlflow.Dominance
 
 /**
  * any `.size()` check above our access
  */
-predicate size_checked_above(ContainerIteratorAccess it, IteratorSource source) {
-  exists(STLContainer c, FunctionCall guardCall |
-    c.getACallToSize() = guardCall and
-    guardCall = it.getAPredecessor*() and
+predicate sizeCheckedAbove(ContainerIteratorAccess it, IteratorSource source) {
+  exists(ContainerAccessWithoutRangeCheck::ContainerSizeCall guardCall |
+    strictlyDominates(guardCall, it) and
     //make sure its the same container providing its size as giving the iterator
     globalValueNumber(guardCall.getQualifier()) = globalValueNumber(source.getQualifier()) and
     // and the size call we match must be after the assignment call
@@ -30,16 +30,22 @@ predicate size_checked_above(ContainerIteratorAccess it, IteratorSource source) 
 }
 
 /**
- * some loop check exists like: `iterator != end`
+ * some guard exists like: `iterator != end`
  * where a relevant`.end()` call flowed into end
  */
-predicate valid_end_bound_check(ContainerIteratorAccess it, IteratorSource source) {
-  exists(STLContainer c, Loop l, ContainerIteratorAccess otherAccess, IteratorSource end |
+predicate validEndBoundCheck(ContainerIteratorAccess it, IteratorSource source) {
+  exists(
+    STLContainer c, BasicBlock b, GuardCondition l, ContainerIteratorAccess otherAccess,
+    IteratorSource end
+  |
     end = c.getAnIteratorEndFunctionCall() and
-    //flow exists between end() and the loop condition
-    DataFlow::localFlow(DataFlow::exprNode(end), DataFlow::exprNode(l.getCondition().getAChild())) and
-    l.getCondition().getAChild() = otherAccess and
-    //make sure its the same iterator being checked as incremented
+    //guard controls the access
+    l.controls(b, _) and
+    b.contains(it) and
+    //guard is comprised of (anything flowing to) end check and an iterator access
+    DataFlow::localFlow(DataFlow::exprNode(end), DataFlow::exprNode(l.getChild(_))) and
+    l.getChild(_) = otherAccess and
+    //make sure its the same iterator being checked in the guard as accessed
     otherAccess.getOwningContainer() = it.getOwningContainer() and
     //make sure its the same container providing its end as giving the iterator
     globalValueNumber(end.getQualifier()) = globalValueNumber(source.getQualifier())
@@ -52,7 +58,7 @@ where
   it.isAdditiveOperation() and
   not exists(RangeBasedForStmt fs | fs.getUpdate().getAChild*() = it) and
   source = it.getANearbyAssigningIteratorCall() and
-  not size_compare_bounds_checked(source, it) and
-  not valid_end_bound_check(it, source) and
-  not size_checked_above(it, source)
+  not sizeCompareBoundsChecked(source, it) and
+  not validEndBoundCheck(it, source) and
+  not sizeCheckedAbove(it, source)
 select it, "Increment of iterator may overflow since its bounds are not checked."
--- a/cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
+++ b/cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
@@ -17,16 +17,28 @@ import codingstandards.cpp.Iterators
 import semmle.code.cpp.controlflow.Dominance
 
 /**
- * any `.size()` check above our access
+ * something like:
+ * `end = begin() + size()`
  */
-predicate sizeCheckedAbove(ContainerIteratorAccess it, IteratorSource source) {
-  exists(ContainerAccessWithoutRangeCheck::ContainerSizeCall guardCall |
-    strictlyDominates(guardCall, it) and
-    //make sure its the same container providing its size as giving the iterator
-    globalValueNumber(guardCall.getQualifier()) = globalValueNumber(source.getQualifier()) and
-    // and the size call we match must be after the assignment call
-    source.getASuccessor*() = guardCall
+Expr calculatedEndCheck(AdditiveOperatorFunctionCall calc) {
+  exists(
+    ContainerAccessWithoutRangeCheck::ContainerSizeCall size,
+    ContainerAccessWithoutRangeCheck::ContainerBeginCall begin
+  |
+    calc.getTarget().hasName("operator+") and
+    DataFlow::localFlow(DataFlow::exprNode(size), DataFlow::exprNode(calc.getAChild*())) and
+    DataFlow::localFlow(DataFlow::exprNode(begin), DataFlow::exprNode(calc.getAChild*())) and
+    //make sure its the same container providing its size as giving the begin
+    globalValueNumber(begin.getQualifier()) = globalValueNumber(size.getQualifier()) and
+    result = begin.getQualifier()
   )
+}
+
+Expr validEndCheck(FunctionCall end) {
+  end instanceof ContainerAccessWithoutRangeCheck::ContainerEndCall and
+  result = end.getQualifier()
+  or
+  result = calculatedEndCheck(end)
 }
 
 /**
@@ -35,19 +47,23 @@ predicate sizeCheckedAbove(ContainerIteratorAccess it, IteratorSource source) {
  */
 predicate validEndBoundCheck(ContainerIteratorAccess it, IteratorSource source) {
   exists(
-    ContainerAccessWithoutRangeCheck::ContainerEndCall end, BasicBlock b, GuardCondition l,
-    ContainerIteratorAccess otherAccess
+    FunctionCall end, BasicBlock b, GuardCondition l, ContainerIteratorAccess otherAccess,
+    Expr qualifierToCheck
   |
+    //sufficient end guard
+    qualifierToCheck = validEndCheck(end) and
     //guard controls the access
     l.controls(b, _) and
     b.contains(it) and
-    //guard is comprised of (anything flowing to) end check and an iterator access
+    //guard is comprised of end check and an iterator access
     DataFlow::localFlow(DataFlow::exprNode(end), DataFlow::exprNode(l.getChild(_))) and
     l.getChild(_) = otherAccess and
     //make sure its the same iterator being checked in the guard as accessed
     otherAccess.getOwningContainer() = it.getOwningContainer() and
-    //make sure its the same container providing its end as giving the iterator
-    globalValueNumber(end.getQualifier()) = globalValueNumber(source.getQualifier())
+    //if its the end call itself (or its parts), make sure its the same container providing its end as giving the iterator
+    globalValueNumber(qualifierToCheck) = globalValueNumber(source.getQualifier()) and
+    // and the guard call we match must be after the assignment call (to avoid valid guards protecting new iterator accesses further down)
+    source.getASuccessor*() = l
   )
 }
 
@@ -58,5 +74,5 @@ where
   not exists(RangeBasedForStmt fs | fs.getUpdate().getAChild*() = it) and
   source = it.getANearbyAssigningIteratorCall() and
   not validEndBoundCheck(it, source) and
-  not sizeCheckedAbove(it, source)
+  not sizeCompareBoundsChecked(source, it)
 select it, "Increment of iterator may overflow since its bounds are not checked."
--- a/cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
+++ b/cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
@@ -60,16 +60,19 @@ Expr getReferenceToOnePassedTheEndElement(Expr containerReference) {
  */
 predicate isUpperBoundEndCheckedIteratorAccess(IteratorSource source, ContainerIteratorAccess it) {
   exists(
-    Expr referenceToOnePassedTheEndElement, BasicBlock basicBlockOfIteratorAccess, GuardCondition upperBoundCheck,
-    ContainerIteratorAccess checkedIteratorAccess, Expr containerReferenceFromEndGuard
+    Expr referenceToOnePassedTheEndElement, BasicBlock basicBlockOfIteratorAccess,
+    GuardCondition upperBoundCheck, ContainerIteratorAccess checkedIteratorAccess,
+    Expr containerReferenceFromEndGuard
   |
     //sufficient end guard
-    referenceToOnePassedTheEndElement = getReferenceToOnePassedTheEndElement(containerReferenceFromEndGuard) and
+    referenceToOnePassedTheEndElement =
+      getReferenceToOnePassedTheEndElement(containerReferenceFromEndGuard) and
     //guard controls the access
     upperBoundCheck.controls(basicBlockOfIteratorAccess, _) and
     basicBlockOfIteratorAccess.contains(it) and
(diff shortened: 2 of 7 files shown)


cpp/cert/src/rules/CTR52-CPP/GuaranteeGenericCppLibraryFunctionsDoNotOverflow.ql
```diff
@@ -80,16 +80,7 @@ where
...
-    exists(
-      GuardCondition guard, ContainerAccessWithoutRangeCheck::ContainerSizeCall sizeCall,
-      boolean branch
-    |
-      globalValueNumber(sizeCall.getQualifier()) =
-        globalValueNumber(iteratorCreationCall.getQualifier()) and
-      guard.controls(c.getBasicBlock(), branch) and
-      relOpWithSwapAndNegate(guard, sizeCall, _, Greater(), _, branch)
-    )
+    size_compare_bounds_checked(iteratorCreationCall, c)
```
rvermeulen, line 83:
```suggestion
    isSizeCompareBoundsChecked(iteratorCreationCall, c)
```

We should adhere to camelcase for predicates.

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
```diff
@@ -15,23 +15,44 @@ import cpp
 import codingstandards.cpp.cert
 import codingstandards.cpp.Iterators
 
-from ContainerIteratorAccess it
+/**
+ * any `.size()` check above our access
+ */
+predicate size_checked_above(ContainerIteratorAccess it, IteratorSource source) {
```
rvermeulen, line 21:
```suggestion
predicate isSizeCheckedAbove(ContainerIteratorAccess it, IteratorSource source) {
```

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
```diff
@@ -15,23 +15,44 @@ import cpp
...
+    // and the size call we match must be after the assignment call
+    source.getASuccessor*() = guardCall
+  )
+}
+
+/**
+ * some loop check exists like: `iterator != end`
+ * where a relevant`.end()` call flowed into end
+ */
+predicate valid_end_bound_check(ContainerIteratorAccess it, IteratorSource source) {
```
rvermeulen, line 36:
```suggestion
predicate isValidEndBoundCheck(ContainerIteratorAccess it, IteratorSource source) {
```

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
```diff
@@ -15,23 +15,44 @@ import cpp
...
 import codingstandards.cpp.Iterators
 
-from ContainerIteratorAccess it
+/**
+ * any `.size()` check above our access
+ */
+predicate size_checked_above(ContainerIteratorAccess it, IteratorSource source) {
+  exists(STLContainer c, FunctionCall guardCall |
+    c.getACallToSize() = guardCall and
+    guardCall = it.getAPredecessor*() and
```
rvermeulen, line 24: We want to make sure that the guard call dominates the iterator access.

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
```diff
@@ -15,23 +15,44 @@ import cpp
 import codingstandards.cpp.cert
 import codingstandards.cpp.Iterators
 
-from ContainerIteratorAccess it
+/**
+ * any `.size()` check above our access
+ */
+predicate size_checked_above(ContainerIteratorAccess it, IteratorSource source) {
+  exists(STLContainer c, FunctionCall guardCall |
```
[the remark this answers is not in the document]
knewbury01: yes that looks to be the case, is the concern that by using the  STL container size calls it may miss some custom (not STL) container types?

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
rvermeulen: How does this compare to `size_compare_bounds_checked`, or what is the intended difference?

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
```diff
@@ -15,23 +15,44 @@ import cpp
...
+  )
+}
+
+/**
+ * some loop check exists like: `iterator != end`
+ * where a relevant`.end()` call flowed into end
+ */
+predicate valid_end_bound_check(ContainerIteratorAccess it, IteratorSource source) {
+  exists(STLContainer c, Loop l, ContainerIteratorAccess otherAccess, IteratorSource end |
+    end = c.getAnIteratorEndFunctionCall() and
```
rvermeulen, line 38: Can we generalize this with `getANonConstIteratorEndFunctionCall` and `getAConstIteratorEndFunctionCall`?

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
```diff
@@ -15,23 +15,44 @@ import cpp
...
+    //flow exists between end() and the loop condition
+    DataFlow::localFlow(DataFlow::exprNode(end), DataFlow::exprNode(l.getCondition().getAChild())) and
+    l.getCondition().getAChild() = otherAccess and
```
rvermeulen, line 41: Can we generalize this with a guard?

rvermeulen (changes requested):
I think there are some options to generalize the implementation a bit. Maybe because it is Friday 😅 , but I think we should add some explaining comments to the steps in the bound checks to guide the reader.

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
knewbury01: I dont understand the suggestion,
this predicate is already [returning the combination of those two you suggest](https://github.com/github/codeql-coding-standards/blob/main/cpp/common/src/codingstandards/cpp/Iterators.qll#L312)

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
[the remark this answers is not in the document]
knewbury01: have used the more general type here now

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
rvermeulen: True, but it is limited to `STLContainer` iterators. I assume we want to include user defined iterators as well?

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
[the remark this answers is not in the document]
rvermeulen: Yes, I think we should look at supporting user defined container follow as well.

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
knewbury01: valid, will extend `ContainerAccessWithoutRangeCheck` lib to include the more general end defn (similar to the more general size call)

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
```diff
@@ -14,24 +14,65 @@
...
+ * `end = begin() + size()`
+ */
+Expr calculatedEndCheck(AdditiveOperatorFunctionCall calc) {
+  exists(
+    ContainerAccessWithoutRangeCheck::ContainerSizeCall size,
+    ContainerAccessWithoutRangeCheck::ContainerBeginCall begin
+  |
+    calc.getTarget().hasName("operator+") and
+    DataFlow::localFlow(DataFlow::exprNode(size), DataFlow::exprNode(calc.getAChild*())) and
+    DataFlow::localFlow(DataFlow::exprNode(begin), DataFlow::exprNode(calc.getAChild*())) and
```
rvermeulen, line 30:
```suggestion
    DataFlow::localFlow(DataFlow::exprNode(size), DataFlow::exprNode(calc.getAChild+())) and
    DataFlow::localFlow(DataFlow::exprNode(begin), DataFlow::exprNode(calc.getAChild+())) and
```

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
```diff
@@ -14,24 +14,65 @@
...
+    //make sure its the same container providing its size as giving the begin
+    globalValueNumber(begin.getQualifier()) = globalValueNumber(size.getQualifier()) and
+    result = begin.getQualifier()
+  )
+}
+
+Expr validEndCheck(FunctionCall end) {
```
rvermeulen, line 37:
```suggestion
Expr getReferenceToOnePassedTheEndElement(Expr containerReference) {
```
I think we should clearly articulate what the predicates are doing to make the query more readable.
This predicate gets the expression that references to the element one passed the end element using either `end()` or `begin() + size()`.

Note that the `result` and parameter are swapped in my suggestion.

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
```diff
@@ -14,24 +14,65 @@
...
 import codingstandards.cpp.cert
 import codingstandards.cpp.Iterators
+import semmle.code.cpp.controlflow.Dominance
 
-from ContainerIteratorAccess it
+/**
+ * something like:
+ * `end = begin() + size()`
+ */
+Expr calculatedEndCheck(AdditiveOperatorFunctionCall calc) {
```
rvermeulen, line 23:
```suggestion
Expr getComputedReferenceToOnePassedTheEndElement(Expr containerReference) {
```

I think we should clearly articulate what the predicates are doing to make the query more readable.

Note that the `result` and the parameter are swapped in my suggestion and the result is more general than `AdditiveOperationFunctionCall`, but we should still cast in the body for now until we support more patterns.

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
```diff
@@ -14,24 +14,65 @@
...
+  result = end.getQualifier()
+  or
+  result = calculatedEndCheck(end)
+}
+
+/**
+ * some guard exists like: `iterator != end`
+ * where a relevant`.end()` call flowed into end
+ */
+predicate validEndBoundCheck(ContainerIteratorAccess it, IteratorSource source) {
```
rvermeulen, line 48:
```suggestion
predicate isBoundsCheckedIteratorAccess(ContainerIteratorAccess it, IteratorSource source) {
```

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
```diff
@@ -14,24 +14,65 @@
...
+    //guard is comprised of end check and an iterator access
+    DataFlow::localFlow(DataFlow::exprNode(end), DataFlow::exprNode(l.getChild(_))) and
+    l.getChild(_) = otherAccess and
+    //make sure its the same iterator being checked in the guard as accessed
+    otherAccess.getOwningContainer() = it.getOwningContainer() and
+    //if its the end call itself (or its parts), make sure its the same container providing its end as giving the iterator
+    globalValueNumber(qualifierToCheck) = globalValueNumber(source.getQualifier()) and
+    // and the guard call we match must be after the assignment call (to avoid valid guards protecting new iterator accesses further down)
+    source.getASuccessor*() = l
+  )
```
rvermeulen, line 67:
```suggestion
    Expr referenceToOnePassedTheEndElement, BasicBlock basicBlockOfIteratorAccess, GuardCondition upperBoundCheck, ContainerIteratorAccess checkedIteratorAccess,
    Expr containerReference
  |
    //sufficient end guard
    referenceToOnePassedTheEndElement = getReferenceToOnePassedTheEndElement(containerReference) and
    //guard controls the access
    upperBoundCheck.controls(basicBlockOfIteratorAccess, _) and
    basicBlockOfIteratorAccess.contains(it) and
    //guard is comprised of end check and an iterator access
    DataFlow::localFlow(DataFlow::exprNode(referenceToOnePassedTheEndElement), DataFlow::exprNode(upperBoundCheck.getChild(_))) and
    upperBoundCheck.getChild(_) = otherAccess and
    //make sure its the same iterator being checked in the guard as accessed
    checkedIteratorAccess.getOwningContainer() = it.getOwningContainer() and
    //if its the end call itself (or its parts), make sure its the same container providing its end as giving the iterator
    globalValueNumber(containerReference) = globalValueNumber(source.getQualifier()) and
    // and the guard call we match must be after the assignment call (to avoid valid guards protecting new iterator accesses further down)
    source.getASuccessor+() = upperBoundCheck
  )
```

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
```diff
@@ -14,24 +14,65 @@
...
+    //sufficient end guard
+    qualifierToCheck = validEndCheck(end) and
+    //guard controls the access
+    l.controls(b, _) and
+    b.contains(it) and
+    //guard is comprised of end check and an iterator access
+    DataFlow::localFlow(DataFlow::exprNode(end), DataFlow::exprNode(l.getChild(_))) and
+    l.getChild(_) = otherAccess and
+    //make sure its the same iterator being checked in the guard as accessed
+    otherAccess.getOwningContainer() = it.getOwningContainer() and
```
rvermeulen, line 62: This seems to not account for the case where two different iterators are created from the same container.

I don't think `globalValueNumber` will be useful because the iterator access might have a different value due to crements.

For now it is probably sufficient 🤔.

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
```diff
@@ -14,24 +14,65 @@
...
-    guardCall = it.getAPredecessor*() and
-    // and it uses the same qualifier as the one we were just assigned
-    nearbyAssigningIteratorCall.getQualifier().(VariableAccess).getTarget() =
-      guardCall.getQualifier().(VariableAccess).getTarget() and
-    // and the size call we match must be after the assignment call
-    nearbyAssigningIteratorCall.getASuccessor*() = guardCall
-  )
+  source = it.getANearbyAssigningIteratorCall() and
+  not validEndBoundCheck(it, source) and
+  not sizeCompareBoundsChecked(source, it)
```
rvermeulen, line 77: Small nitpick, can we make the order the same to make it easier to read?
```suggestion
  not validEndBoundCheck(it, source) and
  not sizeCompareBoundsChecked(it, source)
```

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
```diff
@@ -14,24 +14,65 @@
 import cpp
 import codingstandards.cpp.cert
 import codingstandards.cpp.Iterators
+import semmle.code.cpp.controlflow.Dominance
 
-from ContainerIteratorAccess it
+/**
+ * something like:
+ * `end = begin() + size()`
+ */
```
rvermeulen, line 22:
```suggestion
/**
 *  Get a derived one passed the end element for `containerReference`.
 *  The current supported pattern is:
 *     `end = begin() + size()`
 */
```

cpp/cert/test/rules/CTR55-CPP/test.cpp
```diff
@@ -20,10 +20,26 @@ void f1(std::vector<int> &v) {
...
+  {
+    auto end = v.end();
+    for (auto i = v.begin(); i != end; ++i) { // COMPLIANT
+    }
+  }
+
+  {
+    auto end2 = v.end();
+    end2++;                                    // NON_COMPLIANT
+    for (auto i = v.begin(); i != end2; ++i) { // NON_COMPLIANT[FALSE_NEGATIVE]
```
rvermeulen, line 42: Isn't the access `++i` not guarded by a upperbound check?
Or is there still local data flow `end2` to `end2` in `i !=end2` even though it is modified using `end2++`?

If that is the case, couldn't we invalidate `end2++` by looking at modifications?

rvermeulen (changes requested):
I add suggestion that imho improve the readability of the query.
Regarding the FN, can't we detect modifications if the local data flow isn't handling this?

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
```diff
@@ -14,24 +14,65 @@
...
+Expr calculatedEndCheck(AdditiveOperatorFunctionCall calc) {
+  exists(
+    ContainerAccessWithoutRangeCheck::ContainerSizeCall size,
+    ContainerAccessWithoutRangeCheck::ContainerBeginCall begin
+  |
+    calc.getTarget().hasName("operator+") and
```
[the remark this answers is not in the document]
knewbury01: yeah but that models both + and ++ , which we dont want the ++ operators here, those wouldnt make sense. including them most likely wont fp bc it would capture a more complicated ++ expr with a size and a begin flowing into it, but we still dont want those to count as our idea of sufficient end calculations

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
knewbury01: modified slightly to make it clear its end check specific, otherwise, great suggestion!

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
knewbury01: globalValueNumber here isnt checking the iterator accesses themselves, its checking the qualifier of the container function accesses (ie crements not relevant)

so for this:
```
   {
   auto end = v.end(); //line 1
    for (auto i = v.begin(); i != end; ++i) { // COMPLIANT //line 2
    }
  }
```
it will be checking is the v on line 1 is the same as the v on line 2

it cuts out cases like this, where the vector end check is not for the corresponding iterator access
```  
void test(std::vector<int> &v, std::vector<int> &v2) {
{
   auto end = v2.end(); //line 1
    for (auto i = v.begin(); i != end; ++i) { // NON_COMPLIANT
    }
  }
}

```

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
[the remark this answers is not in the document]
rvermeulen: That make sense, can you create a subclass that captures that to clean it up?

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
knewbury01: completely agree with the clearer naming required! thanks!

along those lines, actually this doesnt return the call that provides the sufficient end check, the predicate says - there is such a check, and returns the qualifier of one of the important components of that, so that we can validate that it was for the same vector object as the iterator access comes from

ie we cant just swap the parameter and the result without a bit more of a rewrite,

lmk if this new naming /comments still doesnt make it clear enough though

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
knewbury01: see [comment above](https://github.com/github/codeql-coding-standards/pull/561/files#r1543426947)

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
rvermeulen: My comment might have confusing, but I was not referring to the current use of `globalValueNumber`, but the potential use to determine if `otherAccess` and `it` are the same iterator. Our current check isn't sufficient in the case of multiple iterators.

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
rvermeulen: I only see a change in qldoc that is now inconsistent with the predicate.

The current implementation relates the qualifier of the calls to `begin` and `size` and their use in an `AdditivOperatorFunctionCall`. My suggestion just swaps that around where the qualifier is named `containerReference` and the operation representing the result of the end computation, the `AdditiveOperatorFunctionCall`, is now `result`. I think we can just swap those with a rewrite.

e.g.,

```ql
Expr getDerivedReferenceToOnePassedTheEndElement(Expr containerReference) {
  exists(
    ContainerAccessWithoutRangeCheck::ContainerSizeCall size,
    ContainerAccessWithoutRangeCheck::ContainerBeginCall begin,
    AdditiveOperatorFunctionCall calc
  |
    calc.getTarget().hasName("operator+")  and
    result = calc
  |
    DataFlow::localFlow(DataFlow::exprNode(size), DataFlow::exprNode(calc.getAChild*())) and
    DataFlow::localFlow(DataFlow::exprNode(begin), DataFlow::exprNode(calc.getAChild*())) and
    //make sure its the same container providing its size as giving the begin
    globalValueNumber(begin.getQualifier()) = globalValueNumber(size.getQualifier()) and
    containerReference = begin.getQualifier()
  )
}
```

knewbury01: accepted! minus the get an arg part - in this its not as simple as an arg - the size access is a child - 
```
l = (i + std::min(static_cast<std::vector<int>::size_type>(10), v.size()));
```

and I know any child is not the most precise modelling of this, but I still think this is overall a lot more precise than before

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
knewbury01: change accepted

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
knewbury01: hm, 
this case seems to me to be more corner case, or at least I am having trouble thinking of an example where this seems common/makes sense
feel free to suggest one if you have one in mind! otherwise, we can leave as a future improvement if we see cases where this misses out

cpp/cert/test/rules/CTR55-CPP/test.cpp
knewbury01: "Or is there still local data flow end2 to end2 in i !=end2 even though it is modified using end2++?" - yes, that is correct, it would need explicit invalidation modelling

I have made a note in the testcase about not handling this, think we can leave it as a future improvement

cpp/cert/src/rules/CTR55-CPP/DoNotUseAnAdditiveOperatorOnAnIterator.ql
rvermeulen: It is definitely a corner case, so I don't think we need to solve now.

rvermeulen (approved):
Great work, let's :shipit:!

PR #570, opened 2024-04-12
M9-3-3: address fp issue 381

### LINKED ISSUE #381: `M9-3-3`: Identify indirect assignment of member data
### Affected rules
 - `M9-3-3`

### Description

Improve the rule to detect assignment into references or pointers into member data.

### Example
In this example we see a member function which modifies a member variable called `values_`, by iterating through and taking a reference to each element.

```cpp
template <typename U = T>
void fill(const T& val) {
    for (auto& elem : values_) {
      elem = val;
    }
  }
```

knewbury01:
## Description

fixes #381 

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - M9-3-3

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


cpp/autosar/src/rules/M9-3-3/MemberFunctionConstIfPossible.ql
```diff
@@ -57,7 +64,9 @@ class ConstMemberFunctionCandidate extends NonConstMemberFunction {
     this.hasDefinition() and
     // For uninstantiated templates we have only partial information that prevents us from determining
     // if the candidate calls non-const functions. Therefore we exclude these.
-    not this.isFromUninstantiatedTemplate(_)
+    not this.isFromUninstantiatedTemplate(_) and
+    //cannot recommend const if it returns a nonconst reference
```
rvermeulen, line 68:
```suggestion
    // Cannot recommend const if it returns a non-const reference.
```

cpp/autosar/test/rules/M9-3-3/test.cpp
```diff
@@ -193,3 +193,24 @@ void test_template() {
...
+    for (auto &elem : values) {
+      elem = val;
+    }
+  }
+  constexpr int &front() noexcept { return values[0]; } // COMPLIANT
+};
+
+void fp_reported_in_381() {
+  Z4 z;
+  int i = z.front();
```
rvermeulen, line 214: Missing compliance annotation.

rvermeulen (changes requested):
Suggestion for comment and I think we are missing a compliance annotation in the added test case.

cpp/autosar/test/rules/M9-3-3/test.cpp
knewbury01: the rule alerts on functions, so their declarations not the usage (the compliance annotation is on [line 209 ](https://github.com/knewbury01/codeql-coding-standards/blob/knewbury01/fix-381/cpp/autosar/test/rules/M9-3-3/test.cpp#L209) but the usage snip is probably in there as I was checking if template instantiation was a factor of this fp/fn case...

good to check tho! thanks!

rvermeulen (approved):
LGTM.

PR #596, opened 2024-05-23
A18-5-8: address fp report 

### LINKED ISSUE #20: `A18-5-8`: Objects with an elided copy/move are falsely reported as not outliving a function
### Affected rules
 - `A18-5-8`

### Description

_Copy elision_ allows a call to a copy or move constructor to be omitted in certain cases where it is safe to do so (`[class.copy]/31`), such as copying/moving from a temporary object or where a copy/move occurs as part of a return from a function and the source is an local scope variable within that function.

Where copy elision applies, the CodeQL C++ extractor appears to remove the implicit copy/move calls. This is problematic because the query for this rule uses the presence of a copy/move constructor to determine whether an object outlives the lifetime of the function, thus causing false positives.

### Example
Consider the following example:
```cpp
std::unique_ptr<C1> Create(const std::string& s) noexcept {
  return std::make_unique<C1>(s);
}
```
Without copy elision, we would assume an implicit move/copy constructor call would exist here. However, due to copy elision the constructor call does not exist.

### Proposed solution

[`MakeSharedOrUnique.isAlwaysFreed()`](https://github.com/github/codeql-coding-standards/blob/main/cpp/autosar/src/rules/A18-5-8/UnnecessaryUseOfDynamicStorage.ql#L51) should be updated to consider the local data flow to the expression of a return statement as evidence that the heap memory associated with the shared or unique pointer is not freed within this function.

knewbury01:
## Description

fixes #20, add return expression flow for [NRVO elision](https://en.cppreference.com/w/cpp/language/copy_elision)

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - A18-5-8

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 

### DIFF
--- a/change_notes/2024-05-22-fix-fp-rule-A18-5-8.md
+++ b/change_notes/2024-05-22-fix-fp-rule-A18-5-8.md
@@ -1,0 +1,2 @@
+- `A18-5-8` - `UnnecessaryUseOfDynamicStorage.ql`:
+  - Address FP reported in #20. Add model of flow from MakeSharedOrUnique to return expression to capture copy/move elision case NRVO.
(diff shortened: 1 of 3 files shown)


nicolaswill (approved):
LGTM

PR #635, opened 2024-07-05
STR34-C: Do not consider integer type aliases in templates

### LINKED ISSUE #576: `STR34-C`: Do not consider integer type aliases in templates
### Affected rules
 - `STR34-C`

### Description

This query identifies conversions from signed `char`s to larger signed integers. This is a C rule, however it is part of the collection of C rules that can be applied to C++. In the case of C++, we observe potential false positives where such conversions happen in a template.

This is because the query usually only reports cases where `char` or `signed char` are directly referenced. This is to avoid flagging code using typedefs of `char` which are intended to be used integer types, not char types. For example, it's common for `int8_t` to be typedef'd to `char`, and the rule wouldn't apply in this case because there's no developer confusion over the conversion. However, in template instantiations we see the fully resolved types, which means we would flag conversions if they occur in the template.

### Example

```cpp
template <typename S, typename T> S get(T t) {
   S s = t;  // FALSE_POSITIVE - for instantiation
  return s;
}

void test(int8_t c) {
  int32_t a = c; // COMPLIANT - conversion occurs, but type is not char
  int32_t b = get<int32_t, int8_t>(c); // triggers a false positive in the template
}
```

mbaluda:
Fixes #576 

## Description

- Changed to a shared query so that it can be tested with c++
- Added c++ test case
- Excluded conversions in templates
- Included conversions to larger types with underlying integral type

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - STR34-C

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


rule_packages/c/Strings3.json
```diff
@@ -12,6 +12,7 @@
           "precision": "very-high",
```
knewbury01, line 12: when the rule is shared between c and c++, doesnt it need to have the shared implementation name portion in the package description file for both c/c++?

for example [like this](https://github.com/search?q=repo%3Agithub%2Fcodeql-coding-standards+IncludeGuardsNotUsed+language%3AJSON&type=code&l=JSON)?

lcartey (approved):

PR #637, opened 2024-07-10
M0-2-1: make into split and shared query

### LINKED ISSUE #119: `M0-2-1`: Consider array copies with overlapping data
### Affected rules
 - `M0-2-1`

### Description

The query as currently written only considers overlapping as caused by unions. We should also consider whether overlapping arrays are covered by the same rule.

### Example

```cpp
#include <cstring>
int16_t a[20];
void f2(void) {
  std::memcpy(&a[0], &a[1], 10u * sizeof(a[0]));  // Non-compliant
  std::memmove(&a[0], &a[1], 10u * sizeof(a[0])); // Compliant 
  std::memcpy(&a[1], &a[0], 10u * sizeof(a[0]));  // Non-compliant
  std::memmove(&a[1], &a[0], 10u * sizeof(a[0])); // Compliant 
}
```

knewbury01:
## Description

additionally moved some libraries to allow for use in shared query

Fixes #119 

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - M0-2-1
     - EXP43-C

## Release change checklist

  - [ ] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


rule_packages/cpp/Representation.json
```diff
@@ -53,6 +53,18 @@
           "tags": [
             "correctness"
           ]
+        },
+        {
+          "description": "Passing an aliased pointer to a conceptually restrict-qualified parameter is undefined behavior.",
```
nicolaswill, line 58: Could you please clarify what you mean by "conceptually"?

rule_packages/c/Pointers3.json
```diff
@@ -72,6 +72,7 @@
           "precision": "medium",
           "severity": "error",
           "short_name": "DoNotPassAliasedPointerToRestrictQualifiedParam",
+          "shared_implementation_short_name": "DoNotPassAliasedPointerToRestrictQualifiedParam_Shared",
```
nicolaswill, line 75: Our other shared queries typically use "Shared" as a suffix as opposed to "_Shared," if there is any suffix at all.

rule_packages/cpp/Representation.json
```diff
@@ -53,6 +53,18 @@
...
+          "kind": "problem",
+          "name": "Do not pass aliased pointers as parameters of functions where it is undefined behaviour for those pointers to overlap",
+          "precision": "medium",
+          "severity": "error",
+          "short_name": "DoNotPassAliasedPointerToRestrictQualifiedParam",
+          "shared_implementation_short_name": "DoNotPassAliasedPointerToRestrictQualifiedParam_Shared",
```
nicolaswill, line 64: See other suffix comment

nicolaswill (changes requested):
LGTM other than some naming and documentation points

rule_packages/cpp/Representation.json
knewbury01: was mostly referring to the fact that the `restrict` keyword comes from C (not C++) so technically this rule means something slightly different in C++ but checks the same concept

ie - in the C version of the [memcpy description](https://en.cppreference.com/w/c/string/byte/memcpy) - overlapping arrays are the problem (but using those violates the restrict contract)

in the C++ [memcpy description](https://en.cppreference.com/w/cpp/string/byte/memcpy) there is no restrict contract broken, but it is just that the arrays overlap

I think I can just reword this to make it clearer for the C++ version that its really only about overlapping arrays

nicolaswill (approved):
LGTM!

Issue #646, opened 2024-07-18
`M0-1-10`:  Functions called from int32_t main() are not considered as "called"


rak3-sh:
### Affected rules
 - M0-1-10

### Description
The `class MainFunction` defined in [EncapsulatingFunctions.qll](https://github.com/github/codeql-coding-standards/blob/main/cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll#L17) doesn't consider `int32_t main()` as a "main" function. Due to this, multiple false positives get raised in case the user has defined `main` with `int32_t` return type. 

### Example
Definitions of func1 and func2 are left out for reasons of brevity. 

```cpp
namespace mains {
    static int32_t var;

    // @brief namespace_func
    static void namespace_func(void) noexcept { // FP: Function is reported as not called, but it is called from "main" below.
        mains::var = -1;
        return;
    }
}  // namespace

// @brief main
// @return exit code
int32_t main(void) {
    int32_t ret {0};
    try {
        ret = func1(); // FP: Reported as not called in the definition location
        mains::var += ret;
        ret = func2(); // FP: Reported as not called in the definition location
        mains::var += ret;
    }
    catch(...) { 
        mains::namespace_func(); // namespace_func is called here.
    }
    return ret; 
}
```

rak3-sh:
A possible fix strategy would be to consider Int32_t type as well for `MainFunction` [here](https://github.com/github/codeql-coding-standards/blob/main/cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll#L17).  It works locally. Happy to provide a PR that can be reviewed and integrated. Kindly let me know.

```
class MainFunction extends MainLikeFunction {
  MainFunction() {
    hasGlobalName("main") and
    (
      getType() instanceof IntType or
      getType() instanceof Int32_t
    )
  }
}
```

lcartey:
Thanks for the report!

For a `main` function to be valid it needs to be using `int`, or an alias of `int`. So the above is only valid on platforms where `int32_t` is an alias for `int`. For this reason, I would suggest changing `MainFunction` to resolve typedefs, then perform the check against `IntType`, e.g.:
```codeql
class MainFunction extends MainLikeFunction {
  MainFunction() {
    hasGlobalName("main") and
    (
      getType().resolveTypeDefs() instanceof IntType
    )
  }
}
```
This would also have the benefit of capturing `main`s which use other aliases for `int`.

rak3-sh:
[...]

PR #647, opened 2024-07-23
Consider "main()" with typedef'd int return type as MainFunction

rak3-sh:
## Description

Currently, CodeQL considers `main` with an `int` return type as [MainFunction](https://github.com/github/codeql-coding-standards/blob/main/cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll#L14). However, `main` can also be defined to return other typdef'd variants of int like `int32_t`. The lack of this consideration causes CodeQL to report false positives whenever a query wants to use the `MainFunction` class. An example of this is the issue [#646](https://github.com/github/codeql-coding-standards/issues/646) for [M0-1-10](https://github.com/github/codeql-coding-standards/blob/main/cpp/autosar/src/rules/M0-1-10/UnusedFunction.ql). This PR modifies the definition of `MainFunction` to also consider typedef'd variants of `int`.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - M0-1-10

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


cpp/autosar/test/rules/M0-1-10.1/test.cpp
```diff
@@ -0,0 +1,34 @@
+#include <cstdint>
```
lcartey, line 1: Could you please add a non-compliant function into this test case? e.g. a function which is never called.

lcartey (changes requested):
One small change - other than that, looks good!

cpp/autosar/test/rules/M0-1-10.1/test.cpp
rak3-sh: Done in [0c20066](https://github.com/github/codeql-coding-standards/pull/647/commits/0c200662d4ccb8de83792676e9d2f3d0030d7cd2)

lcartey (approved):

PR #713, opened 2024-09-24
Implement Misra-c Noreturn rule package.

MichaelRFairhurst:
## Description

Implement Misra-c Noreturn rule package.

Covers rules 17-9 through 17-11, regarding the use of the _Noreturn attribute.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - RULE-17-9
     - RULE-17-10
     - RULE-17-11
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


c/misra/src/rules/RULE-17-11/FunctionWithNoReturningBranchShouldBeNoreturn.ql
```diff
@@ -0,0 +1,24 @@
...
+import codingstandards.c.misra
+import codingstandards.c.Noreturn
+
+from Function f
+where
+  not isExcluded(f, NoReturnPackage::returnStatementInNoreturnFunctionQuery()) and
+  not f instanceof NoreturnFunction and
+  not mayReturn(f) and
+  f.hasDefinition() and
+  f.getName() != "main" // Allowed exception; _Noreturn main() is undefined behavior.
```
lcartey, line 22: I would generally recommend the use of `not X = Y` instead of `X != Y`, because I tend to find the semantics are a better match for my intentions when I write this type of formula.

As a concrete example, imagine `File` had a predicate called `getALanguage()`, and we wanted to filter on files which were not one of `cpp` or `c`. With `!=` we might write:
```codeql
from File f
where f.getALanguage() != ["cpp", "c"]
select f
```
Somewhat surprisingly, if a file was compiled as both `c` and `cpp` (as can happen with, say, a header file), then this will still report it! This is because `!=` essentially says there is at least one pair of left and right for which the two are not equal. If we instead used `where not f.getALanguage() = ["cpp", "c"]` then we would get our expected behaviour - this is because `not .. = ..` essentially says that there doesn't exist a pair of left and right for which the two are equal.

In the specific case of this query, I don't think it matters - each function has one name, and is being compared to one named value - but getting into the habit of using `!=` means you may end up using it in a situation where it's not appropriate in the future.

More on this in the language reference:
https://codeql.github.com/docs/ql-language-reference/formulas/#footnote-1

```suggestion
  not f.getName() = "main" // Allowed exception; _Noreturn main() is undefined behavior.
```

c/misra/src/rules/RULE-17-11/FunctionWithNoReturningBranchShouldBeNoreturn.ql
```diff
@@ -0,0 +1,24 @@
...
+import cpp
+import codingstandards.c.misra
+import codingstandards.c.Noreturn
+
+from Function f
+where
+  not isExcluded(f, NoReturnPackage::returnStatementInNoreturnFunctionQuery()) and
+  not f instanceof NoreturnFunction and
+  not mayReturn(f) and
+  f.hasDefinition() and
```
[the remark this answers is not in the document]
lcartey: We should do similar for the other queries in this set.

c/misra/test/rules/RULE-17-11/test.c
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+
+void test_noreturn_f7(int i) { // COMPLIANT
+  if (i > 0) {
+    abort();
+  }
+}
+
+void test_noreturn_f8(int i) { // NON_COMPLIANT
```
lcartey, line 47: Can we also add a test case here for a `_Noreturn` function calling another `_Noreturn` function declared by us? I don't know if the control flow liveness reachability takes `_Noreturn` into account or not, so it would be good to validate.

c/misra/test/rules/RULE-17-11/test.c
```diff
@@ -0,0 +1,91 @@
...
+  } else {
+    abort();
+  }
+}
+
+_Noreturn void test_noreturn_f9(int i) { // COMPLIANT
+  if (i > 0) {
+    abort();
+  } else {
+    abort();
```
lcartey, line 59: We've tested `abort` and `exit(1)`. Are there other standard library functions we should test that abort early? `quick_exit` for example?

c/misra/test/rules/RULE-17-11/test.c
lcartey: Same for 17.9.

rule_packages/c/NoReturn.json
```diff
@@ -0,0 +1,55 @@
...
+    "RULE-17-11": {
+      "properties": {
+        "obligation": "advisory"
+      },
+      "queries": [
+        {
+          "description": "Functions which cannot return should be declared with _Noreturn.",
+          "kind": "problem",
+          "name": "A function without a branch that returns shall be declared with _Noreturn",
+          "precision": "high",
```
lcartey, line 29: Could this be `very-high`? When might be miss a live branch?

rule_packages/c/NoReturn.json
```diff
@@ -0,0 +1,55 @@
...
+      },
+      "queries": [
+        {
+          "description": "Function declared with _noreturn will by definition not return a value, and should be declared to return void.",
+          "kind": "problem",
+          "name": "A function declared with _noreturn shall have a return type of void",
+          "precision": "very-high",
+          "severity": "recommendation",
+          "short_name": "NonVoidReturnTypeOfNoreturnFunction",
+          "tags": []
```
lcartey, line 15: Correctness tag for each of these queries?

c/misra/src/rules/RULE-17-11/FunctionWithNoReturningBranchShouldBeNoreturn.ql
MichaelRFairhurst: Thanks for the detailed explanation! I thought I had handled this correctly...but I was only worrying about the case of `not exists(getName())`, and _not_ the case where a function has multiple names! Thanks! :brain:

rule_packages/c/NoReturn.json
MichaelRFairhurst: You're right, this should be very high. I was confusing precision with accuracy. It is possible to miss a _dead_ branch -- for instance, a third party library that didn't mark their noreturn functions as such...general halting problem goodness...

Is this worth marking 17-9 as "high" precision, or should 17-9 stay "very high"?

lcartey: I think both as `very-high` is good.

lcartey (approved):

PR #724, opened 2024-09-30
Michaelrfairhurst/implement types2 package

MichaelRFairhurst:
## Description

Implement Types2 package, with rules `RULE-7-5` and `RULE-7-6`

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - `RULE-7-5`
     - `RULE-7-6`
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


MichaelRFairhurst:
Seeing false positives that I don't understand here:

https://boringssl.googlesource.com/boringssl/+/master/third_party/fiat/curve25519_64.h#940

Can't reproduce them locally. Perhaps a gcc issue, will try linux.

MichaelRFairhurst:
The issue from the previous comment has been handled, and a test case added for it.

MichaelRFairhurst:
Talking to Vincent, preference seems to be to disallow nesting macros here, e.g., `INT8_C(INT8_MAX)`. Added that logic.

That discussion also prompted me to fully realizing that `INT64_C(-9223372036854775807 -1)` is risky and should not be allowed. The better option would be `INT64_C(-9223372036854775807) - 1`. This is because `INT64_C(c)` is defined as `c ## LL`, so the former pattern expands to `-9223372036854775807 -1LL` which is unexpected and could be dangerous. I removed this exception.

c/misra/src/rules/RULE-7-5/IncorrectlySizedIntegerConstantMacroArgument.ql
```diff
@@ -0,0 +1,46 @@
...
+import codingstandards.cpp.IntegerConstantMacro
+import codingstandards.cpp.Literals
+
+predicate matchesSign(IntegerConstantMacro macro, PossiblyNegativeLiteral literal) {
+  literal.isNegative() implies macro.isSigned()
+}
+
+predicate matchesSize(IntegerConstantMacro macro, PossiblyNegativeLiteral literal) {
+  // Wait for BigInt support to check 64 bit macro types.
+  (macro.getSize() < 64 and matchesSign(macro, literal))
```
lcartey, line 25: I think it's confusing that `matchesSize` holds if the macro size is >=64, or if the sign doesn't match - because it's not necessarily true that the size does match. I would therefore rewrite this predicate to:
```codeql
literal.getRawValue() <= macro.maxValue() and
literal.getRawValue() >= macro.minValue()
```
Then modify the `where` clause of the query to add in the extra conditions:
```codeql
    not matchesSign(macro, invoke.getExpr()) and
    explanation = " cannot be negative"
    or
    // Wait for BigInt support to check 64 bit macro types.
    macro.getSize() < 64 and
    // Sign must match, otherwise prefer to report the "cannot be negative" message
    matchesSign(macro, invoke.getExpr())
    not matchesSize(macro, invoke.getExpr()) and
    explanation = " is outside of the allowed range " + macro.getRangeString()
```

c/misra/src/rules/RULE-7-5/IncorrectlySizedIntegerConstantMacroArgument.ql
```diff
@@ -0,0 +1,46 @@
...
+
+from
+  PossiblyNegativeLiteral literal, MacroInvocation invoke, IntegerConstantMacro macro,
+  string explanation
+where
+  not isExcluded(invoke, Types2Package::incorrectlySizedIntegerConstantMacroArgumentQuery()) and
+  invoke.getMacro() = macro and
+  literal = invoke.getExpr() and
+  (
+    not matchesSign(macro, invoke.getExpr()) and explanation = " cannot be negative"
```
lcartey, line 41: Use `literal` not `invoke.getExpr()` in the rest of this query.

c/misra/src/rules/RULE-7-5/IntegerConstantMacroArgumentUsesSuffix.ql
```diff
@@ -0,0 +1,35 @@
...
+ *       external/misra/obligation/required
+ */
+
+import cpp
+import codingstandards.c.misra
+import codingstandards.cpp.IntegerConstantMacro
+import codingstandards.cpp.Literals
+
+string argumentSuffix(MacroInvocation invoke) {
+  // Compiler strips the suffix unless we look at the unexpanded argument text.
```
lcartey, line 21:
```suggestion
  // Extractor strips the suffix unless we look at the unexpanded argument text.
```

c/misra/src/rules/RULE-7-5/IntegerConstantMacroArgumentUsesSuffix.ql
```diff
@@ -0,0 +1,35 @@
...
+  result = invoke.getUnexpandedArgument(0).regexpCapture("([0-9]+|0[xX][0-9A-F]+)([uUlL]+)$", 2)
+}
+
+from MacroInvocation invoke, PossiblyNegativeLiteral argument, string suffix
+where
+  not isExcluded(invoke, Types2Package::integerConstantMacroArgumentUsesSuffixQuery()) and
+  invoke.getMacro() instanceof IntegerConstantMacro and
+  invoke.getExpr() = argument and
+  suffix = argumentSuffix(invoke)
+select invoke.getExpr(),
```
lcartey, line 33:
```suggestion
select argument,
```

c/misra/src/rules/RULE-7-5/InvalidIntegerConstantMacroArgument.ql
```diff
@@ -0,0 +1,35 @@
...
+}
+
+from MacroInvocation invoke, IntegerConstantMacro macro
+where
+  not isExcluded(invoke, Types2Package::invalidIntegerConstantMacroArgumentQuery()) and
+  invoke.getMacro() = macro and
+  (
+    not invoke.getExpr() instanceof PossiblyNegativeLiteral
+    or
+    containsMacroInvocation(invoke, _)
```
lcartey, line 32: There is already a predicate `getParentInvocation` you can use for this:
```suggestion
    any(MacroInvocation inner).getParentInvocation() = invoke
```

cpp/common/src/codingstandards/cpp/IntegerConstantMacro.qll
```diff
@@ -0,0 +1,37 @@
...
+  boolean signed;
+  int size;
+
+  IntegerConstantMacro() {
+    signed = true and size = getName().regexpCapture("INT(8|16|32|64)_C", 1).toInt()
+    or
+    signed = false and size = getName().regexpCapture("UINT(8|16|32|64)_C", 1).toInt()
+  }
+
+  predicate isSmall() { size < 32 }
```
lcartey, line 18: Use the size of the int type in this database - it's not guaranteed to be 32.
```suggestion
  predicate isSmall() { size <   any(IntType it | it.isSigned()).getSize() }
```

lcartey (changes requested):
A couple of stylistic suggestions, and noting an edge case on int type. Otherwise looks good!

c/misra/src/rules/RULE-7-5/InvalidIntegerConstantMacroArgument.ql
MichaelRFairhurst: Thanks!

cpp/common/src/codingstandards/cpp/IntegerConstantMacro.qll
MichaelRFairhurst: Done!

cpp/common/src/codingstandards/cpp/Literals.qll
```diff
@@ -70,3 +71,46 @@ class BoolLiteral extends Literal {
...
+ * "negative literal."
+ */
+class NegativeLiteral extends PossiblyNegativeLiteral, UnaryMinusExpr {
+  Cpp14Literal::NumericLiteral literal;
+
+  NegativeLiteral() { literal = getOperand() }
+
+  override Cpp14Literal::NumericLiteral getBaseLiteral() { result = literal }
+
+  override float getRawValue() { result = -lowerBound(literal) }
```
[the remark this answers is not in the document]
MichaelRFairhurst: That passes tests, yes, thanks!

Issue #711, opened 2024-09-24
`M0-1-10`:  Generates lot of false positives for used functions


rak3-sh:
### Affected rules
 - M0-1-10

### Description

The query M0-1-10 generates a significant amount of false positives while analyzing "unused functions". These are noticed especially in the following scenarios.
- constexpr functions
- template instantiations
- special member functions (constructors, destructors, operators, conversion operators)

This results from the fact that CodeQL cannot track the usages of such functions in compile time constructs like constexpr, static_asserts, templates etc.

### Example
`test.hpp`
```cpp
template <class T>
constexpr T to_underlying(T value) noexcept { // M0-1-3 violation reported here
  return static_cast<T>(value);
}
```

`test.cpp`
```cpp
#include <type_traits>
#include "test.hpp"

template <typename T1, typename T2>
constexpr bool StaticAssertTypeEq() noexcept {
  static_assert(std::is_same<T1, T2>::value, "T1 and T2 are not the same type");
  return true;
}

template <typename T, int val>
class X
{
  T arr[val];
};

void foo()
{
  struct dummy {
    dummy() noexcept(false) { static_cast<void>(0); } // M0-1-3 violation reported here
  };
  // usage of dummy
  static_assert(!std::is_nothrow_default_constructible<X<dummy, 5>>::value,
                "Must not be nothrow default constructible");
}

int main()
{
  int a;
  StaticAssertTypeEq<decltype(to_underlying(a)), int>(); // usage of to_underlying
  foo();
  return 0;
}
```

rak3-sh:
### Proposed Fix(es)
Hi @lcartey, kindly let me know your valuable insights whether the following can be acceptable or not.

The MISRA2008 standard doesn't highlight any specific exception for this rule. However, considering CodeQL's limitation and to make it more practical to check the results of this query, can we propose to restrict this query in the following way? I would like to seek opinions about the same.
- Avoid analyzing functions that are constexpr because CodeQL has known limitations about tracking the usages of constexpr functions/expressions.
- Avoid analyzing explicitly defaulted functions. Since the user has marked it as "default", should we consider these as candidates for unused functions?
- Avoid analyzing functions which result from template instantiations. Since the function has been instantiated, there should be an explicit call.
- Avoid analyzing special member functions (constructors, destructors, operators, conversion operators). This is proposed mainly because their usages in compile time constructs are not tracked properly by CodeQL.
[...]

PR #725, opened 2024-10-02
Fix false positives of M0-1-10 (#711)

rak3-sh:
## Description

M0-1-10 produces false positives described in #711. This PR fixes the same by implementing the following:
1. Doesn't consider constexpr functions for analysis. This is because CodeQL has known limitations about tracking the usages of the same.
2. Considers a function from GoogleTest framework as an EntryPoint. Previously functions that were invoked from such test functions were considered as unused because these were not considered as EntryPoints.
3. Splits the query into two queries - one that reports unused functions that are not Special Member functions and the other that reports only unused Special Member functions. This is done so as to enable deviations to be applied to this case.  

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - M0-1-10

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


rak3-sh:
Hi @lcartey, as discussed on #711 , I've implemented the changes on this PR. Requesting for your valuable inputs and suggestions.

cpp/autosar/src/rules/M0-1-10/UnusedSplMemberFunction.ql
```diff
@@ -0,0 +1,32 @@
+/**
+ * @id cpp/autosar/unused-spl-member-function
```
lcartey, line 2: Adding a new query requires updating the rule packages metadata file, then running a regeneration script in order to setup the new query.

The full details are described in the development handbook:
https://github.com/github/codeql-coding-standards/blob/main/docs/development_handbook.md#step-2-review-and-update-the-rule-package-description-file

But the quick guide would be:
 1. Modify the `cpp/DeadCode.json` file to add a new element to the array for the `queries` property for `M0-1-10`.
 2. Set the metadata appropriately on that query.
 3. Run `python3.9 scripts/generate_rules/generate_package_files.py cpp DeadCode` to regenerate the relevant files.
 4. You will need to update your `isExcluded` predicate to refer to your new query - likely `DeadCodePackage::unusedSplMemberFunctionQuery()` if you keep with the same name.

Doing this will ensure that this query is properly integrated into the deviations mechanism.

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
```diff
@@ -18,6 +18,39 @@ class MainFunction extends MainLikeFunction {
...
 
+/**
+ * A test function from the GoogleTest infrastructure.
+ *
+ * Such functions can be treated as valid EntryPoint functions during analysis
+ * of "called" or "unused" functions. It is not straightforward to identify
+ * such functions, however, they have certain features that can be used for
+ * identification. This can be refined based on experiments/real-world use.
+ */
+class GTestFunction extends MainLikeFunction {
```
lcartey, line 29: I would have a small preference for calling this `GoogleTestFunction` - it's more verbose, but easier to understand what it covers if you see it in code.

cpp/common/src/codingstandards/cpp/deadcode/UnusedFunctions.qll
```diff
@@ -75,7 +75,9 @@ module UnusedFunctions {
    */
 
   private class MainLikeFunctionEntryPoint extends EntryPoint, MainLikeFunction {
-    MainLikeFunctionEntryPoint() { this instanceof MainLikeFunction }
+    MainLikeFunctionEntryPoint() {
+      this instanceof MainLikeFunction or this instanceof GTestFunction
```
lcartey, line 79: `MainLikeFunction` is an abstract class, which means it represents the set of all values from all subclasses already, so you don't need to specifically mention GTestFunction here.
```suggestion
      this instanceof MainLikeFunction
```

cpp/common/src/codingstandards/cpp/deadcode/UnusedFunctions.qll
```diff
@@ -111,6 +113,26 @@ module UnusedFunctions {
     }
   }
 
+  /**
+   * A `MemberFunction` which is either a Default constructor, Destructor
+   * CopyConstructor, CopyAssingmentOperator, MoveConstructor or a
+   * MoveAssignmentOperator
+   */
+  predicate isASpecialMemberFunction(MemberFunction f) {
```
lcartey, line 121: There is already a CodeQL class for representing SpecialMemberFunctions:

https://github.com/github/codeql-coding-standards/blob/main/cpp/common/src/codingstandards/cpp/Class.qll#L87-L103

I would recommend using that instead.

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
```diff
@@ -18,6 +18,39 @@ class MainFunction extends MainLikeFunction {
...
+  GTestFunction() {
+    // A GoogleTest function is named "TestBody" and
+    this.hasName("TestBody") and
+    // is enclosed by a class that inherits from a base class
+    this.getEnclosingAccessHolder() instanceof Class and
```
lcartey, line 34: I think this condition is redundant, because the condition on line 36 will enforce the same constraint.

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
```diff
@@ -18,6 +18,39 @@ class MainFunction extends MainLikeFunction {
...
+    exists(Class base |
+      base = this.getEnclosingAccessHolder().(Class).getABaseClass() and
+      (
+        // called "Test" or
+        exists(Class c | base.getABaseClass() = c and c.hasName("Test"))
+        or
+        // defined under a namespace called "testing" or
+        exists(Namespace n | n = base.getNamespace() | n.hasName("testing"))
```
lcartey, line 42: I think we want to `and` these two conditions - i.e. we are looking for a base class which is called `Test` and is in the `testing` namespace. Just looking for base classes called `Test` could result in false negatives for code which uses other frameworks.

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
```diff
@@ -18,6 +18,39 @@ class MainFunction extends MainLikeFunction {
...
+ * identification. This can be refined based on experiments/real-world use.
+ */
+class GTestFunction extends MainLikeFunction {
+  GTestFunction() {
+    // A GoogleTest function is named "TestBody" and
+    this.hasName("TestBody") and
+    // is enclosed by a class that inherits from a base class
+    this.getEnclosingAccessHolder() instanceof Class and
+    exists(Class base |
+      base = this.getEnclosingAccessHolder().(Class).getABaseClass() and
```
lcartey, line 36: I'm not deeply familiar with Google Test, but if you can have multiple levels of sub-classing, we should use `getABaseClass+()` to find all transitive base classes.
```suggestion
      base = this.getEnclosingAccessHolder().(Class).getABaseClass+() and
```

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
```diff
@@ -18,6 +18,39 @@ class MainFunction extends MainLikeFunction {
...
+        exists(Class c | base.getABaseClass() = c and c.hasName("Test"))
+        or
+        // defined under a namespace called "testing" or
+        exists(Namespace n | n = base.getNamespace() | n.hasName("testing"))
+        or
+        // is templatized by a parameter called "gtest_TypeParam_"
+        exists(TemplateParameter tp |
+          tp = base.getATemplateArgument() and
+          tp.hasName("gtest_TypeParam_")
+        )
```
lcartey, line 48: Can you provide an example of what this is capturing?

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
```diff
@@ -18,6 +18,39 @@ class MainFunction extends MainLikeFunction {
...
+      )
+    )
+  }
+}
```
lcartey, line 52: We would also need to add a test case for Google Test entry points. This would require adding a stub for Google Test to `cpp/common/test/includes/googletest/gtest.h`, update the [options](https://github.com/github/codeql-coding-standards/blob/main/cpp/common/test/options) file to add the new google test directory to the include path, then referencing the relevant stubs in your test.

lcartey (changes requested):
Thanks @rak3-sh! I've made some comments, mainly around tying the new query into the deviations mechanism and how we represent/test the Google Test entry points change.

cpp/autosar/src/rules/M0-1-10/UnusedSplMemberFunction.ql
rak3-sh: Addressed in: [7bf12c4](https://github.com/github/codeql-coding-standards/pull/725/commits/7bf12c4902c6acd803b90856b8af6b3c2fff317c)

cpp/common/src/codingstandards/cpp/deadcode/UnusedFunctions.qll
rak3-sh: thanks for noticing! I had missed to modify it after I made GTestFunction extend from MainLikeFunction.

cpp/common/src/codingstandards/cpp/deadcode/UnusedFunctions.qll
rak3-sh: Sounds good!

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
rak3-sh: I want to check if a function called `TestBody` is in a class that inherits a base class templatized by `gtest_TypeParam_`. If we look into the definitions of the [TYPED_TEST](https://github.com/google/googletest/blob/main/googletest/include/gtest/gtest-typed-test.h#L202) macro of GoogleTest, they use this specific name for the template parameter. We can also check it for the class that the `TestBody` member function belongs to, but since we already have a handle to the Base class, I used that. Let me know in case it can be improved.

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
rak3-sh: Yes, I think + would be better here! Thanks for the insights.

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
rak3-sh: I am thinking of making this simpler. Kindly wait.

rak3-sh: Requesting to review commit: [a48c7cc](https://github.com/github/codeql-coding-standards/pull/725/commits/a48c7cc6af11ed445cee5a43d1e823db6b2c3833)

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
rak3-sh: Requesting to review commit: [a48c7cc](https://github.com/github/codeql-coding-standards/pull/725/commits/a48c7cc6af11ed445cee5a43d1e823db6b2c3833)

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
rak3-sh: Requesting to review commit: [a48c7cc](https://github.com/github/codeql-coding-standards/pull/725/commits/a48c7cc6af11ed445cee5a43d1e823db6b2c3833)

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
rak3-sh: Requesting to review commit: [a48c7cc](https://github.com/github/codeql-coding-standards/pull/725/commits/a48c7cc6af11ed445cee5a43d1e823db6b2c3833)

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
rak3-sh: I understand. I will work on this.

cpp/common/src/codingstandards/cpp/deadcode/UnusedFunctions.qll
rak3-sh: Fixed in [bd83048](https://github.com/github/codeql-coding-standards/pull/725/commits/bd83048f7d9d9e76a9c7e8a5d6ee0d252947ff83)

cpp/autosar/src/rules/M0-1-10/UnusedSplMemberFunction.ql
```diff
@@ -0,0 +1,31 @@
...
+ *       external/autosar/obligation/advisory
+ */
+
+import cpp
+import codingstandards.cpp.autosar
+import codingstandards.cpp.deadcode.UnusedFunctions
+
+from UnusedFunctions::UnusedSplMemberFunction unusedSplMemFunction, string name
+where
+  not isExcluded(unusedSplMemberFunctionQuery, DeadCodePackage::unusedFunctionQuery()) and
```
lcartey, line 23: I think this line still needs changing, to `DeadCodePackage::unusedSplMemberFunctionQuery()`.

rak3-sh: Addressed in https://github.com/github/codeql-coding-standards/pull/725/commits/97bf53af4294acb6afe3a0a794f4840e75c93bb1

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
rak3-sh: Addressed in [c5cdf0e](https://github.com/github/codeql-coding-standards/pull/725/commits/c5cdf0e0b7e40239966e5bfb0127264f0b1e5627)

rak3-sh:
Thanks a lot @lcartey for your insightful comments! I've addressed all open issues and I would like to request for a re-review.

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
```diff
@@ -18,6 +18,34 @@ class MainFunction extends MainLikeFunction {
...
+ *
+ * Such functions can be treated as valid EntryPoint functions during analysis
+ * of "called" or "unused" functions. It is not straightforward to identify
+ * such functions, however, they have certain features that can be used for
+ * identification. This can be refined based on experiments/real-world use.
+ */
+class GoogleTestFunction extends MainLikeFunction {
+  GoogleTestFunction() {
+    // A GoogleTest function is named "TestBody" and
+    this.hasName("TestBody") and
```
lcartey, line 32: I think you could also mark the destructor as an "entry point", as it will also presumably be called by the gtest framework. That would eliminate the false positive you see in your test case.

lcartey (changes requested):
One final change, to address a further FP for gtest, otherwise looks good!

cpp/common/src/codingstandards/cpp/EncapsulatingFunctions.qll
rak3-sh: thanks for your suggestion .. let me try that out!

rak3-sh: It works! I've added SpecialMemberFunction instead of just the Destructor as entry points from GoogleTestFunction. This is addressed in [4b1429b](https://github.com/github/codeql-coding-standards/pull/725/commits/4b1429b6bf94bce541c32395eab0529a6d5ce9cb)

lcartey (approved):

PR #730, opened 2024-10-03
Implement Language4 package, banning obsolete language features.

MichaelRFairhurst:
## Description

Implement Language4 package, banning obsolete language features.

Many of the cases outlined in the amendment are covered by other rules. Add support for new cases where possible (was not possible for ID 3, storage class specifiers not at beginning of declaration, or ID 2, which is a feature of the implementation not determinable by static analysis), and reference existing rules in one comprehensive test for maximal clarity that those parts of rule 1-5 are indeed supported by our existing queries.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - `RULE-1-5`
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


c/misra/src/rules/RULE-1-5/CallToReallocWithSizeZero.ql
```diff
@@ -0,0 +1,25 @@
...
+import cpp
+import codingstandards.c.misra
+import semmle.code.cpp.rangeanalysis.new.RangeAnalysis
+
+from FunctionCall call, Expr arg
+where
+  not isExcluded(call, Language4Package::callToReallocWithSizeZeroQuery()) and
+  call.getTarget().hasGlobalOrStdName("realloc") and
+  arg = call.getArgument(1) and
+  upperBound(arg) = 0
```
lcartey, line 24: I can think of two different ways to write this rule:
 * Look for the use of `0` as a literal, that is then used in the realloc call.
 * Use range analysis, and report if `0` is a possible value that can be assigned.

The first is more precise - we're unlikely to get false positives - but may miss cases where a computed value is used in a `realloc` call which could be zero in some circumstances.

The second approach addresses that problem, but can cause false positives because our range is a single contiguous range and may not be able to exclude cases where 0 isn't possible.

In general, I think the range analysis approach is reasonable, although we want to confirm that it doesn't create significant false positives in common scenarios.

Your current implementation is a restricted version of the range analysis approach, because you force the upper bound to be zero. Using this approach, we would not find, e.g., this case:
```cpp
int n = some_bool ? 0 : 5;
realloc(t,n)
```
This is because the range of `n` is `0..5`, so the `upperBound` is `5`, even though `0` is set specifically.

I also don't think `upperBound` takes into account conversions. Given realloc takes `size_t` (i.e. an unsigned value), this type of scenario - which is morally equivalent to the above - would be reported:
```
int n = some_bool ? 0 : -5;
realloc(t,n)
```

I think my recommendations would be to try using `upperBound(arg.getConversion()) >= 0 and lowerBound(arg.getConversion()) <=0`, and review using MRVA to see how many FPs we get, and whether we can address them.

lcartey (changes requested):
Thanks! I've provided some feedback on Rule 1.5. There's also the open question of whether we should include (via shared rules) the existing rules into Rule 1.5 directly itself.

c/misra/src/rules/RULE-1-5/CallToReallocWithSizeZero.ql
MichaelRFairhurst: Running MRVA, this may have something like a 20% ish false positive rate (90 results out of 512 total calls to `realloc()`).

I'm not sure how to address these false positives. It looks like most of them are allocating a fixed number of bytes extra per some event, and either that event parameter takes an `int` size instead of a `uint size`, and/or, range analysis might be predicting possible overflow.

Other false positives are the result of doing more complex computations (perhaps, multiplication, subtracting from space remaining, etc).

Given that `realloc()` is already generally banned...I'm not sure if it is common that customers run these misra suites on non-embedded code where `malloc()` is allowed & used....I'm not sure if we should err on the side of false positives here, or false negatives here, etc.

And of course, for codebases worried about compliance, which for some reason have decided to allow `malloc`/`free`/`realloc`, they should be able to insert checks that `size > 0` to better inform range analysis and suppress false positives.

lcartey (approved):
Looks good, thanks!

PR #732, opened 2024-10-04
A0-1-1: Fix useless assignment false positive on constexpr array size

fjatWbyT:
## Description

Addresses false positive like the one in #690 in another AUTOSAR ule. 

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - A0-1-1

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


fjatWbyT:
Hello @lcartey,

I looked for unit tests analogous to the one updated in #690, in this case for A0-1-1 "useless assignment", in contrast with M0-1-9 "no dead code", but couldn't find any already existing. I run locally the (153) tests in `cpp/common/test/rules` after applying the change and all pass. I could find tests for useless assignment in deviations, but I doubt those need updating for this change since there's no new deviation.

Does it make sense to add explicit coverage for A0-1-1?
I wonder if it would be sensible to (a) add a new directory `cpp/common/uselessassignment`, (b) update configuration so that the ones in `cpp/common/deadcode` are also run against A0-1-1, or (c) something else :)
On the one hand, I think that covering these other `isUselessSsaDefinition` predicate and `InterestingStackVariable` class (assuming they aren't already) would be an improvement. On the other hand, since they are (at least partially) used by the dead code qll, it might not be worth further updates.

lcartey:
Hi @fjatWbyT! Thanks for the PR.

The test cases for `A0-1-1` live here:
https://github.com/github/codeql-coding-standards/tree/main/cpp/autosar/test/rules/A0-1-1

The tests live in a common directory when one query implementation is shared by multiple rules - for example, where an AUTOSAR and CERT C++ rule cover the same issue. `A0-1-1` is only required to be reported to AUTOSAR, so the tests live in the AUTOSAR test directory, under the rule ID.

fjatWbyT:
Aaah, it is clear now. Thank you very much for the explanation. I will update the PR with extra coverage for this case in that file.

lcartey (approved):
Thank you!

Issue #739, opened 2024-10-07
`A14-5-2`: type member dependent on class' template parameter detected not dependent


fjatWbyT:
### Affected rules
 - `A14-5-2`

### Description

Type member defined with a dependent name and template parameter results in recommendation to be defined in a separated class because it is identified to be not dependent on the template parameter.

### Example

```cpp
template<typename T>
struct a_template_struct {
  using type = T;
};

template<typename T>
struct another_template_struct {
  using type = typename a_template_struct<T>::type;
};
```

Query `cpp/autosar/src/rules/A14-5-2/NonTemplateMemberDefinedInTemplate.ql` outputs that `another_template_struct::type` does not use any of the template arguments of `another_template_struct`.

fjatWbyT:
### Strategy proposal

Add an additional [omission case](https://github.com/github/codeql-coding-standards/blob/main/cpp/autosar/src/rules/A14-5-2/NonTemplateMemberDefinedInTemplate.ql#L161) for type members declared with [using aliases](https://codeql.github.com/codeql-standard-libraries/cpp/semmle/code/cpp/TypedefType.qll/type.TypedefType$UsingAliasTypedefType.html). This is motivated by the many `A14-5-2` alerts overall produced in STL-like libraries (out of which, I've estimated an 80% of them to be false positives).

For example, I measured 1988 `A14-5-2` alerts in [abseil-cpp](https://github.com/abseil/abseil-cpp) including false positives in aliases such as the ones starting [at this line](https://github.com/abseil/abseil-cpp/blob/40cab244558f225cadcb943f84c5209f7ce876b1/absl/container/internal/btree.h#L1103), and reduced it to 1540 with this strategy.

I was happy to see that [current unit tests](https://github.com/github/codeql-coding-standards/blob/main/cpp/autosar/test/rules/A14-5-2/test.cpp#L10) would finely do to capture the impact:

```cpp
 using T1 = typename template_base<T>::type;   // COMPLIANT[FALSE_POSITIVE]
 using T2 = typename template_base<int>::type; // NON_COMPLIANT
 ```
 
 would become
 
 ```cpp
 using T1 = typename template_base<T>::type;   // COMPLIANT
 using T2 = typename template_base<int>::type; // NON_COMPLIANT[FALSE_NEGATIVE]
 ```

fjatWbyT:
Hi, I think this issue could be closed ✅


I have just verified on abseil-cpp that results using current main reconcile:
[...]

PR #746, opened 2024-10-15
`A14-5-2` do not consider type members declared with using aliases.

fjatWbyT:
## Description

Implement [proposal](https://github.com/github/codeql-coding-standards/issues/739#issuecomment-2409588476) to omit type members declared with using aliases in rule `A14-5-2`.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

     - `A14-5-2`

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


cpp/autosar/src/rules/A14-5-2/NonTemplateMemberDefinedInTemplate.ql
```diff
@@ -167,7 +167,9 @@ where
         mf = c.getAMemberFunction() and not mf.isCompilerGenerated() and not exists(mf.getBlock())
       )
     )
-  )
+  ) and
+  // Omit using alias (cf. https://github.com/github/codeql-coding-standards/issues/739)
+  not d instanceof UsingAliasTypedefType
```
lcartey, line 172: I think we can actually do better here. In the case we want to exclude, the entire "base type" of the using alias is considered to be template parameter. So we can exclude it directly:
```suggestion
  // Exclude Using alias which refer directly to a TypeParameter
  not d.(UsingAliasTypedefType).getBaseType() instanceof TemplateParameter
```
That removes the false positive and retains the true positive.

cpp/autosar/test/rules/A14-5-2/test.cpp
```diff
@@ -7,8 +7,8 @@ template <typename T> class C1 {
 public:
   enum E1 : T { e1, e2 }; // COMPLIANT
 
-  using T1 = typename template_base<T>::type;   // COMPLIANT[FALSE_POSITIVE]
-  using T2 = typename template_base<int>::type; // NON_COMPLIANT
+  using T1 = typename template_base<T>::type;   // COMPLIANT
+  using T2 = typename template_base<int>::type; // NON_COMPLIANT[FALSE_NEGATIVE]
```
lcartey, line 11: If you adopt my suggestion above, this will be correctly found.
```suggestion
  using T2 = typename template_base<int>::type; // NON_COMPLIANT
```

change_notes/2024-10-15-fix-fp-739-a14-5-2.md
```diff
@@ -0,0 +1,2 @@
+- `A14-5-2` - `NonTemplateMemberDefinedInTemplate.ql`
+  - Fixes #739. Omit type members declared with using aliases.
```
lcartey, line 2:
```suggestion
  - Fixes #739. Correctly detect template parameters specified in using alias base types, e.g. `using T1 = some_type<T>::Type;`.
```

lcartey (changes requested):
Thanks for this! I think we can actually do better, and retain the true positive while removing the false positive in this case.

fjatWbyT:
Thank you very much, @lcartey! I adopted the suggestions in the PR as well as tested it locally and it is indeed working as you indicated.

Great that it can be updated to get rid of the false positive while retaining the true positive, thanks again!

lcartey (approved):

PR #750, opened 2024-10-16
Implement InvalidMemory3, Rule 18-8 amendment.

MichaelRFairhurst:
## Description

Implements package InvalidMemory3, which detects pointer-to-array conversions and modification of array values with temporary lifetimes, (`RULE-18-9`) and bans pointers to variably modified types (`RULE-18-10`).

`RULE-18-10` is marked as "split" from `RULE-18-8` in the amendments file, so update `RULE-18-8` to _only_ detect _allocating_ VLA declarations. In the process, remove potential false positives around incomplete array parameter types, eg, `f(int arr[])`. To protect against the same false positives from occurring in `RULE-18-10`, use locations that correspond with `VlaDimensionStmt` instances.

If it is better to split this PR up, I'm happy to do so!

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - `RULE-18-9`
     - `RULE-18-10`
     - `RULE-18-8`

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


c/misra/src/rules/RULE-18-9/ArrayToPointerConversionOfTemporaryObject.ql
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * A pointer-to-array conversion does not need to be flagged if the result of
+ * that conversion is not used or stored.
+ */
+Expr usedValuesOf(Expr expr) {
+  result = expr.(BinaryOperation).getLeftOperand()
+  or
+  result = expr.(BinaryOperation).getRightOperand()
+  or
+  result = expr.(UnaryOperation).getOperand()
```
lcartey, line 33: The three above can be replaced with `Operation.getAnOperand()`

c/misra/src/rules/RULE-18-9/ArrayToPointerConversionOfTemporaryObject.ql
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Get the expression(s) whose value is "used" by this expression.
+ *
+ * For instance, `(x)` does not use any values, but `x + y` uses `x` and `y`.
+ *
+ * A pointer-to-array conversion does not need to be flagged if the result of
+ * that conversion is not used or stored.
+ */
+Expr usedValuesOf(Expr expr) {
```
lcartey, line 28: The parameter is unused. I'd suggest using the same model as `isStored(..)`, and instead do:
```
predicate isUsed(Expr e) {
```

I would also integrate the `isStored` predicate here.

c/misra/src/rules/RULE-18-9/ArrayToPointerConversionOfTemporaryObject.ql
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Get the expression(s) whose value is stored by this declaration.
+ *
+ * A pointer-to-array conversion does not need to be flagged if the result of
+ * that conversion is not used or stored.
+ */
+predicate isStored(Expr e) {
+  e = any(VariableDeclarationEntry d).getDeclaration().getInitializer().getExpr()
+  or
+  e = any(ClassAggregateLiteral l).getAFieldExpr(_)
```
lcartey, line 49: Integrate this predicate into the `usedValuesOf` predicate.

lcartey (changes requested):
Thanks! Some suggestions and comments to follow up on individual files.

c/misra/src/rules/RULE-18-10/PointersToVariablyModifiedArrayTypesUsed.ql
```diff
@@ -0,0 +1,122 @@
...
+  Expr sizeExpr;
+  CandidateVlaType vlaType;
+  // `before` and `after` are captured for debugging, see doc comment for
+  // `declarationSubsumes`.
+  Location before;
+  Location after;
+
+  InvalidDeclaration() {
+    sizeExpr = any(VlaDimensionStmt vla).getDimensionExpr() and
+    declarationSubsumes(this, sizeExpr.getLocation(), before, after) and
```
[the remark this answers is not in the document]
MichaelRFairhurst: Agreed. I added one.

I also realized this class probably belongs in `VariablyModifiedTypes.qll`, and moved it there under the name `VmtDeclarationEntry`. The VMT types in that library should really only be used with this kind of double checking. Not bundling this class in with that library increases the risk that that library is misused in the future.

c/misra/test/rules/RULE-18-8/test.c
```diff
@@ -1,15 +1,37 @@
...
   int a5[TEST];   // COMPLIANT
   int a6[1 + 1];  // COMPLIANT
+  int a7[n][1];   // NON_COMPLIANT
+  int(*a8)[n];    // COMPLIANT - pointer to VLA, see RULE-18-10
+
+  extern int e1[]; // COMPLIANT
+
+  // A typedef is not a VLA. However, `VlaDeclStmt`s match the typedef.
+  typedef int vlaTypedef[n]; // COMPLIANT[FALSE_POSITIVE]
+  vlaTypedef t1;             // NON_COMPLIANT[FALSE_NEGATIVE]
```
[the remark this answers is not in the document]
MichaelRFairhurst: I'm happy to do it that way, and I think the first point is a good reason why we would want to do it this way. That said, in my conversations with Robert, our conclusion was that 18-8 is intended to reject allocating VLAs. It's the allocation of a VLA that makes them so dangerous to use. In my head, a VLA is an object in memory, while a VLA type is the static type of that object, and the rule now states that it's VLAs which are banned, not VLA types.

Of course, an unused VLA typedef should be deleted. And most uses of a VLA typedef would be invalid. But in theory, not all. A VLA typedef could be used in a `sizeof()`, or it could be used in a context where it is adjusted to a pointer.

I don't think these cases are likely to come up, I did want to document them correctly in the tests.

Perhaps I should add this explanation in a comment?

c/misra/src/rules/RULE-18-10/PointersToVariablyModifiedArrayTypesUsed.ql
```diff
@@ -0,0 +1,122 @@
...
+      )
+    ) and
+    before.isBefore(inner, _) and
+    inner.isBefore(after, _)
+    or
+    exists(DeclStmt s |
+      s.getADeclaration() = entry.getDeclaration() and
+      before = s.getLocation() and
+      after = before and
+      before.subsumes(inner)
```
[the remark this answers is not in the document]
MichaelRFairhurst: We can exclude that case. In general, it'd be easy to exclude all cases where the declaration entry isGeneratedByMacro.

However, it's worth noting that the false positives are incomplete array types. Variables can't be declared with incomplete array types. So the cases where this would flag multiple things because they all match would be a pretty unusual case where, for instance, multiple declared variables have function types with incomplete array types as parameters.

c/misra/test/rules/RULE-18-8/test.c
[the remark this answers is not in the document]
lcartey: Re-reading the rule, I think your interpretation as it intending to identify allocations is correct. The rule starts by saying:
> Variable-length arrays are specified when the size of an array declared in a block or function prototype is not an integer constant expression.

Which would exclude an unused typedef, or a typedef that is not used to declare an array in a block or function prototype. This was a change from the previous interpretation for Rule 18.8 in MISRA C 2012.

I think there's two options for how we handle this:
 1. We continue to flag the typedef, but only report it if we find it used as the type in a variable declaration in a "valid" location - e.g. where there exists a `Variable` whose type is the typedef (or derived from the typedef?). We could add the `Variable`s as placeholders in this case.
 2. We re-jig the query to report `Variable`s, where the `Variable` is the `getVariable()` of the `VlaDeclStmt` or where the variable type is a use of a the `TypeDef` associated with a `VlaDeclStmt`.

If you've used a `typedef` I think it's likely you're intending to use the type in multiple places, so I would say that's an argument to continue to flag the `typedef`. However, it may not be that common in practice, so it's possibly a premature optimization in reporting.

cpp/common/src/codingstandards/cpp/VariablyModifiedTypes.qll
```diff
@@ -0,0 +1,214 @@
+import cpp
+
+/**
+ * A declaration involving a variably-modified type.
+ */
+class VmtDeclarationEntry extends DeclarationEntry {
```
lcartey, line 6: I think we're only interested in variable declarations, right?
```suggestion
class VmtDeclarationEntry extends VariableDeclarationEntry {
```

cpp/common/src/codingstandards/cpp/VariablyModifiedTypes.qll
```diff
@@ -0,0 +1,214 @@
...
+ * For parameters, the `before` and `end` `Location` objects will be
+ * constrained to the closest earlier element (parameter or function body),
+ * these values can therefore be captured and inspected for debugging.
+ *
+ * For declarations which occur in statements, the `before` and `end`
+ * `Location` objects will be both constrained to be equal, and equal to,
+ * the `Location` of the containing `DeclStmt`.
+ */
+private predicate declarationSubsumes(
+  DeclarationEntry entry, Location inner, Location before, Location after
```
lcartey, line 50: Same here.
```suggestion
  VariableDeclarationEntry entry, Location inner, Location before, Location after
```

c/misra/src/rules/RULE-18-10/PointersToVariablyModifiedArrayTypesUsed.ql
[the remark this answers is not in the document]
lcartey: Could we detect when this issue had occurred by determining if the number of VlaDeclStmt and the number candidates that occur at the same location match? Then avoid reporting in the case where it doesn't?

c/misra/src/rules/RULE-18-9/ArrayToPointerConversionOfTemporaryObject.ql
```diff
@@ -0,0 +1,70 @@
...
+
+from
+  TemporaryLifetimeArrayAccess fa, TemporaryLifetimeExpr temporary,
+  ArrayToPointerConversion conversion
+where
+  not isExcluded(conversion, InvalidMemory3Package::arrayToPointerConversionOfTemporaryObjectQuery()) and
+  fa.getTemporary() = temporary and
+  conversion.getExpr() = fa and
+  isUsedOrStored(temporaryObjectFlowStep*(conversion.getExpr()))
+select conversion, "Array to pointer conversion of array $@ from temporary object $@",
```
lcartey, line 69: Add a full stop to the alert messages, as per the style guide.

c/misra/src/rules/RULE-18-10/PointersToVariablyModifiedArrayTypesUsed.ql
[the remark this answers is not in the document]
MichaelRFairhurst: Very reasonable, will do!

c/misra/test/rules/RULE-18-8/test.c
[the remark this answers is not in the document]
MichaelRFairhurst: Resolved via 2!

cpp/common/src/codingstandards/cpp/VariablyModifiedTypes.qll
MichaelRFairhurst: This actually causes the query to exclude the typedef results, since those also use `DeclarationEntry`s

lcartey: Ah, makes sense.

lcartey (approved):

Issue #768, opened 2024-10-21
`RULE-5-4`:  Ignore macros defined on different compilation paths


lcartey:
### Affected rules
 - `RULE-5-4`

### Description

Macros defined on different compilation paths do not fall afoul of this rule - the first macro must be defined in the same compilation to trigger the rule. We should consider:

 * Conditional includes
 * Conditional macro assignment e.g. macros defined in different branches of a `#ifdef`/`#elsif`.
 * Macros defined inside an `#ifndef <MACRO_NAME>` check for the same name.

### Example
```cpp
#ifdef FOO
#define X 1 // COMPLIANT
#else
#define X 2 // COMPLIANT
#endif
```

PR #769, opened 2024-10-21
`RULE-5-4`: Exclude results which do not occur in the same compilation, improve alert message

lcartey:
## Description

 * Fixes https://github.com/github/codeql-coding-standards/issues/768 by
   - Excluding results which occur within an `#ifndef` for the same macro
   - Excluding mutually exclusive definitions in the same file, e.g. where two definitions are guarded by different cases in a sequence of `#if`,`#elif`, `#else` branches.
   - Only report cases where there exists a file which unconditionally includes the file containing the first macro and the file containing the second macro.
   - If both macros are expanded at least once, we enforce that there exists a single link target in which both macros were expanded.
 * Improves the alert message in the case that the size limit is not hit. This makes it easier to understand why we've flagged this particular case.

Note: ee use locations of multiple elements in this query, which required some careful performance optimisations to perform well across different codebase.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - `RULE-5-4`

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


cpp/common/src/codingstandards/cpp/PreprocessorDirective.qll
```diff
@@ -40,3 +40,71 @@ class PreprocessorIfOrElif extends PreprocessorBranch {
...
+/**
+ * Holds if the pair of macros are "conditional" i.e. only one of the macros is followed in any
+ * particular compilation of the containing file.
+ */
+predicate mutuallyExclusiveMacros(Macro firstMacro, Macro secondMacro) {
+  exists(
+    PreprocessorBranchDirective b1, PreprocessorBranchDirective b2, string filepath,
+    int b1StartLocation, int b2StartLocation
+  |
+    isBranchDirectivePair(b1, b2, filepath, b1StartLocation, b2StartLocation) and
```
nicolaswill, line 106: `b1StartLocation`, `b2StartLocation`, and `filepath` can be replaced with don't-care expressions.

cpp/common/src/codingstandards/cpp/PreprocessorDirective.qll
```diff
@@ -40,3 +40,71 @@ class PreprocessorIfOrElif extends PreprocessorBranch {
...
+    startline < macroline and
+    endline > macroline
+  )
+}
+
+/**
+ * Holds if the pair of macros are "conditional" i.e. only one of the macros is followed in any
+ * particular compilation of the containing file.
+ */
+predicate mutuallyExclusiveMacros(Macro firstMacro, Macro secondMacro) {
```
nicolaswill, line 101: I think `mutuallyExclusiveBranchDirectiveMacros` would make the purpose of this predicate clearer.

nicolaswill (changes requested):

cpp/common/src/codingstandards/cpp/PreprocessorDirective.qll
lcartey: Good spot.

nicolaswill (approved):

PR #784, opened 2024-10-25
Implement RULE-2-8, project should not contain unused object definitions

MichaelRFairhurst:
## Description

Implement RULE-2-8, project should not contain unused object definitions

Also add a new AlertReporting shared query library for deduplicating results across macro definitions/invocations/etc.

Split `__attribute__((unused))` variables (and similar) to a Strict pair of queries.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - `RULE-2-8`
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


cpp/common/src/codingstandards/cpp/deadcode/UnusedObjects.qll
```diff
@@ -0,0 +1,176 @@
...
+ * This therefore only reports variables (local or top level) which have a definition, and are
+ * unused.
+ */
+class UnusedObjectDefinition extends VariableDeclarationEntry {
+  UnusedObjectDefinition() {
+    not exists(VariableAccess access | access.getTarget() = getVariable()) and
+    getVariable().getDefinition() = this and
+    not this instanceof ParameterDeclarationEntry and
+    not getVariable() instanceof MemberVariable
+  }
```
lcartey, line 22: I think we should review, and potentially integrate, the queries from previous standards that look for unused local variables.

 * [UnusedLocalVariable.ql](https://github.com/github/codeql-coding-standards/blob/main/cpp/autosar/src/rules/M0-1-3/UnusedLocalVariable.ql)
 * [UnusedVariables.qll](https://github.com/github/codeql-coding-standards/blob/main/cpp/common/src/codingstandards/cpp/deadcode/UnusedVariables.qll)

These are both C++ queries, so not everything will be applicable, but notably:
 * Ignoring results in functions that use ASM statements (as we cannot determine if the assembly instructions access the the variable).
 * Consider ignoring results in functions with error expressions.
 * Excluding compiler generated variables.

c/misra/test/rules/RULE-2-8/test.c
```diff
@@ -0,0 +1,113 @@
+// Not a definition, only a declaration:
```
lcartey, line 1: Could you add some examples of `const` variables to this file?

For C++ one of the big problems with unused variable queries has been related to the value of `constexpr` variables being inlined in our AST, appearing to make such variables unused, when they are not. I don't think a similar problem exists with `const` variables, but it would be great to have a test case or two to (a) validate that and (b) protect against any future changes to the AST.

c/misra/src/rules/RULE-2-8/UnusedObjectDefinitionInMacroStrict.ql
```diff
@@ -0,0 +1,27 @@
...
+ * @description A strict query which reports all unused object definitions in macros with
+ *              '__attribute__((unused))'.
+ * @kind problem
+ * @precision very-high
+ * @problem.severity recommendation
+ * @tags external/misra/id/rule-2-8
+ *       maintainability
+ *       performance
+ *       external/misra/c/2012/amendment4
+ *       external/misra/c/strict
```
lcartey, line 13: I think it makes sense to use the tag "strict" here to cover these cases. Although we've added strict tags before (to AUTOSAR queries), we've never integrated them into our query suites, or provided user facing recommendations around them.

There are a few ways we could consider doing so, but I think my preference would be to:

 1. Exclude strict queries from the default query suites (for only MISRA C to start with - the AUTOSAR strict queries would need to be re-reviewed)
 2. Provide a strict query suite separately.
 3. Update the `user_manual.md` to discuss the option of "strict" mode.
 4. Update the `development_handbook.md` to discuss the criteria for a query to be considered for "strict" mode.

We should also, in this case, put into the "implementation_scope" why we consider this to be a "strict" query - i.e. why it is safe to default disable it.

The alternative to a "strict mode tag", that I think is worth mentioning here, is to provide a default "deviations" configuration that disables such queries that we would consider to be "strict".

lcartey (commented):
Thanks! Providing some initial feedback. I still need to review the deduplicate macro code.

cpp/common/src/codingstandards/cpp/deadcode/UnusedObjects.qll
MichaelRFairhurst: Integrated, thanks!

c/misra/src/rules/RULE-2-8/UnusedObjectDefinitionInMacroStrict.ql
MichaelRFairhurst: Offered a strict definition of what should be considered a strict query, hopefully it LGTY

MichaelRFairhurst:
Apologies, I forgot I had unaddressed feedback here! Incorporated.

c/misra/src/rules/RULE-2-8/UnusedObjectDefinitionInMacroStrict.ql
lcartey: Looks good, thanks!

cpp/common/src/codingstandards/cpp/deadcode/UnusedObjects.qll
```diff
@@ -0,0 +1,180 @@
...
+
+/* Module config to use the `DeduplicateUnusedMacroObjects::Report` module */
+module ReportDeadObjectInMacroConfig implements MacroReportConfigSig<UnusedObjectDefinition> {
+  bindingset[description]
+  string getMessageSameResultInAllExpansions(Macro m, string description) {
+    result = "Macro '" + m.getName() + "' defines unused object '" + description + "'."
+  }
+
+  string getMessageVariedResultInAllExpansions(Macro m) {
+    result = "Macro '" + m.getName() + "' defines unused object of varied names, for example, '$@'."
```
lcartey, line 163: Suggestion on wording:
```suggestion
    result = "Macro '" + m.getName() + "' defines unused object with an invocation dependent name, for example, '$@'."
```

cpp/common/src/codingstandards/cpp/alertreporting/DeduplicateMacroResults.qll
```diff
@@ -0,0 +1,379 @@
...
+ *
+ *   import DeduplicateFooInMacros::Report<InvalidFooInMacroReportConfig> as Report;
+ *
+ *   from Report::ReportResultInMacro report
+ *     where not excluded(report.getPrimaryElement(), ...)
+ *   select report.getPrimaryElement(), report.getMessage(), report.getOptionalPlaceholderLocation(),
+ *     report.getOptionalPlaceholderMessage()
+ * ```
+ *
+ * Note that this does *not* (currently) generate a result for elements not contained by a macro.
```
lcartey, line 119: I'm wondering if it would create a cleaner interface to do both in this library? That would reduce the number of queries required in this case to two queries, instead of four. It would also mean we could replace/integrate `MacroUnwrapper`, which is only currently used by one query.

cpp/common/src/codingstandards/cpp/alertreporting/DeduplicateMacroResults.qll
```diff
@@ -0,0 +1,379 @@
...
+  /**
+   * A macro invocation for which the target macro does not always produce a `ResultElement`, but
+   * this specific invocation of it does.
+   *
+   * This is "primary" / most specific macro for these result elements. It will also does not match
+   * `MacroInvocation`s inside of a `MacroInvocation` of a `Macro` which always produces a
+   * `ResultElement`, indicating that the real problem lies with that other `Macro` instead of with
+   * this particular invocation.
+   */
+  class IsolatedMacroExpansionWithResultElement extends ResultMacroExpansion {
```
lcartey, line 256: I wonder if we should exclude `LibraryMacro`s from this case, for consistency - as results grouped by a library macro will be excluded (because the location is outside the source archive).

See `A5-2-2` for an example where we already do this.

cpp/common/src/codingstandards/cpp/alertreporting/DeduplicateMacroResults.qll
```diff
@@ -0,0 +1,379 @@
...
+ *     where not excluded(report.getPrimaryElement(), ...)
+ *   select report.getPrimaryElement(), report.getMessage(), report.getOptionalPlaceholderLocation(),
+ *     report.getOptionalPlaceholderMessage()
+ * ```
+ *
+ * Note that this does *not* (currently) generate a result for elements not contained by a macro.
+ * To do report such cases, either add support for that in this module, or write a separate query
+ * that reports `InvalidFoo` cases where not `.isInMacroExpansion()`.
+ */
+module DeduplicateMacroResults<
```
lcartey, line 123: I believe we could use this module in `A5-2-2`, which already has some manual handling for duplicate macro results, but could benefit from the improvements in this module. I don't think we need to do it in this PR, but should open an issue to track this as a follow up item.

cpp/common/src/codingstandards/cpp/deadcode/UnusedVariables.qll
```diff
@@ -56,6 +56,16 @@ class PotentiallyUnusedLocalVariable extends LocalVariable {
   }
 }
 
+/**
+ * A `LocalVariable` which is a candidate for being unused, and not defined in a macro.
+ */
+class PotentiallyUnusedLocalVariable extends BasePotentiallyUnusedLocalVariable {
+  PotentiallyUnusedLocalVariable() {
+    // Ignore variables declared in macro expansions
+    not exists(DeclStmt ds | ds.getADeclaration() = this and ds.isInMacroExpansion())
```
lcartey, line 65: The existing queries simply ignore variables declared in macro expansions - we should add an issue to review backporting the macro handling capabilities from this PR.

lcartey (changes requested):
Thanks! A few minor housekeeping bits, and some suggestions around the macro deduplication pieces.

cpp/common/src/codingstandards/cpp/alertreporting/DeduplicateMacroResults.qll
MichaelRFairhurst: What was making this more complicated was `MacroExpansionWithOnlyUnusedObjectDefinition` which made it so some macros were being deduped and others weren't. I don't think that class was adding much UX value...probably some, but probably very marginal. I removed that class in favor of deduping all macro invocations the same way, which still has a case for when only particular invocations of macros result in unused vars. That allowed the deletion of all of this bloat, much better now!

cpp/common/src/codingstandards/cpp/alertreporting/DeduplicateMacroResults.qll
MichaelRFairhurst: I think this already works correctly. It checks that there isn't a more specific ResultMacroExpansion, so this should also find a `LibraryMacro` and report it there.

cpp/common/src/codingstandards/cpp/deadcode/UnusedVariables.qll
MichaelRFairhurst: filed: https://github.com/github/codeql-coding-standards/issues/838

cpp/common/src/codingstandards/cpp/alertreporting/DeduplicateMacroResults.qll
MichaelRFairhurst: Filed https://github.com/github/codeql-coding-standards/issues/839

cpp/common/src/codingstandards/cpp/alertreporting/DeduplicateMacroResults.qll
```diff
@@ -0,0 +1,410 @@
...
+     *  - `report.getPrimaryElement()`
+     *  - `report.getMessage()`
+     *  - `report.getOptionalPlaceholderLocation()`
+     *  - `report.getOptionalPlaceholderMessage()`
+     *
+     * The values returned by these methods are configured by the `MacroReportConfigSig`
+     * signature parameter.
+     */
+    class ReportResult extends TReportResult {
+      string toString() { result = getMessage() }
```
lcartey, line 305: I think we should simplify the `toString()` implementation here to be a string description of the type of report result. Otherwise `Element.toString()` depends on evaluating the entire query to produce the right message, which can create potential performance problems.

cpp/common/src/codingstandards/cpp/alertreporting/DeduplicateMacroResults.qll
lcartey: I was thinking of the following case, where `library.h` is outside the source location for the database:

```cpp
// library.h
#define UNUSED_ALWAYS int x;
#define UNUSED_SOMETIMES int y;

// user.c
# include "third_party/library.h"

void unused_a() {
  UNUSED_ALWAYS;
  UNUSED_SOMETIMES;
  y = 1;
}

void unused_b() {
  UNUSED_ALWAYS;
  UNUSED_SOMETIMES;
}

```

In this case `UNUSED_ALWAYS` will be flagged at the macro definition, and will be excluded as the location of `library.h` is outside the source location.

However, `UNUSED_SOMETIMES` will be reported by this case, but I'm not certain we should, given that the underlying problem _could_ still be in the third party macro definition. For other queries (e.g. A5-2-2) we would exclude this case.

However, on reflection, I think given we've separated out the `UNUSED_ALWAYS` case, I think it's more likely than not that these cases are related to the user context for the macro expansion, rather than the macro itself. So I'm going to suggest we keep this as it is, but consider changing it if we receive FP reports for this case.

lcartey (changes requested):
One performance related comment, otherwise looks good.

lcartey (approved):

cpp/common/src/codingstandards/cpp/alertreporting/DeduplicateMacroResults.qll
lcartey: Ah, I was observing this on `grpc/grpc`, not OpenPilot - it may not reproduce on OpenPilot.

PR #788, opened 2024-10-30
Fix #629 about `A7-1-7`

### LINKED ISSUE #629: `A7-1-7`: Exclude expression statements in macros
### Affected rules
 - `A7-1-7`

### Description

Macro expansion can cause multiple expressions and statements to appear at the same location. We exclude macro expanded declarations, I think we should do the same for expression statements.

Reviewing the query, I think this is actually caused by a bracketing issue:
```ql
    not isAffectedByMacro() and
   // MISSING OPENING BRACKET HERE
    exists(Declaration d |
       ...
    )
    or
    this instanceof ExprStmt and
    not exists(ForStmt f | f.getInitialization().getAChild*() = this) and
    not exists(LambdaExpression l | l.getLambdaFunction().getBlock().getAChild*() = this)
   // MISSING CLOSED BRACKET HERE
```

There's also an interesting thing happening here with locations - as we might expect such macro expansions to by the `not l1 = l2` line in the select clause:
```ql
  exists(Location l1, Location l2 |
    e1.getLocation() = l1 and
    e2.getLocation() = l2 and
    not l1 = l2 and
....
```

The reason this doesn't exclude this case is that when we expand the macro, we may provide different locations for the expressions and statements within, if we can associate them with a specific macro parameter.

### Example

```cpp
#define foo(x, y)                                                              \
  x++;                                                                         \
  y++;

void test_macro() {
  int a = 1;
  int b = 1;
  foo(a, b); // COMPLIANT
}
```

fjatWbyT:
## Description

Fix issue #629 as suggested, adding brackets in `A7-1-7` query. I verified that before adding them, the added test case failed.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

     - `A7-1-7`

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


lcartey (approved):
Looks good, thank you!

PR #791, opened 2024-11-01
Implement Concurrency6 package -- split out Concurrency7

MichaelRFairhurst:
## Description

Implement Concurrency6 package -- split out Concurrency7.

I had added support for `pthread_create`, and left it in. Should I enable support for pthreads in the control flow graph in `Concurrency.qll`?

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - `DIR-5-2`
     - `DIR-5-3`
     - `RULE-12-6`
     - `RULE-21-25`
     - `RULE-22-11`
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


c/misra/src/rules/DIR-5-3/BannedDynamicThreadCreation.ql
```diff
@@ -0,0 +1,29 @@
...
+ * @id c/misra/banned-dynamic-thread-creation
+ * @name DIR-5-3: There shall be no dynamic thread creation
+ * @description Creating threads outside of a well-defined program start-up phase creates
+ *              uncertainty in program behavior and concurrency overhead costs.
+ * @kind problem
+ * @precision low
+ * @problem.severity error
+ * @tags external/misra/id/dir-5-3
+ *       external/misra/c/2012/amendment4
+ *       external/misra/c/audit
```
lcartey, line 11: We don't currently have an audit query suite - could you add one? Thanks!

c/misra/src/rules/DIR-5-3/ThreadCreatedByThread.ql
```diff
@@ -0,0 +1,45 @@
...
+ *       concurrency
+ *       performance
+ *       external/misra/obligation/required
+ */
+
+import cpp
+import codingstandards.c.misra
+import codingstandards.cpp.Concurrency
+
+Function callers(Function f) { result = f.getACallToThisFunction().getEnclosingFunction() }
```
lcartey, line 22: There's already a standard library predicate for this - `Function.calls(Function)`.

c/misra/src/rules/DIR-5-3/ThreadCreatedByThread.ql
```diff
@@ -0,0 +1,45 @@
...
+
+class ThreadReachableFunction extends Function {
```
lcartey, line 24: I think this would be nicer if you rewrote this as a `ThreadRoot` class, that provides accessors `getThreadCreateCall()` and `getAReachableFunction()`.

Then the query could be rewritten as:
```codeql
from ThreadRoot threadRoot, CThreadCreateCall tc
where
  not isExcluded(tc, Concurrency6Package::threadCreatedByThreadQuery()) and
  threadRoot.getAReachableFunction() = tc.getEnclosingFunction()
select tc, "Thread creation call reachable from threaded function '$@' created by call to $@.", threadRoot,
  threadRoot.getName(), threadRoot.getThreadCreateCall() as tcOriginal, tcOriginal.getTarget().getName()
```

Note, I'm also suggesting here tagging in the original `CThreadCreateCall`, so that it's easier for a user to verify that `threadRoot` is actually a threaded function.

c/misra/src/rules/RULE-12-6/AtomicAggregateObjectDirectlyAccessed.ql
```diff
@@ -0,0 +1,35 @@
...
+
+import cpp
+import codingstandards.c.misra
+
+from Expr expr, Field field
+where
+  not isExcluded(expr, Concurrency6Package::atomicAggregateObjectDirectlyAccessedQuery()) and
+  not expr.isUnevaluated() and
+  (
+    exists(FieldAccess fa |
```
lcartey, line 23: Are the `atomic_..` access functions alway guaranteed to be function calls and not macros? Could we get false positives for this query if a standard library decided to implement a macro which did access a field? In general the C library can be implemented with functions or macros. Or, in practice do we expect all standard libraries to implement this using a function call?

c/misra/src/rules/RULE-12-6/AtomicAggregateObjectDirectlyAccessed.ql
```diff
@@ -0,0 +1,35 @@
...
+      expr = fa and
+      fa.getQualifier().getUnderlyingType().hasSpecifier("atomic") and
```
lcartey, line 25: Technically you don't need the getUnderlyingType here

c/misra/src/rules/RULE-12-6/AtomicAggregateObjectDirectlyAccessed.ql
```diff
@@ -0,0 +1,35 @@
...
+      field = fa.getTarget()
+    )
+    or
+    exists(PointerFieldAccess fa |
+      expr = fa and
+      fa.getQualifier().getUnderlyingType().(PointerType).getBaseType().hasSpecifier("atomic") and
```
lcartey, line 31: Do we need to strip top level specifiers here?
```suggestion
      fa.getQualifier().getUnderlyingType().stripTopLevelSpecifiers().(PointerType).getBaseType().hasSpecifier("atomic") and
```

c/misra/src/rules/RULE-21-25/InvalidMemoryOrderArgument.ql
```diff
@@ -0,0 +1,172 @@
...
+ */
+class MemoryOrderedStdAtomicFunction extends Function {
+  int orderParamIdx;
+
+  MemoryOrderedStdAtomicFunction() {
+    exists(int baseParamIdx, int baseParams, string prefix, string suffix |
+      prefix = ["__", "__c11_"] and
+      suffix = ["", ".*", "_explicit"] and
+      (
+        getName().regexpMatch(prefix + ["atomic_thread_fence", "atomic_signal_fence"] + suffix) and
```
lcartey, line 63: We are generating `2 * 2 * 3 = 12` regexes, and a similar number in the other cases. I've seen with other uses of this pattern that it can cause performance problems, so I think we might be better off generating a single regex for each case.

lcartey: However we do this, lets check performance carefully.

c/misra/src/rules/RULE-21-25/InvalidMemoryOrderArgument.ql
```diff
@@ -0,0 +1,172 @@
...
+        or
+        // Clang case, no inserted parameters:
+        // __c11_atomic_load(object, order)
+        suffix = "" and
+        prefix = "__c11_" and
+        orderParamIdx = baseParamIdx
+        or
+        // Non-macro case, may occur in a subset of gcc/clang functions:
+        prefix = "" and
+        suffix = "_explicit" and
```
lcartey, line 103: The `.*` suffix would also capture `_explicit`, but not be handled by this case.

c/misra/src/rules/RULE-21-25/InvalidMemoryOrderArgument.ql
```diff
@@ -0,0 +1,172 @@
...
+  }
+}
+
+module MemoryOrderFlow = DataFlow::Global<MemoryOrderFlowConfig>;
+
+import MemoryOrderFlow::PathGraph
+
+/**
+ * If the node is a memory order constant, or shares a value with a memory order constant, then
+ * return the name of that cnonstant. Otherwise, simply print the node.
```
lcartey, line 152:
```suggestion
 * return the name of that constant. Otherwise, simply print the node.
```

c/misra/src/rules/RULE-21-25/InvalidMemoryOrderArgument.ql
```diff
@@ -0,0 +1,172 @@
...
+where
+  not isExcluded(argument, Concurrency6Package::invalidMemoryOrderArgumentQuery()) and
+  MemoryOrderFlow::flowPath(source, sink) and
+  argument = sink.getNode().asExpr() and
+  value = describeMemoryOrderNode(source.getNode()) and
+  // Double check that we didn't find flow from something equivalent to the allowed value.
+  not value = any(AllowedMemoryOrder e).getName() and
+  function.getACallToThisFunction().getAnArgument() = argument
+select argument, source, sink, "Invalid memory order '$@' in call to function '$@'.", value, value,
+  function, function.toString()
```
lcartey, line 172: In general prefer to use `.getName()` for functions, as `toString()` may be evaluated for all elements in the database.
```suggestion
  function, function.getName()
```

c/misra/src/rules/RULE-21-25/InvalidMemoryOrderArgument.ql
```diff
@@ -0,0 +1,172 @@
...
+    )
+    or
+    // A literal with a disallowed constant integer value
+    exists(Literal literal |
+      node.asExpr() = literal and
+      not literal.getValue().toInt() = any(AllowedMemoryOrder a).getValue().toInt()
+    )
+    or
+    // Everything else: not a memory order constant or an integer valued literal, also exclude
+    // variables and functions, things that flow further back.
```
lcartey, line 127: What else would this be? And if we cannot determine the value for it, presumably we can't tell whether it's in violation or not?

c/misra/test/rules/RULE-12-6/test.c
```diff
@@ -0,0 +1,61 @@
...
+  atomic_s1 = (s1){0};      // COMPLIANT
+  *ptr_atomic_s1 = (s1){0}; // COMPLIANT
+  atomic_s1 = *l2;          // COMPLIANT
+  ptr_atomic_s1 = l2;       // COMPLIANT
+
+  // Banned: circumvents data-race protection, results in UB.
+  atomic_s1.x;          // NON-COMPLIANT
+  ptr_atomic_s1->x;     // NON-COMPLIANT
+  atomic_s1.x = 0;      // NON-COMPLIANT
+  ptr_atomic_s1->x = 0; // NON-COMPLIANT
```
lcartey, line 46: Can we add some user macro-wrapped cases?

lcartey (changes requested):
Looks good overall, thanks!

> Should I enable support for pthreads in the control flow graph in Concurrency.qll?

Either add it, or add an issue to support it. Which queries does it affect? If the change is large, then a separate PR would be better.

c/misra/src/rules/RULE-12-6/AtomicAggregateObjectDirectlyAccessed.ql
MichaelRFairhurst: I **think** we don't have to worry about this in practice, for a few reasons. Here's part 2 of a very interesting [blog post from RedHat](https://developers.redhat.com/blog/2016/01/19/toward-a-better-use-of-c11-atomics-part-2#dynamic_initialization) about locking vs lock free atomics, and some of the features in `atomic.h`.

> Let's now see why the macro wouldn't be useful ... since atomic types aren't structs but basic types, the macro would ... employ some compiler magic to access the hidden mutex and value members of the struct

It seems all current atomic implementations are lock free, and a locking implementation may never exist. It's easier to make lock free atomic pointers. Additionally, an `atomic s` _is_ of type `s` so something like `s->mutex` would require extra effort to support in the language, and implementations would probably instead use a magic builtin compiler function e.g. `__get_mutex(s)` than add support for a magic `s->mutex` on an `int` type, etc.

I'm happy to make this more defensive, though, so it doesn't bite us.

c/misra/src/rules/RULE-12-6/AtomicAggregateObjectDirectlyAccessed.ql
MichaelRFairhurst: Would it not be possible to have, e.g.,

```c
typedef atomic struct { ... } my_atomic_struct;

my_atomic_struct foo;
foo->bar
```

which would require `getUnderlyingType` to detect?

MichaelRFairhurst: I see `hasSpecifier` recursively checks through typedefs, confirmed working with a test, this can indeed be removed. Thanks!

c/misra/src/rules/RULE-12-6/AtomicAggregateObjectDirectlyAccessed.ql
MichaelRFairhurst: Seems like in practice, no. The extractor is stripping the specifiers in this case, I can't make it fail without (even when `getQualifier().(VariableAccess).getTarget().getUnderlingType().(PointerType)` fails).

I added it for resilience anyway.

c/misra/src/rules/RULE-21-25/InvalidMemoryOrderArgument.ql
MichaelRFairhurst: Also, prefix = "" doesn't match the above. This must have been a dead branch. Deleted.

c/misra/src/rules/RULE-21-25/InvalidMemoryOrderArgument.ql
MichaelRFairhurst: Changed to just one regex.

This is now the slowest predicate on glfw:

```
        364ms |     2 | 203ms @ 2    | Element::ElementBase.toString/0#dispred#6e016f86@4589bave
```

c/misra/src/rules/RULE-21-25/InvalidMemoryOrderArgument.ql
MichaelRFairhurst: The goal was to essentially look for proofs that the value is set correctly. All flow should track back to an explicitly correct value, something like a `BinaryOperation` should be incorrect, given that `memory_order_seq_cst` is a simple enum value.

I'm assuming, looking at this again, that this approach is inefficient. I deleted this case, but I opened https://github.com/github/codeql-coding-standards/issues/848 and simply deleted this case for now.

This gives us a false positive for `atomic_thread_fence(g1 + 1)`, which is reasonable.

lcartey: > This gives us a false positive for atomic_thread_fence(g1 + 1), which is reasonable.

I think it's a false negative, rather than a false positive!

> I'm assuming, looking at this again, that this approach is inefficient. I deleted this case, but I opened #848 and simply deleted this case for now.

I don't actually think it would be too inefficient - if I recall correctly, the shared dataflow library has some optimisations enabling the use of forward or reverse flow calculations to help with this case. I think it's fine to follow up in #848, though.

PR #794, opened 2024-11-11
Fix #789: Reduce False positives on A7-1-2 (VariableMissingConstexpr.ql)

### LINKED ISSUE #789: `A7-1-2`:  Query proposes to add constexpr to non-static data members (and other problems)
### Affected rules
 - A7-1-2

#### 1. `constexpr` on non-static data members
This query should exclude variables that are non-static data members. This is because the compiler raises an error in case a non-static data member is marked as `constexpr`.

### Example

```cpp
class a_class
{
  int y;
  public:
  // CodeQL says to make 'x' as constexpr. 
  // But when we do so, there is a compiler error: "error: non-static data member ‘x’ declared ‘constexpr’"
  int x = 10; 
  a_class():y(x)
  {
  }
  int getY()
  {
    return y;
  } 
};

int main()
{
  a_class o;
  return o.getY();
}

```
Fixing as per the [alerts](https://github.com/github/codeql-coding-standards/blob/main/cpp/autosar/test/rules/A7-1-2/VariableMissingConstexpr.expected) raised by CodeQL in the [test case](https://github.com/github/codeql-coding-standards/blob/main/cpp/autosar/test/rules/A7-1-2/test.cpp) of this query (e.g. [here](https://github.com/github/codeql-coding-standards/blob/main/cpp/autosar/test/rules/A7-1-2/test.cpp#L130) or [here](https://github.com/github/codeql-coding-standards/blob/main/cpp/autosar/test/rules/A7-1-2/test.cpp#L55)) also raises the error. 

```bash
test.cpp:55:3: error: non-static data member ‘m2’ declared ‘constexpr’
   55 |   constexpr int m2 = 0; // NON_COMPLIANT

test.cpp:130:3: error: non-static data member ‘m1’ declared ‘constexpr’
  130 |   constexpr int m1 = 0;
```

#### 2. `constexpr` on member variables updated in un-instantiated templates
This query alerts to add constexpr to member variables which are only used in uninstantiated template functions. At first glance, it seems to be correct because since the template function is not instantiated, the member variable isn't getting accessed anywhere and hence becomes a candidate for adding constexpr. However, when the user attempts to fix it, the compiler raises an error and points to the modification in the uninstantiated template function. So should we avoid member variables of templates?

### Example

```cpp
#include <vector>
#include <memory>

template <typename T>
class a_class
{
  // CodeQL says to make "size_" as constexpr because the "add()" function below isn't instantiated and hence, 
  // there is no access of "size_", but if we do this, the compiler complains that "size_" is updated in add.
  // "test.cpp:12:7: error: increment of read-only variable ‘a_class<T>::size_’"
  std::size_t size_ = 0;   
  std::vector<T> values_{};
  public:
  void add(T arg) {
    values_[size_] = arg;
    ++size_;
  }
};

int main()
{
  a_class<int> o;
  return 0;
}
```
Note: The error about `size_` being a non-static data member and constexpr also exists but even if we make it static, we cant fix the error because compiler complains about the update on `size_` in the uninstantiated template function `add()`.

#### 3. `constexpr` for range-based for loop variables
I'm not able to provide a reproducible case but this query alerts on variables used in range-based for loop.

### Example

```cpp
void a_function()
{
  auto origin = std::vector<int> {1, 2, 3, 4, 5, 6, 7, 8, 9};
  auto values = std::vector<std::unique_ptr<int>>{};
  for (auto& value : origin) { // value can be constexpr
    values.emplace_back(std::make_unique<int>(value));
  }
}
```

The reason is that when the range based for loop is de-sugared there is an assignment like the following:
```cpp
for (..)
auto &value = *(__begin) // operator call to * with __begin variable generated by the compiler front end.
...
```
We have an `operator*` as an initializer expression and it has a variable access to the `__begin` variable. The line;  “any(Call call | isCompileTimeEvaluatedCall(call))...” [here](https://github.com/github/codeql-coding-standards/blob/main/cpp/autosar/src/rules/A7-1-2/VariableMissingConstexpr.ql#L51) is True for this case because the `operator *`  is marked as `constexpr` by CodeQL and the getAnArgument() [here](https://github.com/github/codeql-coding-standards/blob/main/cpp/common/src/codingstandards/cpp/Expr.qll#L228) for this call doesn’t return any arguments. Please note that, the qualifier for this call expression is the `__begin` variable which is compiler generated. So should we avoid such calls which has a qualifier that has `VariableAccess` to variables that are compiler generated?

rak3-sh:
## Description

This PR fixes #789. Specifically it implements the below changes.
1. Don't raise alerts on non-static member variables.
2. Check the qualifiers of Call as well in isCompileTimeEvaluatedCall (using isCompileTimeEvaluatedExpression) to avoid raising this alert variables initialized by compiler generated variables. (e.g. range based for-loops). 

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - A7-1-2

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


rak3-sh:
@lcartey : Requesting your review on this PR to fix #789. Let me know your valuable suggestions!

cpp/autosar/src/rules/A7-1-2/VariableMissingConstexpr.ql
```diff
@@ -42,6 +42,12 @@ where
   not v.isConstexpr() and
   not v instanceof Parameter and
   not v.isAffectedByMacro() and
+  // Don't consider non-static member variables.
```
lcartey, line 45: As I suggested on the issue (#789) I think we should report non-static member variables, but only if all the member functions of the class have bodies. That should resolve the issue in the example.

lcartey (changes requested):
Thanks @rak3-sh! I have a suggestion on how to handle the non-static members.

cpp/autosar/src/rules/A7-1-2/VariableMissingConstexpr.ql
rak3-sh: Done in [d584e30](https://github.com/github/codeql-coding-standards/pull/794/commits/d584e309b212ad29e05e74a4a0760fb567ed92e7)

cpp/autosar/test/rules/A7-1-2/test.cpp
```diff
@@ -127,7 +127,7 @@ class MissingConstexprClass {
   MissingConstexprClass(int i) = delete;           // NON_COMPLIANT
   MissingConstexprClass(int i, LiteralClass lc) {} // NON_COMPLIANT
 private:
-  int m1 = 0; // COMPLIANT - non-static member variable
+  int m1 = 0;
```
lcartey, line 130: Needs "NON_COMPLIANT" marker

cpp/autosar/test/rules/A7-1-2/test.cpp
```diff
@@ -138,9 +138,9 @@ class DerivedClass : public virtual VirtualBaseClass {
   DerivedClass(int i) = delete;           // COMPLIANT
   DerivedClass(int i, LiteralClass lc) {} // COMPLIANT
 private:
-  static int m1; // NON_COMPLAINT - static member variable can be constexpr
+  int m1 = 0;
```
lcartey, line 141: Needs "NON_COMPLIANT" marker

lcartey:
@rak3-sh a couple of missing markers, otherwise looks good!

cpp/autosar/test/rules/A7-1-2/test.cpp
rak3-sh: Fixed in [e9b3ebd](https://github.com/github/codeql-coding-standards/pull/794/commits/e9b3ebd1c9e4e92d83ac47a6a65f119709e0a5a5)

cpp/autosar/test/rules/A7-1-2/test.cpp
rak3-sh: Fixed in [e9b3ebd](https://github.com/github/codeql-coding-standards/pull/794/commits/e9b3ebd1c9e4e92d83ac47a6a65f119709e0a5a5)

Issue #326, opened 2023-07-18
Enhancement: add support for suppressing CodeQL alerts on the next line


adam-vonderviszt:
Currently `code-identifier` comments of a deviation are required to be on the same line as the CodeQL alert it refers to, which can cause code-formatters (clang-format etc.) to break very long lines into multiple lines that may require further position adjustment from the user for the deviation comments.
Therefore, a feature for suppressing CodeQL alerts on the NEXT line would be inevitable in the future, similar to clang-tidy`s `NOLINTNEXTLINE` feature.

**Current State:**
Lets say there is a function with a return type that generates a CodeQL alert which should be ignored by a deviation comment:
```
ReturnTypeGeneratesAnAlert MyLongFunction(ParamType1 p1, ParamType2 p2, ParamType3 p3) {  // DeviationCommentForReturnType
...
}
```
 Due to the line length limit, clang-format may reformat the line into:
```
ReturnTypeGeneratesAnAlert
MyLongFunction(ParamType1 p1, ParamType2 p2, ParamType3 p3) {  // DeviationCommentForReturnType
...
}
```
which needs to be manually repositioned to the following way by the user (otherwise the CodeQL alert will be re-triggered):
```
ReturnTypeGeneratesAnAlert // DeviationCommentForReturnType
MyLongFunction(ParamType1 p1, ParamType2 p2, ParamType3 p3) {
...
}
```

**Proposed Solution:**
Introduce `NOCODEQLNEXTLINE` or similar feature for suppressing CodeQL alerts on the next line:
```
// NOCODEQLNEXTLINE(DeviationCommentForReturnType)
ReturnTypeGeneratesAnAlert MyLongFunction(ParamType1 p1, ParamType2 p2, ParamType3 p3) {
...
}

PR #807, opened 2024-12-03
Add support for deviations on next line and multiple lines

lcartey:
## Description

This PR proposes an extension of our deviations mechanism to support easier suppression of results at scale - including applying deviations to a block of code, to code on the next line and to code in macros.

We introduce a C/C++ attribute format - `[[codeql::<standard>_deviation(<code-identifier>)]]` - which applies the deviation distinguished by the `<code-identifier>` to the program element, and those nested within it. This attribute can be applied to:

 * Statements
 * Variables
 * Functions

For example, `[[codeql::autosar_deviation("a0-4-2")]]` would apply the deviation specified using the code identifier `a0-4-2`, related to the `autosar` Coding Standard.

Custom attributes are not universally supported - for example, they were only introduced in C23 - so we also provide an attribute-like syntax to be used in comment markers:

 - `codeql::<standard>_deviation(<code-identifier>)` - the deviation applies to results on the current line (duplication of existing non-annotated support, included for consistency).
 - `codeql::<standard>_deviation_next_line(<code-identifier>)` - this deviation applies to results on the next line.
 - `codeql::<standard>_deviation_begin(<code-identifier>)` - marks the beginning of a range of lines where the deviation applies.
 - `codeql::<standard>_deviation_end(<code-identifier>)` - marks the end of a range of lines where the deviation applies.

This would address https://github.com/github/codeql-coding-standards/issues/326.

## Change request type

 - [ ] Internal documentation
 - [x] External documentation

 - [x] No rules added
     - _rule number here_
     - _rule number here_

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


cpp/common/src/codingstandards/cpp/deviations/CodeIdentifierDeviation.qll
```diff
@@ -0,0 +1,349 @@
...
+class DeviationAttribute extends StdAttribute {
+  DeviationRecord record;
+
+  DeviationAttribute() {
+    this.hasQualifiedName("codeql", supportedStandard() + "_deviation") and
+    // Support multiple argument deviations
+    "\"" + record.getCodeIdentifier() + "\"" = this.getAnArgument().getValueText()
+  }
+
+  DeviationRecord getDeviationRecord() { result = record }
```
MichaelRFairhurst, line 215: should be getADeviationRecord

cpp/common/src/codingstandards/cpp/deviations/InvalidDeviationCodeIdentifier.ql
```diff
@@ -0,0 +1,35 @@
...
+ * @description Deviation code identifiers must be valid.
+ * @kind problem
+ * @problem.severity error
+ */
+
+import cpp
+import CodeIdentifierDeviation
+
+predicate deviationCodeIdentifierError(Element e, string message) {
+  exists(DeviationEnd end |
```
MichaelRFairhurst, line 13: We should also report unclosed `DeviationStart`s

docs/user_manual.md
```diff
@@ -430,7 +431,90 @@ The `process_coding_standards_config.py` has a dependency on the package `pyyaml
...
+ - `codeql::<standard>_deviation_end(<code-identifier>)` - marks the end of a range of lines where the deviation applies.
+
+Here are some examples, using the deviation record with the `a-0-4-2-deviation` code-identifier specified above:
+```cpp
+  long double x1; // NON_COMPLIANT
+
+  long double x2; // a-0-4-2-deviation - COMPLIANT
+  long double x3; // COMPLIANT - a-0-4-2-deviation
+
+  long double x4; // codeql::<standard>_deviation(a-0-4-2-deviation) - COMPLIANT
```
MichaelRFairhurst, line 488: should probably say `codeql::autosar_deviation` here

MichaelRFairhurst (changes requested):
Fun PR to read through, very cool to see the stack as a solution for implementing pairings in an immutable context like CodeQL! Some thoughts on it since its a cool weedy coding task!

cpp/common/src/codingstandards/cpp/deviations/InvalidDeviationCodeIdentifier.md
```diff
@@ -0,0 +1,19 @@
+# Invalid deviation dode identifier
```
fjatWbyT, line 1:
```suggestion
# Invalid deviation code identifier
```

cpp/common/src/codingstandards/cpp/deviations/CodeIdentifierDeviation.qll
lcartey: The intention is that a code identifier is unique to a specific deviation record, but it's not currently enforced. I've opened an issue for that here:
https://github.com/github/codeql-coding-standards/issues/861

cpp/common/src/codingstandards/cpp/deviations/CodeIdentifierDeviation.qll
```diff
@@ -0,0 +1,349 @@
...
+  )
+  or
+  // Next token is end
+  exists(DeviationEnd end, BeginStack prevStack |
+    record = end.getRecord() and
+    hasDeviationCommentFileOrdering(record, end, file, index) and
+    mkBeginStack(record, file, prevStack, index - 1)
+  |
+    // There is, so pop the most recent begin off the stack
+    prevStack = TConsBeginStack(_, stack)
```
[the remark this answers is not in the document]
lcartey: > If I'm not mistaken, this will hold for values I find counter-intuitive when foo is popped from a stack with a head of bar. I would expect record to be the top of stack but instead mkBeginStack(foo, _, TConsBeginStack(bar, _), _) holds.

I'm not sure if I fully follow. The previous call to `mkBeginStack` will enforce that all of `prevStack` is for the same deviation record - i.e. `mkBeginStack(record, file, prevStack, index - 1)` will only hold for `prevStack`s with the same `record`. The `stack` argument is then set to the tail of the `prevStack` to represent the popping, that `stack` argument will also be for the same deviation record. Does that answer your question/concern?

> Another thing I could be incorrect about. But I believe this predicate won't hold once we find an unmatched begin/end, correct? It's hard to know exactly how behavior should degrade in this case. But I might suggest that we leave the current stack unaltered, or pop whatever is on the stack, or return EmptyBeginStack(), so that at least later correctly matched pushes and pops continue to work.

This predicate still holds for unclosed begins. For example, if we had a file (`f1`) with a single begin (`b1`) for a record (`r1`) and no end, the predicate values for that file would be:
```
mkBeginStack(r1, f1, TEmptyBeginStack(), 0)
mkBeginStack(r1, f1, TConsBeginStack(b1, TEmptyBeginStack()), 1)
```
Nothing fails in this predicate, but `isDeviationRangePaired` would not hold for `b1`.

For ends, the failure state is that there's no available begins on the stack - i.e. the stack is empty. That case is handled below by propagating the empty stack to the next case. For example, if we consider a different file (`f2`) and record (`r2`), where there's an end (`e2`) without a begin, the predicate values would be:
```
mkBeginStack(r2, f2, TEmptyBeginStack(), 0)
mkBeginStack(r2, f2, TEmptyBeginStack(), 1)
```
If the file had further begins/ends that were correctly paired, those would be handled under the current scheme.

cpp/common/src/codingstandards/cpp/deviations/CodeIdentifierDeviation.qll
```diff
@@ -0,0 +1,349 @@
...
+    or
+    // Error, no begin on the stack, ignore and continue
+    prevStack = TEmptyBeginStack() and
+    stack = TEmptyBeginStack()
+  )
+}
+
+newtype TBeginStack =
+  TConsBeginStack(DeviationBegin begin, TBeginStack prev) {
```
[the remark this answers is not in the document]
lcartey: I don't think we need to store it - the record is implied by the `DeviationBegin`, and the `TBeginStack`. The only reason to add it here would be to improve the evaluation performance, but I haven't seen that to be necessary.

cpp/common/src/codingstandards/cpp/deviations/CodeIdentifierDeviation.qll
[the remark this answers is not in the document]
lcartey: Ok, I've made some minor changes - is that any clearer?

cpp/common/src/codingstandards/cpp/deviations/CodeIdentifierDeviation.qll
```diff
@@ -0,0 +1,349 @@
...
+    exists(File file, int index |
+      hasDeviationCommentFileOrdering(begin.getRecord(), begin, file, index) and
+      mkBeginStack(begin.getRecord(), file, prev, index - 1)
```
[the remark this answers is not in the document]
MichaelRFairhurst: I see that we're selecting DeviationBegin so this isn't true.

cpp/common/src/codingstandards/cpp/deviations/CodeIdentifierDeviation.qll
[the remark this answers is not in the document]
MichaelRFairhurst: Ah, I see. What I was missing is that there is a separate stack for each deviation record, rather than a stack per file. LGTM!

MichaelRFairhurst (commented):
Thanks, LGTM! Just one question left.

cpp/common/src/codingstandards/cpp/deviations/CodeIdentifierDeviation.qll
lcartey: Oh, you're right - let me change that. I'd forgotten I added support for multiple records through multiple arguments.

fjatWbyT:
> ```
> 
> Release                            : v2.40.0
> Platform                           : x86-linux
> Language                           : c
> [ snip ]
> Mean_Predicate_Execution_Time_Ms   : 49.33329796264856
> [ snip ]
> Mean_Query_Execution_Time_s        : 0.0493332979626485
> [ snip ]
> 
> ```

While going through the first table, I think that `Mean_Query_Execution_Time_s` is `Mean_Predicate_Execution_Time_s`.

The salient result I noticed in 807 is in the second table:

> ```
> Release           : 807
> Run               : 2025-02-18_23-18-05
> Platform          : x86-linux
> Language          : c
> Suite             : misra-default
> Predicate         : ResourceLeakAnalysis::ResourceLeakConfigSig[MutexObjectsNotAlwaysUnlocked::MutexLeakConfig]::getAnAlias/1#e76446be
> Execution_Time_Ms : 104697
> ```

MichaelRFairhurst (approved):
LGTM!

MichaelRFairhurst:
@fjatWbyT yes, good catch. That was introduced past the last release (by me!) and was fixed. Overall, esp. with that fixed, everything here looks great to me.

PR #814, opened 2024-12-10
Exclude rvalue references from `const` in AUTOSAR rule 7-1-1.

fjatWbyT:
## Description

In the updated query for `A7-1-1`, unmodified objects (including arguments) that aren't declared `const` are warned about missing `const`.

I have observed the warning for rvalue references (in particular, for forwarding/universal references of variadic arguments) in utilities similar to `std::function`, although it isn't trivial to reproduce (i.e. not every universal reference to a variadic argument results in the warning).

This change attempts to avoid that case of false positives by generally excluding rvalue references from being warned about a missing `const`.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

     - `A7-1-1`

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


lcartey:
@fjatWbyT How about this test case?
```cpp
#include <utility>

void variadic_forwarding() {}

template <typename T, typename... Args>
void variadic_forwarding(T &&first, Args &&...rest) {
  first;
  variadic_forwarding(std::forward<Args>(rest)...);
}

int test_variadic_forwarding() { variadic_forwarding(1, 1.1, "a"); }
```

fjatWbyT:
Wow, nice one, @lcartey, thanks a lot for looking into it!

With that test case I have indeed been able to reproduce the alert at the `rest` parameter.

This is the snippet, composed of two independent test cases, I had used:

```cpp
#include <future>
#include <utility>

template<typename F, typename... Args>
constexpr decltype(auto) f(F&& func, Args&&... args)
{
  return std::forward<F>(func)(std::forward<Args>(args)...);
}

class awaitable
{
  awaitable(std::future<void>&& _future) : future(std::move(_future)) {}

 private:
  std::future<void> future;
};

int main()
{
}
```

But didn't get alert for any of `func`, `args`, or `_future`. I am not sure at the moment what is the culprit, but it should now be more feasible to find having the new test case. Thanks again!

I will be updating the pull request shortly adding the test case.

fjatWbyT:
> 
> But didn't get alert for any of `func`, `args`, or `_future`. I am not sure at the moment what is the culprit, but it should now be more feasible to find having the new test case. Thanks again!
> 
Ah, it may have to do with actually instantiating the template. So not for the case with `_future`, but for the other case getting it at `args` making an actual call to `template f`.

cpp/autosar/test/rules/A7-1-1/test.cpp
fjatWbyT, line 1: I verified the test failed without the change in the previous commit due to
```
+| test.cpp:93:47:93:50 | rest | Non-constant variable rest points to an object and is not modified. |
```

cpp/autosar/test/rules/A7-1-1/test.cpp
```diff
@@ -83,4 +83,16 @@ template <bool... Args> extern constexpr bool recurse_var = true; // COMPLIANT
 template <bool B1, bool... Args>
 extern constexpr bool recurse_var<B1, Args...> = B1 &&recurse_var<Args...>;
 
-void fp_621() { recurse_var<true, true, true>; }
\ No newline at end of file
+void fp_621() { recurse_var<true, true, true>; }
+
+#include <utility>
```
fjatWbyT, line 88: I added all at the end of the file to minimize updates to the expected files due to line numbers. But not sure if due to (auto) formatting rules or else it would be preferable otherwise.

cpp/autosar/test/rules/A7-1-1/test.cpp
```diff
@@ -83,4 +83,16 @@ template <bool... Args> extern constexpr bool recurse_var = true; // COMPLIANT
...
+
+void variadic_forwarding() {}
+
+template <typename T, typename... Args>
+void variadic_forwarding(T &&first, Args &&...rest) {
```
fjatWbyT, line 93:
```suggestion
void variadic_forwarding(T &&first, Args &&... rest) {
```

lcartey (approved):
Looks good, thanks!

cpp/autosar/test/rules/A7-1-1/test.cpp
```diff
@@ -83,4 +83,16 @@ template <bool... Args> extern constexpr bool recurse_var = true; // COMPLIANT
...
-void fp_621() { recurse_var<true, true, true>; }
\ No newline at end of file
+void fp_621() { recurse_var<true, true, true>; }
+
+#include <utility>
+
+void variadic_forwarding() {}
+
+template <typename T, typename... Args>
+void variadic_forwarding(T &&first, Args &&... rest) {
```
lcartey, line 93:
```suggestion
void variadic_forwarding(T &&first, Args &&...rest) {
```

PR #852, opened 2025-02-06
M5-3-1: Exclude unknown types

### LINKED ISSUE #851: `M5-3-1`: False positives in unevaluated contexts associated with uninstantiated templates
### Affected rules
 - `M5-3-1`

### Description

This rule is reporting false positives in unevaluated contexts associated with uninstantiated templates. For example:

 * `noexcept` specifiers of functions.
 * `static_assert`s within template functions (not consistently).

In these cases, the type of the operand of the boolean operator is `UnknownType`, which is flagged by the query as not boolean.

In principle, we should be able to exclude these cases by specifying `not operand.isFromUninstantiatedTemplate(_)`. However, this doesn't apply to e.g. `noexcept` specifiers. We also see examples where `static_asserts` for the uninstantiated template are orphaned, and not associated with the uninstantiated template.

The fix is to exclude operands with the type `UnknownType`.

### Example

```cpp
template <typename T> constexpr bool some_variable_template_v = false;
template <> constexpr bool some_variable_template_v<int> = true;

template <typename S>
void template_with_no_except()
   noexcept(some_variable_template_v<S> && true) { // COMPLIANT
}
void test_template() { template_with_no_except<int>(); }
```

lcartey:
## Description

Fixes #851.

Ensure we consistently exclude unknown results for unevaluated contexts in uninstantiated templates.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - `M5-3-1`

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 

### DIFF
--- a/cpp/autosar/src/rules/M5-3-1/EachOperandOfTheOperatorTheLogicalAndOrTheLogicalOperatorsShallHaveTypeBool.ql
+++ b/cpp/autosar/src/rules/M5-3-1/EachOperandOfTheOperatorTheLogicalAndOrTheLogicalOperatorsShallHaveTypeBool.ql
@@ -25,6 +25,11 @@ where
   ) and
   t = operand.getType() and
   not t.getUnderlyingType().getUnspecifiedType() instanceof BoolType and
+  // Ignore cases where the type is unknown - this will typically be in unevaluated contexts
+  // within uninstantiated templates. It's necessary to check for this explicitly because
+  // not all unevaluated contexts are considered to be `isFromUninstantiatedTemplate(_)`,
+  // e.g. `noexcept` specifiers
+  not t instanceof UnknownType and
   not exists(ReferenceType rt |
     rt = t.getUnderlyingType().getUnspecifiedType() and rt.getBaseType() instanceof BoolType
   ) and
--- a/change_notes/2025-02-06-m5-3-1-exclude-unknown-type.md
+++ b/change_notes/2025-02-06-m5-3-1-exclude-unknown-type.md
@@ -1,0 +1,2 @@
+ - `M5-3-1` - `EachOperandOfTheOperatorOfTheLogicalAndOrTheLogicalOperatorsShallHaveTypeBool.ql`:
+   - Consistently exclude results in unevaluated contexts associated with uninstantiated templates, for example `noexcept` specifiers and `static_assert`s.
--- a/cpp/autosar/test/rules/M5-3-1/test.cpp
+++ b/cpp/autosar/test/rules/M5-3-1/test.cpp
@@ -26,3 +26,12 @@ void f() {
   A<int> a;
   a.test1();
 }
+
+template <typename T> constexpr bool some_variable_template_v = false;
+template <> constexpr bool some_variable_template_v<int> = true;
+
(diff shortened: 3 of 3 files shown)


cpp/autosar/src/rules/M5-3-1/EachOperandOfTheOperatorTheLogicalAndOrTheLogicalOperatorsShallHaveTypeBool.ql
```diff
@@ -25,6 +25,11 @@ where
   ) and
   t = operand.getType() and
   not t.getUnderlyingType().getUnspecifiedType() instanceof BoolType and
+  // Ignore cases where the type is unknown - this will typically be in unevaluated contexts
+  // within uninstantiated templates. It's necessary to check for this explicitly because
+  // not all unevaluated contexts are considered to be `isFromUninstantiatedTemplate(_)`,
+  // e.g. `noexcept` specifiers
+  not t instanceof UnknownType and
```
MichaelRFairhurst, line 32: This is great, and we should do this...probably just about everywhere!

How about:

```
signature class ExamineType = Type;

module TypeOps<ExamineType T> {
  final class TypeFinal = Type;
  class Subtype extends TypeFinal {
    T superType;
    
    SubType() {
      this instanceof UnknownType
      or
      this = superType
      or
      this.getUnspecifiedType() = superType
    }
  }
}
```

Additional work would have to be done (not right away) to properly support `TypeOps<SpecifiedType>::SubType` and `TypeOps<TypeDefType>::SubType`, and to apply the right subtyping recursively for types like `PointerType`, `FunctionType`, and class types with inheritance.

But generally, `resolveTypedefs()`, `unspecifiedType()`, `stripTopLevelSpecifiers()` are clunky operations on types that don't truly reflect our intention when we use them. Typically when we use them, we're asking, "is this compatible with." I know that subtyping per se is not quite the right term in C/C++, its worth thinking about what relationship(s) we typically ask about in queries.

MichaelRFairhurst (approved):
Looks great, didn't realize we had an `UnknownType`!

Left a comment for discussion but this PR should be merged.

cpp/autosar/src/rules/M5-3-1/EachOperandOfTheOperatorTheLogicalAndOrTheLogicalOperatorsShallHaveTypeBool.ql
lcartey: > This is great, and we should do this...probably just about everywhere!

In general, excluding uninstantiated templates avoids the need to handle `UnknownType`s, but there are exceptions (as in this case). Perhaps we should have a `KnownType` and replace all uses of `Type` with `KnownType` to avoid this problem.

> But generally, resolveTypedefs(), unspecifiedType(), stripTopLevelSpecifiers() are clunky operations on types that don't truly reflect our intention when we use them.

Agreed! Definitely an area we should investigate.

PR #857, opened 2025-02-11
`A5-1-9`: Improve performance, address duplication

### LINKED ISSUE #856: `A5-1-9`: Avoid reporting duplicate results in macro expansions
### Affected rules
 - `A5-1-9`

### Description

A lambda defined in a macro is flagged each time it is expanded, when expanded multiple times.

### Example
```cpp
#define MACRO() [](int i) -> int { return i + 3; }
void test_macros() {
  MACRO(); // COMPLIANT
  MACRO(); // COMPLIANT - no duplication
}
```

lcartey:
## Description

Addresses three performance issues with A5-1-9:
 * `HC_ParamCons` was incorrectly defined, referring to _uses_ of the parameters when calculating the hash, instead of properties of the parameter (name and type). This causes exponential blow up in some databases (with lambdas with a lot of parameters and/or lots of uses).
 * An optimiser issue where `Variable.getName()` was being magic-ked with an insufficiently restrictive set of variables, which ultimately led to the calculation of `Function.getParameter(int i)` creating a cross-product on indexes.
 * `HC_BlockStmt` was calculating for a blocks, but `hashStmtCons` then filtered out any blocks not in lambdas. I've moved this restriction into `HC_BlockStmt`, which has helped enable us to run even on large databases (e.g. grpc/grpc). Ideally we would apply the lambda restriction to all elements in the `LambdaEquivalence.qll`, but this seems to be sufficient for now.

Also fixes #856, by excluding results that come from the same macro expansion.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - `A5-1-9`

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


knewbury01 (approved):
lgtm! thanks!

PR #862, opened 2025-02-21
Fix performance issue in ResourceLeakAnalysis.qll

MichaelRFairhurst:
## Description

After trying a lot of things, this is where I ended up, but open to additional ideas.

Regarding `getAnAliasRecursive` which this PR removes: The intention was to ensure that if code locks `x` which is an alias of `y`, and there's an unlock call on `z` which is alias of `y` (but not `x`), then we don't want to report that. This wasn't previously tested. The newly added test fails even with `getAnAliasRecursive()` as its incorrectly defined. Further, since the identity of an allocation (`TJustResource`) is based on a `DataFlow::Node`, and the out arg dataflow node is different from the expr dataflow node, it's not trivial to handle this edge case.

Constraining `getAnAlias(node)` isn't simply constrained to `isAllocate(node)` for two minor reasons:
- Since the expensive behavior is in the default predicate on a signature module, and CodeQL doesn't support `Self::`, it would require additional refactoring and increase complexity
- In the future we may want to reintroduce `getAnAliasRecursive` or something like it. This could require something like `Config::getAnAlias+(node)`, so that can't require its argument to be an allocate in that case.

I then experimented with `inline_late`, `pragma[only_bind_into](...)`, and other various options, but none were providing a benefit. Eventually I copied `getAnAlias()` into `ResourceLeakAnalysis.qll` and noticed that no matter what changes I made, I couldn't get good performance from it on opencv as long as it used GVN, and I didn't have the same issue with `hashCons`. I did also measure a slight performance improvement by adding `bindingset[node]` to `getAnAlias()` and so included that as well.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - `RULE-22-16`
     - `ERR57-CPP`
     - `A15-1-4`

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


knewbury01 (approved):
looks good to me! thanks!

PR #891, opened 2025-04-27
Michaelrfairhurst/compatible types performance fix alt

MichaelRFairhurst:
## Description

Performance fixes for TypeEquality module.

Fix several bad joins, mostly on function parameter indexes. Recursively (top down) find all pairs of type comparisons in a stage 1, then restrict the recursively (bottom up) validate which of those type comparison pairs hold.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - `RULE-8-3`
     - `RULE-8-4`
     - `RULE-23-5`
     - `DCL40-C`

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


c/misra/src/rules/RULE-8-3/DeclarationsOfAFunctionSameNameAndType.ql
```diff
@@ -16,6 +16,12 @@ import cpp
 import codingstandards.c.misra
 import codingstandards.cpp.types.Compatible
 
+predicate interestedInFunctions(FunctionDeclarationEntry f1, FunctionDeclarationEntry f2) {
+  f1.getDeclaration() = f2.getDeclaration() and
+  not f1 = f2 and
+  f1.getDeclaration() = f2.getDeclaration()
```
lcartey, line 22: Redundant condition.

cpp/common/src/codingstandards/cpp/types/Compatible.qll
```diff
@@ -331,26 +376,55 @@ module TypeEquivalence<TypeEquivalenceSig Config, TypeSubset T> {
...
+
+  private predicate interestedInParameterTypes(Type a, Type b) {
+    exists(FunctionDeclarationEntry aFun, FunctionDeclarationEntry bFun, int i |
+      interestedInFunctions(pragma[only_bind_into](aFun), pragma[only_bind_into](bFun)) and
+      a = aFun.getParameterDeclarationEntry(i).getType() and
+      b = bFun.getParameterDeclarationEntry(i).getType()
+    )
+  }
+
   predicate equalReturnTypes(FunctionDeclarationEntry f1, FunctionDeclarationEntry f2) {
```
lcartey, line 410: I think you need to add `interestedInFunctions(f1,f2)` here - the performance logs suggest this predicate is slow.

cpp/common/src/codingstandards/cpp/types/Compatible.qll
```diff
@@ -247,36 +311,17 @@ module TypeEquivalence<TypeEquivalenceSig Config, TypeSubset T> {
...
 
-  /**
-   * A type that is either part of the type subset, or that is reachable from a type in the subset.
-   */
-  private class RelevantType instanceof Type {
-    RelevantType() { exists(T t | typeGraph*(t, this)) }
-
-    string toString() { result = this.(Type).toString() }
+  /** Whether two types will be compared, regardless of order (a, b) or (b, a). */
+  private predicate interestedInUnordered(Type t1, Type t2) {
```
lcartey, line 315: Unused predicate.

cpp/common/src/codingstandards/cpp/types/Compatible.qll
```diff
@@ -204,79 +208,204 @@ signature module TypeEquivalenceSig {
...
-          not t1 instanceof TypedefType and
-          not t2 instanceof TypedefType and
-          LeafEquiv::getEquivalenceClass(t1) = LeafEquiv::getEquivalenceClass(t2)
-          or
-          equalDerivedTypes(t1, t2)
-          or
-          equalTypedefTypes(t1, t2)
-          or
-          equalFunctionTypes(t1, t2)
+  predicate compares(Type t1, Type t2) {
```
lcartey, line 250: `compares` feels like an unintuitive name for this predicate. Perhaps `transitiveInterestedIn` or `nestedInterestedIn`? Or `equalityCandidate`?

cpp/common/src/codingstandards/cpp/types/Compatible.qll
```diff
@@ -204,79 +208,204 @@ signature module TypeEquivalenceSig {
...
    */
-  private class RelevantType instanceof Type {
-    RelevantType() { exists(T t | typeGraph*(t, this)) }
-
-    string toString() { result = this.(Type).toString() }
-  }
-
-  private class RelevantDerivedType extends RelevantType instanceof DerivedType {
-    RelevantType getBaseType() { result = this.(DerivedType).getBaseType() }
+  private predicate recurses(Type t1, Type t2) {
```
lcartey, line 325: Suggest calling this `shouldRecurseOn` or something similar - that would make the body of the `compares` predicate clearer to read.

c/misra/src/rules/RULE-8-4/CompatibleDeclarationFunctionDefined.ql
```diff
@@ -19,6 +19,16 @@ import codingstandards.c.misra
 import codingstandards.cpp.Identifiers
 import codingstandards.cpp.types.Compatible
 
+predicate interestedInFunctions(FunctionDeclarationEntry f1, FunctionDeclarationEntry f2) {
+  f1.getDeclaration() instanceof ExternalIdentifiers and
+  f1.isDefinition() and
+  f1.getName() = f2.getName() and
```
lcartey, line 25: Are there actually any cases where two function declaration entries for the same function declaration have different names?

cpp/common/src/codingstandards/cpp/types/Compatible.qll
```diff
@@ -204,79 +208,204 @@ signature module TypeEquivalenceSig {
...
+            not t1 instanceof TypedefType and
+            not t2 instanceof TypedefType and
+            equalLeafRelation(t1, t2)
+            or
+            equalDerivedTypes(t1, t2)
+            or
+            equalTypedefTypes(t1, t2)
+            or
+            equalFunctionTypes(t1, t2)
+          )
```
lcartey, line 394: I was wondering about rewriting this to avoid the nested if cases.
```codeql
        not Config::resolveTypedefs() and
        equalTypedefTypes(t1, t2)
        or
        Config::resolveTypedefs() and
        (
          equalTypes(t1.(TypedefType).getBaseType(), t2) or
          equalTypes(t1, t2.(TypedefType).getBaseType())
        )
        or
        equalDerivedTypes(t1, t2)
        or
        equalFunctionTypes(t1, t2)
        or
        not t1 instanceof DerivedType and
        not t2 instanceof DerivedType and
        not t1 instanceof TypedefType and
        not t2 instanceof TypedefType and
        equalLeafRelation(t1, t2)
```

It could be further simplified by adding a `LeafType` which is defined as a non-derived, non-typedef type, and could be used in `equalLeafRelation` to avoid the multiple `not ... instanceof ...` checks.

lcartey (changes requested):
Thanks for fixing the performance issues here! Overall looks good. Just some stylistic suggestions - improved naming, and simplification of predicates.

c/misra/src/rules/RULE-8-4/CompatibleDeclarationFunctionDefined.ql
MichaelRFairhurst: Technically this is supported by the dbscheme, and I found examples in cpp code where the decl name is `(null)` and the function name is `(constructor)` etc, but no examples (even with MRVA) of two decls with the same declaration and different names.

Unfortunately removing this condition reverted performance gains. I can get performance back by either:

- adding `pragma[only_bind_out]` to the columns `f1`, `f2` in `Compatible::...::equalReturnTypes(f1, f2)` and `Compatible::...::equalParameterTypes(f1, f2)`
- or adding `bindingset[f1, f2] pragma[inline_late]` to those predicates.
- adding `pragma[nomagic]` to `Compatible:....equalTypes()` (slight performance degradation relative to the other options).

I attempted to find and choose the least risky option, and opted to keep this probably unnecessary comparison in. But the first two options are also reasonable.

cpp/common/src/codingstandards/cpp/types/Compatible.qll
```diff
@@ -353,45 +362,45 @@ module TypeEquivalence<TypeEquivalenceSig Config, interestedInEquality/2 interes
...
-            equalTypedefTypes(t1, t2)
-            or
-            equalFunctionTypes(t1, t2)
-          )
+      else (
+        equalLeafRelation(t1, t2)
+        or
+        equalDerivedTypes(t1, t2)
+        or
+        equalTypedefTypes(t1, t2)
```
lcartey, line 393: `TypeEquivalenceSig::equalTypeDefTypes` has the following comment:
```
   * This predicate is only used if `resolveTypedefs()` is false. If so, then we assume two
   * typedefs are the same if they have the same name and their base types are equal.
```

So I think this case should only hold if `Config::resolveTypedefs()` is `false` - otherwise the resolution will be irrelevant.

c/misra/src/rules/RULE-8-4/CompatibleDeclarationFunctionDefined.ql
```diff
@@ -51,10 +52,12 @@ where
         not FunctionDeclarationTypeEquivalence<TypesCompatibleConfig, interestedInFunctions/2>::equalReturnTypes(f1,
           f2)
         or
+        //not compatibleReturns(f1, f2)
```
lcartey, line 55: Leftover comment?

lcartey (changes requested):
Sorry, one final clarification required, and a few leftover comments - otherwise looks good.

lcartey (approved):

PR #893, opened 2025-04-29
Implement MISRA-C++23 Preprocesser package rules 19-0-4, 19-1-1, and 19-2-1

MichaelRFairhurst:
## Description

Also includes two files in `cpp/util` which I intend to port to `qtil` package (`CondensedList` and `Pair`).

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - `RULE-19-0-4`, `RULE-19-1-1`, `RULE-19-2-1`
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


cpp/misra/src/rules/RULE-19-2-1/NoValidIfdefGuardInHeader.ql
```diff
@@ -0,0 +1,44 @@
...
+    location.isBefore(includeGuard.getIfndef().getLocation())
+    or
+    includeGuard.getEndif().getLocation().isBefore(location)
+  )
+}
+
+from File included
+where
+  not isExcluded(included, PreprocessorPackage::noValidIfdefGuardInHeaderQuery()) and
+  included = any(Compilation c).getAFileCompiled().getAnIncludedFile+() and
```
lcartey: In other Coding Standard queries we use the `HeaderFile` class to identify header files, based on the extension. The advantage of this approach is that we will report contraventions in files in header libraries that are not included within the header library itself. Given that this can be an important case, and for consistency reasons, I would suggest we switch definitions.

cpp/misra/src/rules/RULE-19-2-1/NoValidIfdefGuardInHeader.ql
```diff
@@ -0,0 +1,44 @@
...
+  not exists(CorrectIncludeGuard includeGuard |
+    includeGuard.getFile() = included and
+    // Stricter: define is before all other contents
+    not included
+        .getATopLevelDeclaration()
+        .getLocation()
+        .isBefore(includeGuard.getDefine().getLocation()) and
+    // Stricter: do not allow includes outside of the inclusion guard
+    not exists(Include include | isOutside(includeGuard, include.getLocation()))
```
lcartey: My recollection is that these two conditions are already factored in within the the `CorrectIncludeGuard`. Is that not the case? Specifically `startsWithIfndef` has the following:

```ql
  exists(Location loc |
    loc = ifndef.getLocation() and
    loc.getFile() = hf and
    loc.getStartLine() = min(int l | includeGuardRelevantLine(hf, l))
  )
```
Which should enforce that the condition appears on the first meaningful line of the file?

cpp/misra/src/rules/RULE-19-2-1/NoValidIfdefGuardInHeader.ql
```diff
@@ -0,0 +1,44 @@
...
+  )
+select included, "File does not have a well formatted include guard."
```
lcartey: I would suggest reviewing the `IncludeGuardsNotUsed` shared query to see if it could be reused here. It does cover one additional case not covered by this query - the case where the identifier used in an include guard is duplicated across multiple header files.

cpp/misra/test/rules/RULE-19-2-1/valid2.h
```diff
@@ -0,0 +1,8 @@
+// COMPLIANT
```
lcartey, line 1: The rule permits comments anywhere - perhaps add a test where there are additional comments after the `#if` and `#endif`.

cpp/common/src/codingstandards/cpp/util/CondensedList.qll
```diff
@@ -0,0 +1,105 @@
+private import codeql.util.DenseRank
+
+/**
+ * Describes how to construct a condensed list from sparse but orderable data, and how that data
+ * should be connected, with one such list per specified division.
+ */
+signature module CondensedListSig {
```
lcartey: Minor: I would recognise this as a sparse list or sparse array - I don't think I've seen the term condensed list before.

MichaelRFairhurst:
Note: Updated to use qtil.

cpp/misra/src/rules/RULE-19-2-1/NoValidIfdefGuardInHeader.ql
MichaelRFairhurst: `includeGuardRelevantLine` allows `#include` directives before the guard, which 19-2-1 does not (it explicitly shows `<start of file>` preceeds the `#ifndef`, and states that _comments_ are allowed anywhere but gives no exception for `#include`s.

MichaelRFairhurst: Includes are the only separate case, deleted the rest.

cpp/misra/src/rules/RULE-19-2-1/NoValidIfdefGuardInHeader.ql
MichaelRFairhurst: Split the query to reuse `IncludeGuardsNotUsed`

cpp/misra/src/rules/RULE-19-2-1/NoValidIfdefGuardInHeader.ql
```diff
@@ -0,0 +1,23 @@
+/**
+ * @id cpp/misra/no-valid-ifdef-guard-in-header
+ * @name RULE-19-2-1: Precautions shall be taken in order to prevent the contents of a header file being included more
```
mbaluda:
```suggestion
 * @name RULE-19-2-1: Precautions shall be taken in order to prevent a header file being included more than once
```

rule_packages/cpp/Preprocessor.json
```diff
@@ -0,0 +1,97 @@
...
+    "RULE-19-2-1": {
+      "properties": {
+        "enforcement": "decidable",
+        "obligation": "required"
+      },
+      "queries": [
+        {
+          "description": "Precautions shall be taken in order to prevent the contents of a header file being included more than once.",
+          "kind": "problem",
+          "name": "Precautions shall be taken in order to prevent the contents of a header file being included more",
```
mbaluda:
```suggestion
          "name": "Precautions shall be taken in order to prevent a header file being included more than once",
```

PR #897, opened 2025-05-13
Fix typo from `occured` to `occurred`

BaseMax:
Fix typo from `occured` to `occurred`


lcartey (approved):
Thanks for the contribution!

PR #909, opened 2025-06-06
Implement `BannedAPIs` package

lcartey:
## Description

This PR implements the BannedAPIs package.

As many of these rules required the detection of banned functions, I've written a library (`BannedFunctions`) to help identify different uses (e.g. calls, accesses etc.), and to aggregate reports in macro definitions.

I've also updated quite a few of the C++ stubs (unlike the C stubs, these are all written by hand). I suspect there may be some broken tests - will update the PR after CI/CD has run.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - ``
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


cpp/misra/src/rules/RULE-21-10-1/NoVariadicFunctionMacros.ql
```diff
@@ -0,0 +1,57 @@
...
+    element instanceof BuiltInVarArgsEnd and
+    message = "Call to 'va_end'."
+    or
+    element instanceof BuiltInVarArg and
+    message = "Call to 'va_arg'."
+    or
+    element instanceof BuiltInVarArgCopy and
+    message = "Call to 'va_copy'."
+    or
+    element.(TypedefType).getBaseType() instanceof VaListType and
```
MichaelRFairhurst: Not sure if this is worth handling, but type uses are more complex than this.

For instance, a variable or typedef could have type `va_list*`, which would not be detected. There are other missing cases technically (return types, casts, struct literals, function types).

I don't think in this case its of any serious concern, as the va_list cannot be initialized without also triggering a warning for the va_start / _arg calls etc. In theory, if we did want to handle it, I think a `TypeUse` class would make a lot of sense and _may_ be beneficial elsewhere.

cpp/misra/src/rules/RULE-21-10-2/NoCsetjmpHeader.ql
```diff
@@ -0,0 +1,53 @@
...
+import cpp
+import codingstandards.cpp.misra
+import codingstandards.cpp.BannedFunctions
+
+class CSetJmpHeader extends Include {
+  CSetJmpHeader() { this.getIncludeText().regexpMatch("[<\\\"](csetjmp|setjmp.h)[>\\\"]") }
+}
+
+class JmpBufVariable extends Variable {
+  JmpBufVariable() { this.getType().(UserType).hasGlobalOrStdName("jmp_buf") }
```
MichaelRFairhurst: Another example of missing type uses, though in theory inconsequential if we detect calls to setjmp / longjmp

cpp/misra/src/rules/RULE-23-11-1/UseSmartPtrFactoryFunctions.ql
```diff
@@ -0,0 +1,32 @@
...
+
+from ConstructorCall call, Class smartPtrClass
+where
+  not isExcluded(call, BannedAPIsPackage::useSmartPtrFactoryFunctionsQuery()) and
+  smartPtrClass = call.getTarget().getDeclaringType() and
+  (
+    smartPtrClass.hasQualifiedName("std", "shared_ptr") or
+    smartPtrClass.hasQualifiedName("std", "unique_ptr")
+  ) and
+  call.getNumberOfArguments() >= 1 and
```
MichaelRFairhurst: This line is unnecessary

I'd also suggest an explanation on why this matches the desired constructors.

- Every constructor we don't care about takes a reference instead of a pointer (for instance, move constructor, aliasing constructor)
- We wish to match (*p, deleter) and (*p, deleter, alloc) constructors

cpp/misra/src/rules/RULE-23-11-1/UseSmartPtrFactoryFunctions.ql
```diff
@@ -0,0 +1,32 @@
...
+  exists(Type argType |
+    argType = call.getArgument(0).getType().getUnspecifiedType() and
```
MichaelRFairhurst: This can be simply

call.getArgument(0).getType().getUnspecifiedType() instanceof PointerType

rule_packages/cpp/BannedAPIs.json
```diff
@@ -0,0 +1,205 @@
...
+      "queries": [
+        {
+          "description": "Using program-terminating functions like abort, exit, _Exit, quick_exit or terminate causes the stack to not be unwound and object destructors to not be called, potentially leaving the environment in an undesirable state.",
+          "kind": "problem",
+          "name": "Program-terminating functions should not be used",
+          "precision": "very-high",
+          "severity": "error",
+          "short_name": "AvoidProgramTerminatingFunctions",
+          "tags": [
+            "scope/single-translation-unit"
```
MichaelRFairhurst: Tags for maintainability, correctness?

MichaelRFairhurst (changes requested):
Only a couple small changes, and a non-blocking comment about type uses. Looks great!

cpp/misra/src/rules/RULE-21-10-1/NoVariadicFunctionMacros.ql
lcartey: I've modified the query to use the "type uses" library. Note: most uses of `va_list` are generally prohibited - for example, as a return type or a cast. However, there are certainly cases we weren't covering - such as a variable of pointer to `va_list` type - which we're now capturing.

rule_packages/cpp/BannedAPIs.json
lcartey: Added.

PR #919, opened 2025-07-07
Implement `Conversions` rule package

lcartey:
## Description

This PR adds support for an updated `Conversions` package.

As a number of the early rules in the original conversions package - particularly Rule 7.0.6 - turned out to be more complex than expected, I've created a `Conversions2` package and shuffled around the queries in the following way:
 * Rules 7.0.1, 7.0.2, 7.0.3, 7.0.5, 7.0.6, 7.11.3 are retained in the `Conversions` package.
 * Rule 7.0.4 is moved into `Conversions` from `Preconditions` - as the rule concepts related more closely to the other `Conversions` queries than to preconditions.
 * The remaining rules are moved into `Conversions2`.

Implementation notes:
 * This PR introduces the use of the `BigInt` type, to help support analysis of larger integer constants. In CodeQL CLI 2.19.4, this feature is considered "experimental". Therefore, I would recommend we merge the PR to upgrade the CodeQL dependency to 2.20.7 (https://github.com/github/codeql-coding-standards/pull/913) before we merge this PR.
 * Many of the queries in this package are backed by the newly created `BuiltInTypeRules` CodeQL library, which implements the behaviour described in the section `4.7.0 The built-in type rules` of MISRA C++ 2023. These are similar in purpose to the essential type rules from MISRA C, but different enough that we needed to provide a clean implementation.
 * This PR introduces improvements to many of the standard library stubs for C++, which were necessary to support the various test cases for the conversion rules.
 * Rule 7.0.5 has a known limitation related to preprocessor directives.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - `RULE-7-0-1`
     - `RULE-7-0-2`
     - `RULE-7-0-3`
     - `RULE-7-0-4`
     - `RULE-7-0-5`
     - `RULE-7-0-6`
     - `RULE-7-11-3`
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


cpp/misra/src/rules/RULE-7-0-3/NoCharacterNumericalValue.ql
```diff
@@ -0,0 +1,61 @@
...
+    not targetType instanceof CharacterType
+    or
+    // Conversion from non-character type to character type
+    not sourceType instanceof CharacterType and
+    targetType instanceof CharacterType
+    // or
+    // // Conversion between different character types
+    // getTypeCategory(sourceType) instanceof Character and
+    // getTypeCategory(targetType) instanceof Character and
+    // not sourceType = targetType
```
lcartey: Commented out code.

cpp/common/src/codingstandards/cpp/ConstantExpressions.qll
```diff
@@ -0,0 +1,97 @@
+import cpp
```
MichaelRFairhurst, line 1: Consider making a new subdirectory `cpp/common/.../ast/` for this and `Call.qll` etc.

cpp/common/src/codingstandards/cpp/ConstantExpressions.qll
```diff
@@ -0,0 +1,97 @@
...
+          .getTarget()
+          .getInitializer()
+          .getExpr()
+          .getFullyConverted()
+          .getValue()
+          .toBigInt()
+    or
+    result = this.(EnumConstantAccess).getTarget().getValue().toBigInt()
+    or
+    result = -this.(UnaryMinusExpr).getOperand().getFullyConverted().getValue().toBigInt()
```
MichaelRFairhurst, line 55: Should these cases be recursive? e.g.

```
result = this.(NotExpr).getOperand().getFullyConverted().(IntegerConstantExpr).getConstantValue().bitNot()
```

cpp/common/src/codingstandards/cpp/ConstantExpressions.qll
```diff
@@ -0,0 +1,97 @@
...
+    or
+    result = this.(UnaryPlusExpr).getOperand().getFullyConverted().getValue().toBigInt()
+    or
+    result = this.(NotExpr).getOperand().getFullyConverted().getValue().toBigInt().bitNot()
```
MichaelRFairhurst: Looks incorrect, like its performing the `~` binary operation instead of boolean casted `!`.

cpp/common/src/codingstandards/cpp/ConstantExpressions.qll
```diff
@@ -0,0 +1,97 @@
+import cpp
+
+final private class FinalExpr = Expr;
+
+/**
+ * An integer constant expression as defined by the C++17 standard.
+ */
+class IntegerConstantExpr extends FinalExpr {
```
MichaelRFairhurst, line 8: 👀 if we import qtil this can be 👀 

```
class IntegerConstantExpr extends Final<Expr>::Type {
```

cpp/misra/src/codingstandards/cpp/misra/BuiltInTypeRules.qll
```diff
@@ -0,0 +1,290 @@
+/**
+ * A library for utility classes related to the built-in type rules in MISRA C++ 2023 (Section 4.7.0).
```
MichaelRFairhurst, line 2: Consider wrapping this file's contents in a namespace such as `MisraCpp23` to make certain types/predicates/concepts slightly clearer such as `CharacterType`.

cpp/misra/src/codingstandards/cpp/misra/BuiltInTypeRules.qll
```diff
@@ -0,0 +1,290 @@
...
+  FloatingPoint() or
+  Character() or
+  Other()
+
+/**
+ * Gets the type category of a built-in type.
+ *
+ * This does not apply the rules related to stripping specifiers or typedefs, or references.
+ */
+TypeCategory getTypeCategory(BuiltInType t) {
```
MichaelRFairhurst: Minor risk to this API, `getTypeCategory(type)` will be pass type checks even when `type` is not resolved to a built-in type.

Consider renaming to `getBuiltinTypeCategory` and adding a predicate `getTypeCategory(Type t) { result = getBuiltinTypeCategory(getBuiltinType(t))`

cpp/misra/src/codingstandards/cpp/misra/BuiltInTypeRules.qll
```diff
@@ -0,0 +1,290 @@
...
+/**
+ * The signedness of a MISRA C++ 2023 numeric type.
+ */
+newtype Signedness =
+  Signed() or
+  Unsigned()
+
+class CharacterType extends Type {
+  // The actual character type, which is either a plain char or a wide char
+  BuiltInType realType;
```
MichaelRFairhurst: consider making a class `MisraType` or `MisraCpp23::Type` that encapsulates the `realType` field and exposes `getTypeCategory()` that can be a base class of `CharacterType` and `NumericType`

cpp/misra/src/codingstandards/cpp/misra/BuiltInTypeRules.qll
```diff
@@ -0,0 +1,290 @@
...
+
+  CharacterType() {
+    // A type whose type category is character
+    getTypeCategory(realType) = Character() and
+    realType = getBuiltInType(this)
+  }
+
+  Type getRealType() { result = realType }
```
MichaelRFairhurst: The naming here is tricky and confusing. For this I have two thoughts:

Firstly, consider renaming from "real type" to "builtin type."

Secondly, the confusion really lies in the difference between "real type" and `this`. Therefore, we may want to consider changing `this`. One option would be to make `this` extend the `BuiltinType`, and add a member such as `getAReferringType()`. I don't know if this would be clearer? Alternatively, we could create a `newtype` here so that it is not a `Type` at all, and we can give it two members with distinguished names, for instance `getRealType()` and `getExactType()`.

cpp/misra/src/codingstandards/cpp/misra/BuiltInTypeRules.qll
```diff
@@ -0,0 +1,290 @@
...
+   */
+  QlBuiltins::BigInt getIntegralLowerBound() { integralTypeBounds(realType, result, _) }
+
+  Type getRealType() { result = realType }
+}
+
+/**
+ * One of the 10 canonical integer types, which are the standard integer types.
+ */
+class CanonicalIntegerTypes extends NumericType, IntegralType {
```
MichaelRFairhurst: consider moving into a new file such as `cpp/common/src/codingstandards/cpp/types/CanonicalTypes.ql`

cpp/misra/src/codingstandards/cpp/misra/BuiltInTypeRules.qll
```diff
@@ -0,0 +1,290 @@
...
+predicate isAssignment(Expr source, Type targetType, string context) {
+  exists(Expr preConversionAssignment |
+    isPreConversionAssignment(preConversionAssignment, targetType, context) and
+    preConversionAssignment.getExplicitlyConverted() = source
+  )
+}
+
+predicate isPreConversionAssignment(Expr source, Type targetType, string context) {
+  // Assignment expression (which excludes compound assignments)
+  exists(AssignExpr assign |
```
MichaelRFairhurst: It looks like we could make a class such as `MisraAssignExpr` or `MisraCpp23::AssignExpr`, that represents an `Expr` being assigned to an `Element`. Then we could add a predicate such as `getAssignedVariable()` which returns either the `Element` if its a `Variable`, or if the `Element` is a `VariableAccess` it can return the target of that access.

Then the bitfield logic can be conditional only on whether `getAssignedVariable() instanceof BitField`, rather than being repeated for each case.

cpp/misra/src/codingstandards/cpp/misra/BuiltInTypeRules.qll
```diff
@@ -0,0 +1,290 @@
...
+    targetType = f.getType() and
+    not targetType.stripTopLevelSpecifiers() instanceof ReferenceType and
+    context = "return"
+  )
+  or
+  // Switch case
+  exists(SwitchCase case, SwitchStmt switch |
+    case.getExpr() = source and
+    case.getSwitchStmt() = switch and
+    targetType = switch.getExpr().getFullyConverted().getType() and
```
MichaelRFairhurst: I think if the switch expr is a `FieldAccess` referring to a `BitField` then the target type should come from the bit field

cpp/misra/src/codingstandards/cpp/misra/BuiltInTypeRules.qll
MichaelRFairhurst: Reviewing more carefully, it looks like a return type is an "assignable" element, but a return type is not an `Element`, so we may need to add a `newtype` if we go down the path proposed above.

```
newtype TMisra23Assignable =
  // = operator, switch expr
  TAssignableExpr(Expr e) { ... }
  or
  // field inits, parameters, aggregate literals, etc
  TAssignableVar(Variable v) { ... }
  or
  // return types only as other cases require BitField info
  TAssignableReturnType(Type t) { ... }
```

And this type may be able to encapsulate bitfield resolution more consistently.

cpp/misra/src/codingstandards/cpp/misra/BuiltInTypeRules.qll
MichaelRFairhurst: Consider renaming to singular instead of plural

cpp/misra/src/codingstandards/cpp/misra/BuiltInTypeRules.qll
```diff
@@ -0,0 +1,290 @@
...
+    // No smaller integral type can hold the bit field value
+    not exists(CanonicalIntegerTypes other |
+      other.getSize() * 8 >= bf.getNumBits() and
+      other.getSignedness() = result.getSignedness()
+    |
+      other.getSize() < result.getRealSize()
+      or
+      // Where multiple types exist with the same size and signedness, prefer shorter names - mainly
+      // to disambiguate between `unsigned long` and `unsigned long long` on platforms where they
+      // are the same size
```
MichaelRFairhurst: wonderful edge case, I would not have thought of this!! :)

Should this be a member on `CanonicalIntegerType` (for instance, `isMinimal()`) ? It may come up again and I would _not_ have thought of this case offhand. Maybe if it were a member of `CanonicalIntegerType` then the edge case itself would be discoverable.

Perhaps even more of this predicate could be moved to be a member predicate of `CanonicalIntegerType`, such as `isSmallestTypeFitting(int bits, boolean signed)`

cpp/misra/src/rules/RULE-7-0-1/NoConversionFromBool.ql
```diff
@@ -0,0 +1,43 @@
...
+  not (
+    // Exception: equality operators with both bool operands
+    exists(EqualityOperation eq |
+      eq.getAnOperand() = e and
+      eq.getLeftOperand().getType().stripTopLevelSpecifiers() instanceof BoolType and
+      eq.getRightOperand().getType().stripTopLevelSpecifiers() instanceof BoolType
+    )
+    or
+    // Exception: explicit constructor calls
+    exists(ConstructorCall cc | cc.getAnArgument() = e)
```
MichaelRFairhurst: I think this is overly allowing, and should only hold for converting constructors.

Note that I think converting constructors can have multiple parameters when some are optional, and this should only hold for the primary argument

cpp/misra/src/rules/RULE-7-0-1/NoConversionFromBool.ql
```diff
@@ -0,0 +1,43 @@
...
+    or
+    // Exception: assignment to bit-field of length 1
+    exists(AssignExpr assign |
```
MichaelRFairhurst: This should use the assignment logic from BuiltInTypeRules.qll

cpp/misra/src/rules/RULE-7-0-2/NoImplicitBoolConversion.ql
```diff
@@ -0,0 +1,98 @@
...
+ *       scope/single-translation-unit
+ *       external/misra/enforcement/decidable
+ *       external/misra/obligation/required
+ */
+
+import cpp
+import codingstandards.cpp.misra
+
+predicate isInContextualBoolContext(Expr expr) {
+  exists(IfStmt ifStmt | ifStmt.getCondition() = expr) or
```
MichaelRFairhurst, line 20: These cases aren't correct in the presence of multiple conversions.

For instance, in the tests this is being flagged as NON_COMPLIANT

```
int l1[5] = {1, 2, 3, 4, 5};
if (l1) { ... }
```

What happens is we get two conversions, an `ArrayToPointerConversion` and then a conversion from that pointer to bool. In the query, the `ArrayToPointerConversion` is allowed because `isPointerType` holds for an `ArrayType` and `isInContextualBoolContext` holds as well. However, the conversion from that pointer to a boolean does not hold for `isInContextualBoolContext` and thus is flagged.

cpp/misra/src/rules/RULE-7-0-2/NoImplicitBoolConversion.ql
```diff
@@ -0,0 +1,98 @@
...
+predicate isBitFieldOfSizeOne(Expr expr) {
+  exists(BitField bf |
+    expr = bf.getAnAccess() and
+    bf.getNumBits() = 1
+  )
+}
+
+predicate isPointerType(Type t) {
+  t.getUnspecifiedType() instanceof PointerType or
+  t.getUnspecifiedType() instanceof ArrayType or
```
MichaelRFairhurst: It doesn't seem that this case should hold.....

cpp/misra/src/rules/RULE-7-0-3/NoCharacterNumericalValue.ql
```diff
@@ -0,0 +1,61 @@
...
+    rightType = eq.getRightOperand().getType() and
+    leftType.getRealType() = rightType.getRealType()
+  ) and
+  // Exclude unevaluated operands
+  not (
+    expr.getParent*() instanceof SizeofExprOperator or
+    expr.getParent*() instanceof SizeofTypeOperator or
+    expr.getParent*() instanceof TypeidOperator or
+    expr.getParent*() = any(Decltype dt).getExpr() or
+    expr.getParent*() instanceof StaticAssert
```
MichaelRFairhurst: Can be expressed as `expr.isUnevaluated()` I think?

cpp/misra/src/rules/RULE-7-0-3/NoCharacterNumericalValue.ql
```diff
@@ -0,0 +1,61 @@
...
+  ) and
+  // Exclude optional comparisons that don't involve conversion
+  not exists(FunctionCall fc |
+    fc.getTarget().hasName("operator==") and
+    fc.getAnArgument() = expr and
+    fc.getQualifier().getType().hasName("optional")
```
MichaelRFairhurst: This shouldn't have to be specified. The following example

```
std::optional<char> opt;
if (opt == 'c') { ... }
```

should not involve a comparison because it should resolve to the friend function

```
bool operator==(std::optional<char> a, char b);
```

and thus there is no conversion.

It seems like this should not have to be specified.

cpp/misra/src/rules/RULE-7-0-4/InappropriateBitwiseOrShiftOperands.ql
```diff
@@ -0,0 +1,144 @@
...
+ *       scope/single-translation-unit
+ *       external/misra/enforcement/decidable
+ *       external/misra/obligation/required
+ */
+
+import cpp
+import codingstandards.cpp.misra
+import codingstandards.cpp.misra.BuiltInTypeRules
+
+predicate isSignedType(NumericType t) { t.getSignedness() = Signed() }
```
MichaelRFairhurst: should be a member predicate of NumericType

cpp/misra/src/rules/RULE-7-0-4/InappropriateBitwiseOrShiftOperands.ql
```diff
@@ -0,0 +1,144 @@
...
+    // Binary bitwise operators (excluding shift operations) - both operands must be unsigned
+    exists(BinaryBitwiseOperation op |
+      not op instanceof BinaryShiftOperation and
+      op = x and
+      not (
+        isUnsignedType(op.getLeftOperand().getExplicitlyConverted().getType()) and
+        isUnsignedType(op.getRightOperand().getExplicitlyConverted().getType())
+      ) and
+      message =
+        "Binary bitwise operator '" + op.getOperator() + "' requires both operands to be unsigned."
```
MichaelRFairhurst: Ideally this would state whether left or right operand is unsigned, or at least stringify the type of the left and right

cpp/misra/src/rules/RULE-7-0-4/InappropriateBitwiseOrShiftOperands.ql
```diff
@@ -0,0 +1,144 @@
...
+    Expr left, Expr right, NumericType leftType, QlBuiltins::BigInt leftVal, int rightVal,
+    int maxBit
+  |
+    left = shift.getLeftOperand() and
+    right = shift.getRightOperand() and
+    leftType = left.getType() and
+    isConstantExpression(left) and
+    isConstantExpression(right) and
+    isSignedType(leftType) and
+    isValidShiftConstantRange(right, leftType) and
```
MichaelRFairhurst, line 44: Seems redundant

cpp/misra/src/rules/RULE-7-0-5/NoSignednessChangeFromPromotion.ql
```diff
@@ -0,0 +1,215 @@
...
+ *       scope/single-translation-unit
+ *       external/misra/enforcement/decidable
+ *       external/misra/obligation/required
+ */
+
+import cpp
+import codingstandards.cpp.misra
+import codingstandards.cpp.misra.BuiltInTypeRules
+
+abstract class RelevantConversion extends Expr {
```
MichaelRFairhurst: This little lib is awesome and should be "promoted" (pun intended) to a qll !

cpp/misra/src/rules/RULE-7-0-5/NoSignednessChangeFromPromotion.ql
```diff
@@ -0,0 +1,215 @@
...
+
+  abstract Expr getConvertedExpr();
+
+  abstract string getKindOfConversion();
+}
+
+/**
+ * A `Conversion` that is relevant for the rule.
+ */
+abstract class RelevantRealConversion extends RelevantConversion, Conversion {
```
MichaelRFairhurst: I don't love the term "real" here. Perhaps something along the lines of "RelevantnConversionInDb" ?

cpp/misra/src/rules/RULE-7-0-5/NoSignednessChangeFromPromotion.ql
```diff
@@ -0,0 +1,215 @@
...
+
+class ImpliedIntegerPromotion extends RelevantConversion {
+  NumericType fromType;
+
+  ImpliedIntegerPromotion() {
+    (
+      exists(AssignLShiftExpr aop | aop.getLValue() = this) or
+      exists(AssignRShiftExpr aop | aop.getLValue() = this)
+    ) and
+    // Only consider integer promotions from MISRA C++ "numeric types" as per the rule
```
MichaelRFairhurst: I'm not sure how to interpret this.

The std says this "applies to all expressions (including sub expressions) of numeric type," which is actually very vague.

If the intention was that the rule only applies to binop "a ? b" when a and b are numeric before promotion/conversions, then the language about changing type categories wouldn't apply. The rule should be enforced when `a` and `b` are character type ... and pointer type too maybe?

cpp/misra/src/rules/RULE-7-0-5/NoSignednessChangeFromPromotion.ql
```diff
@@ -0,0 +1,215 @@
...
+    fromType.getTypeCategory() = Integral() and
+    // If the size is less than int, then it is an implied integer promotion
+    fromType.getRealSize() < sizeOfInt()
+  }
+
+  override NumericType getFromType() { result = fromType }
+
+  override NumericType getToType() {
+    // Only report the canonical type - e.g. `int` not `signed int`
+    result = result.(IntegralType).getCanonicalArithmeticType() and
```
MichaelRFairhurst: result instanceof CanonicalArithmeticType ?

cpp/misra/src/rules/RULE-7-0-5/NoSignednessChangeFromPromotion.ql
```diff
@@ -0,0 +1,215 @@
...
+      )
+    )
+  }
+
+  override string getKindOfConversion() { result = "Usual arithmetic conversion" }
+}
+
+class IntegerPromotion extends RelevantRealConversion {
+  IntegerPromotion() {
+    // Exclude integer promotions combined with usual arithmetic conversions, which are handled separately
```
MichaelRFairhurst: I don't see where these are handled separately?

cpp/misra/src/rules/RULE-7-0-5/NoSignednessChangeFromPromotion.ql
```diff
@@ -0,0 +1,215 @@
...
+    then
+      // Smallest type that can hold the value of the `fromType`
+      result =
+        min(NumericType candidateType |
+          (
+            candidateType instanceof IntType or
+            candidateType instanceof LongType or
+            candidateType instanceof LongLongType
+          ) and
+          fromType.getIntegralUpperBound() <= candidateType.getIntegralUpperBound()
```
MichaelRFairhurst: this could hold for both `LongType` and `LongLongType` right, as you identified earlier?

cpp/misra/src/rules/RULE-7-0-5/NoSignednessChangeFromPromotion.ql
```diff
@@ -0,0 +1,215 @@
...
+
+  override NumericType getFromType() { result = fromType }
+
+  override NumericType getToType() { result = toType }
+
+  override Expr getConvertedExpr() { result = this.getExpr() }
+}
+
+class UsualArithmeticConversion extends RelevantRealConversion {
+  UsualArithmeticConversion() {
```
MichaelRFairhurst: Just noting that this class will match some cases that aren't usual arithmetic conversions. I think harmlessly.

For example, `ArrayToPointerConversion`, `ReferenceDereferenceExpr` (in the odd scenario of performing binary operations on `int&`).

cpp/misra/src/rules/RULE-7-0-6/NumericAssignmentTypeMismatch.ql
```diff
@@ -0,0 +1,211 @@
...
+  sourceType.getSignedness() = targetType.getSignedness() and
+  sourceType.getRealSize() = targetType.getRealSize()
+}
+
+predicate hasConstructorException(FunctionCall call) {
+  exists(Constructor ctor, Class c |
+    call.getTarget() = ctor and
+    c = ctor.getDeclaringType() and
+    // Constructor callable with single numeric argument
+    ctor.getNumberOfParameters() = 1 and
```
MichaelRFairhurst: We unfortunately also need to consider optional parameters in constructors

cpp/misra/src/rules/RULE-7-0-1/NoConversionFromBool.ql
lcartey: Actually, I don't think we need this condition at all - that's because in the cases where the conversion turns into a constructor call, there is no `Conversion` in our model, so there's no cases to exclude.

I also don't believe the rule text means "converting constructors" in the literal sense, as the standard defines a converting constructor to not have the `explicit` keyword, but the example shows the use of the `explicit` keyword, and it makes sense to include those constructors in addition to the "full" converting constructors.

cpp/misra/src/rules/RULE-7-0-2/NoImplicitBoolConversion.ql
lcartey: Great spot, thanks for picking this up. I ended up fixing this in the query itself, by checking if the unconverted expression is in a contextual bool context.

cpp/misra/src/rules/RULE-7-0-4/InappropriateBitwiseOrShiftOperands.ql
lcartey: This is a predicate with an argument to avoid needing to cast values to `NumericType`s, as would be required with member predicates. I will move these to the BuiltInType library, though.

cpp/misra/src/rules/RULE-7-0-4/InappropriateBitwiseOrShiftOperands.ql
lcartey: It's avoiding `rightVal` being negative, which is otherwise not constrained. The upper bound is redundant, but matches the description of the exception in the rule, so I'm inclined to keep it for clarity.

cpp/misra/src/rules/RULE-7-0-4/InappropriateBitwiseOrShiftOperands.ql
lcartey: Improved the reporting in this case by:
 - Creating a BinaryOperations.qll library to unify the assign/binary op cases.
 - Reporting the operand that is of an incorrect type.
 - Reporting the type of that operand.
 - Refining the wording.

cpp/common/src/codingstandards/cpp/ConstantExpressions.qll
lcartey: Arrgh, yes, this was supposed to refer to `ComplementExpr` rather than `NotExpr`.

cpp/common/src/codingstandards/cpp/ConstantExpressions.qll
lcartey: I don't think this needs to be recursive, as anytime we look back "one step" in the constant expression evaluation, the `getValue()` should return the appropriate value we require - as we typically want the post-conversion value.

cpp/common/src/codingstandards/cpp/ConstantExpressions.qll
lcartey: Ooo, nice - I think there's a few places we could make this update.

cpp/common/src/codingstandards/cpp/ConstantExpressions.qll
lcartey: Let's look at doing that refactoring as a separate PR, as there's many libraries within the cpp/common directory which I think could move into that directory.

cpp/misra/src/codingstandards/cpp/misra/BuiltInTypeRules.qll
lcartey: I've created a shared class `CanonicalIntegralType` in the proposed location. However, I've also retained a new sub-class in this file - `CanonicalIntegerNumericType` - because the existing class extends `NumericType`, which is a MISRA specific concept and excludes `char` types.

cpp/misra/src/codingstandards/cpp/misra/BuiltInTypeRules.qll
lcartey: I've created a `isMinimal()` predicate on the new shared class, and updated the constructor for `CanonicalIntegerNumericType` to filter by minimal types by default, which means we can remove this as an "explicit" case in this code.

I did consider moving more of the implementation, but it's tricky because it relies on `NumericType`.

cpp/misra/src/rules/RULE-7-0-5/NoSignednessChangeFromPromotion.ql
lcartey: That's right. However, I've switched the implementation to use `CanonicalIntegerNumericType` which in the `charpred` enforces `isMinimal` to remvoe the duplication.

cpp/misra/src/rules/RULE-7-0-5/NoSignednessChangeFromPromotion.ql
lcartey: Where both integer promotion and usual arithmetic conversion happen within the same expression, we only have one `Conversion` entry, which would otherwise be captured by both the `IntegerPromotion` and `UsualArithmeticConversion` classes. Per the rule:

> For the usual arithmetic conversions, only the final type of an operand is considered. For example, in the expression u32 + u8, u8 is first converted to signed int through integral promotion before being converted to uint32_t; it is the final type of uint32_t that is considered by this rule.

In other words, because integer promotion happens first, then the usual arithmetic conversion, we only care about the usual arithmetic conversion.

I will update the comment to try to clarify.

cpp/misra/src/rules/RULE-7-0-5/NoSignednessChangeFromPromotion.ql
lcartey: > If the intention was that the rule only applies to binop "a ? b" when a and b are numeric before promotion/conversions, then the language about changing type categories wouldn't apply. 

The numeric types include two type categories - integral and floating point. So it's still possible to have a change in type category. This is also highlighted in the examples, which show usual arithmetic conversions from floats. However, you cannot have integer promotion from a float, so this case is correct to limit to numeric types of integral category.

The only other type category that's relevant is the MISRA character type category (which are treated as arithmetic by C++). However, those conversions are already prohibited by 7.0.3, which prohibits conversions from `char`.

I'll update the comment to try to clarify.

cpp/misra/src/rules/RULE-7-0-5/NoSignednessChangeFromPromotion.ql
lcartey: No, neither did I when I wrote it.

I've now rethought the whole naming scheme in this file, and I've come up with:
 * `IntegerPromotionOrUsualArithmeticConversion` as the new name for `RelevantConversion`.
 * `IntegerPromotionOrUsualArithmeticConversionAsCast` as the new name for `RelevantRealConversion`.

These names are a bit verbose, but I think they are much clearer about what is covered. I've also improved the qldoc for clarity.

Finally, I've made this class extend `Cast` instead of `Conversion` - to handle the point below that we were capturing conversions we shouldn't be here.

cpp/misra/src/rules/RULE-7-0-5/NoSignednessChangeFromPromotion.ql
lcartey: Addressed above - I've made the super class extend `Cast` instead of `Conversion`, which should exclude these cases.

cpp/misra/src/codingstandards/cpp/misra/BuiltInTypeRules.qll
lcartey: I've added a MisraBuiltInType class as suggested.

Note: slightly unusual, but I've made it extend `Element` instead of `Type`. This is to prevent misuse of the `getSize()` predicate, that returns the wrong value for e.g. reference types.

cpp/misra/src/codingstandards/cpp/misra/BuiltInTypeRules.qll
lcartey: I've made some changes here that hopefully improve the situation:
 - Renamed `realType` to `builtInType` internally, and pushed the field up to a new `MisraBuiltInType`.
 - Remove `getRealType()` and replaced it with a `isSameType(..)` predicate for public consumption. There's a private `getBuiltInType()`, but only used to implement `isSameType(..)`.
 - `getRealSize()` is renamed `getBuiltInSize()` and I've made the super type extend `Element` instead of `Type` to prevent misuse of `getSize()`.

I considered the other options, but decided against them because:
 - **`MisraBuiltInType` extending `BuiltInType` - the natural assignment from, say, the `.getType()` of an expression to `MisraBuiltInType` would fail silently if the type was specified etc. There's no easy way to add a compile check that `getAReferringType()` would be used correctly.
 - **`MisraBuiltInType` extending a newtype - this would require an explicit conversion every time you wanted to assign to `MisraBuiltInType`, which would be verbose, but at least would be type checked. If we used `BuiltInType` as the "key" for the newtype, then we would also prevent any confusion around `getSize()`, and no longer need an `isSameType` predicate. However, we would lose the original type with specifiers and typedefs, and that's actually quite useful for reporting purposes. Alternatively, the new type could retain the "original" type, but then we'd still need an `isSameType` predicate. In comparison, the approach I adopted requires no conversion to assign to `MisraBuiltInType`, has no confusion over `getSize()`, and still requires an `isSameType` predicate. Weighing up the pros and cons, I decided to stick with the approach above.

cpp/misra/src/codingstandards/cpp/misra/BuiltInTypeRules.qll
lcartey: I looked into this, but ended up taking a slightly more straightforward approach to reducing the duplication around bitfields. I've refactored the `isAssignment` predicate to first check whether `source` is assigned to `BitField`, and handle that case once, before proceeding to the other cases. I think this achieves the deduplication, but avoids having to change the overall API for checking assignments.

cpp/misra/src/rules/RULE-7-0-5/NoSignednessChangeFromPromotion.ql
lcartey: I've moved this to a lib within the MISRA pack - it's difficult to move out fully to common because it's currently tied to the MISRA concept of NumericTypes.

cpp/misra/src/rules/RULE-7-0-4/InappropriateBitwiseOrShiftOperands.ql
MichaelRFairhurst: Ah, makes sense, thanks Luke!

cpp/misra/src/rules/RULE-7-0-2/NoImplicitBoolConversion.ql
```diff
@@ -0,0 +1,100 @@
...
+      // Exception 1: Static cast to bool from class with explicit operator bool
+      not exists(StaticCast conv |
+        op.isExplicit() and
+        conv.getExpr() = conversionCall and
+        conv.getType().getUnspecifiedType() instanceof BoolType
+      ) and
+      // Exception 2: Contextual conversion from class with explicit operator bool is allowed
+      not (
+        op.isExplicit() and
+        isInContextualBoolContext(conversionCall)
```
MichaelRFairhurst: This should be unconverted, too, right?

cpp/misra/test/rules/RULE-7-0-1/test.cpp
```diff
@@ -0,0 +1,117 @@
...
+  // Conditional operator without conversion - compliant
+  s8a = b1 ? 3 : 7; // COMPLIANT
+
+  // Function parameter without conversion - compliant
+  f1(b1 ? 1 : 0); // COMPLIANT
+
+  // Explicit constructor calls - compliant
+  A l1{true};                  // COMPLIANT
+  A l2(false);                 // COMPLIANT
+  A l3 = static_cast<A>(true); // COMPLIANT
```
MichaelRFairhurst: Add tests for dynamic_, reinterpret_, const_ casts, and C-style casts.

cpp/misra/test/rules/RULE-7-0-2/test.cpp
```diff
@@ -0,0 +1,214 @@
...
+  bool l5 = true;
+
+  l1 = l4 ? l2 : l3;       // NON_COMPLIANT
+  l1 = l5 ? l2 : l3;       // COMPLIANT
+  l1 = (l4 < 5) ? l2 : l3; // COMPLIANT
+}
+
+void test_if_statements() {
+  std::int32_t l1;
+  bool l2 = f2();
```
MichaelRFairhurst, line 83: just pointing this out as an example of LLMs producing strange test code. No changes required. It's just interesting how the LLMs struggle to generate stubs, they really want to write code with interesting dynamic behavior rather than just static properties.

cpp/misra/test/rules/RULE-7-0-2/test.cpp
```diff
@@ -0,0 +1,214 @@
...
+}
+
+void test_pointer_conversions() {
+  if (f3()) { // COMPLIANT - exception #2
+  }
+
+  bool l1 = f3();            // NON_COMPLIANT
+  bool l2 = f3() != nullptr; // COMPLIANT
+
+  if (nullptr) { // NON_COMPLIANT
```
MichaelRFairhurst: I think this should be allowed, but this case has some serious dragons.

Basically, the spec states that a prvalue of `nullptr_t` is not a null pointer value, and only states that null pointer values can generally be converted to bool. It specifically states that a prvalue of `nullptr_t` can only be converted to bool during direct initialization, which is during cases like `new`, `static_cast`, or brace list initializers.

So it seems like `nullptr` would have to be cast to `void*` (or some other type `T*`) in order to be a "null pointer value," which would then convert to false. And such a chain of implicit conversions should be not be well formed...so this shouldn't compile?

...Regardless, it seems that this would fall under the case of a pointer being converted to bool.

While this particular example is obviously contrived, I do worry that it would occur in template metaprogramming and that we should allow this?

cpp/misra/test/rules/RULE-7-0-2/test.cpp
```diff
@@ -0,0 +1,214 @@
...
+  }
+}
+
+void test_assignment_to_bool() {
+  bool l1 = static_cast<bool>(4); // NON_COMPLIANT
```
MichaelRFairhurst, line 135: again, may be worth testing the other types of casts here.

cpp/misra/test/rules/RULE-7-0-3/test.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+void test_optional_comparison() {
+  std::optional<char> l1;
+  if (l1 == 'r') { // COMPLIANT
+  }
+}
+
+void test_unevaluated_operand() {
+  decltype('s' + 't') l1;         // COMPLIANT
+  static_assert(sizeof('x') > 0); // COMPLIANT
```
MichaelRFairhurst: test case is compliant regardless of sizeof()

MichaelRFairhurst (changes requested):
Sorry, just a few more comments based on reviewing the test cases!

Overall looking very good and very very close.

cpp/misra/test/rules/RULE-7-0-2/test.cpp
lcartey: I think you're right. I'll change it to accept nullptr here.

cpp/misra/test/rules/RULE-7-0-1/test.cpp
lcartey: C-Style casts are tested above. `dynamic_cast` and `const_cast` can't be used with booleans in a meaningful way (at least, not that I can think of). I have added `reinterpret_cast` as a `NON_COMPLIANT` case as it converts a bool to a pointer.

I've also added some extra NON_COMPLIANT cases for a class which doesn't have a converting constructor with parameter of type `bool`.

cpp/misra/test/rules/RULE-7-0-2/test.cpp
lcartey: Added C-style casts to the bottom of the test case, but other casts are not relevant here.

cpp/misra/test/rules/RULE-7-0-3/test.cpp
lcartey: It's actually the other way around - the char use in the `sizeof()` is excluded because it's an unevaluated operand, but the `static_assert` isn't unevaluated. I've split the two cases to make that clearer.

PR #922, opened 2025-07-08
Convert more queries to the new dataflow library

jketema:
Commit-by-commit review recommended.

This introduces a temporary copy of the `Concurrency` library using the new dataflow library, so we don't have to convert all the queries at once.


MathiasVP (approved):
LGTM!

PR #925, opened 2025-07-10
Fix False Positives of `M5-0-12`

### LINKED ISSUE #541: `M5-0-12`: Incorrect alerts on assignments of valid numerical values
### Affected rules

 - `M5-0-12`

### Description

This query reports false positives where an explicitly signed or unsigned char is implicilty converted to an explicitly unsigned or signed char.

This can occur when converting a value which is typed as signed/unsigned char. It can also occur in the expansion of a template parameter, as in this case integer literals will be fully converted to the desired type, which also causes false positives.

In addition, this query is currently only detecting assignments to variables. It should be expanded to detect implicit conversions from `char` to `signed char` or `unsigned char` - for example, in function call arguments.

### Example
The following is compliant per example in the standard, however is still seen as a contravention.

```cpp
void test() {
  unsigned char x = 10; // COMPLIANT
  unsigned char y = x; // COMPLIANT[FALSE_POSITIVE]
  // ...and similar for signed chars
}


template <typename T, T value>
class C1 {
  public:
    C1() : m_value(value) {} // COMPLIANT[FALSE_POSITIVE]
  private:
    T m_value;
};

void fp_test() {
  C1<std::uint8_t, 10> l1; 
  C1<std::uint8_t, 10U> l2;
}

template <typename T, T value>
class C2 {
  public:
    C2() : m_value(value) {} // NON_COMPLIANT
  private:
    T m_value;
};

void fp_test() {
  C2<std::uint8_t, 'A'> l1; 
}
```

jeongsoolee09:
## Description

Fix false positives of `M5-0-12` (Closes #541).

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - _rule number here_

## Release change checklist

  - [X] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


jeongsoolee09:
This query runs on openpilot-72d1744 in 17 seconds (on an M1 Max MacBook), and produces one true positive result.

cpp/autosar/src/rules/M5-0-12/SignedCharAndUnsignedCharTypeShallOnlyBeUsedForTheStorageAndUseOfNumericValues.ql
```diff
@@ -16,23 +16,237 @@
...
+
+  TemplateVariable asTemplateVariable() { this = TTemplateVariable(result) }
+
+  string toString() {
+    result = this.asTemplateClass().toString() or
+    result = this.asTemplateFunction().toString() or
+    result = this.asTemplateVariable().toString()
+  }
+
+  Location getLocation() {
```
MichaelRFairhurst: If you added `Element getElement()` then this could be `getElement().getLocation()`, and `toString()` could be `getElement().toString()`

cpp/autosar/src/rules/M5-0-12/SignedCharAndUnsignedCharTypeShallOnlyBeUsedForTheStorageAndUseOfNumericValues.ql
```diff
@@ -16,23 +16,237 @@
 import cpp
 import codingstandards.cpp.autosar
 
-from Variable v, Expr aexp
+newtype TTemplateElement =
```
MichaelRFairhurst: I love that you're solving this directly. I think this type (and TemplateElement) could be moved into `cpp/common/src/codinstandards/cpp` under a qll (like `Templates.qll` or something).

I think it also might be less verbose if you made a common `Element` subclass, like `TemplateElement extends Element` with characteristic predicate `this instanceof TemplateClass or this instanceof TemplateFunction or ...`.

cpp/autosar/src/rules/M5-0-12/SignedCharAndUnsignedCharTypeShallOnlyBeUsedForTheStorageAndUseOfNumericValues.ql
```diff
@@ -16,23 +16,237 @@
...
+    result = this.asTemplateClass().getName() or
+    result = this.asTemplateFunction().getName() or
+    result = this.asTemplateVariable().getName()
+  }
+}
+
+newtype TTemplateInstantiation =
+  TClassTemplateInstantiation(ClassTemplateInstantiation c) or
+  TFunctionTemplateInstantiation(FunctionTemplateInstantiation f) or
+  TVariableTemplateInstantiation(VariableTemplateInstantiation v)
```
MichaelRFairhurst: again, awesome and could be in a shared qll! And I think these could also be a class that extends `Element`?

cpp/autosar/src/rules/M5-0-12/SignedCharAndUnsignedCharTypeShallOnlyBeUsedForTheStorageAndUseOfNumericValues.ql
```diff
@@ -16,23 +16,237 @@
...
+  FunctionTemplateInstantiation asFunctionTemplateInstantiation() {
+    this = TFunctionTemplateInstantiation(result)
+  }
+
+  VariableTemplateInstantiation asVariableTemplateInstantiation() {
+    this = TVariableTemplateInstantiation(result)
+  }
+
+  string toString() {
+    result = this.asClassTemplateInstantiation().toString() or
```
MichaelRFairhurst: This member predicate (and `getLocation()`) could be `result = this.asElement().toString()`/`.getLocation()`.

cpp/autosar/src/rules/M5-0-12/SignedCharAndUnsignedCharTypeShallOnlyBeUsedForTheStorageAndUseOfNumericValues.ql
```diff
@@ -16,23 +16,237 @@
...
+ * Note that this class only includes implicit conversions and does not include explicit
+ * type conversions, i.e. casts.
+ */
+class ImplicitConversionFromPlainCharType extends Conversion {
+  ImplicitConversionFromPlainCharType() {
+    this.isImplicit() and
+    this.getExpr().getUnspecifiedType() instanceof PlainCharType and
+    (
+      this.getUnspecifiedType() instanceof SignedCharType or
+      this.getUnspecifiedType() instanceof UnsignedCharType
```
MichaelRFairhurst: Nice on using `.getUnspecifiedType()`! I forgot to mention this quirk of writing cpp queries.

cpp/autosar/src/rules/M5-0-12/SignedCharAndUnsignedCharTypeShallOnlyBeUsedForTheStorageAndUseOfNumericValues.ql
```diff
@@ -16,23 +16,237 @@
...
+    )
+  }
+}
+
+newtype TImplicitConversionElement =
```
MichaelRFairhurst: This is cool, nicely done.

In theory I think what you've written here could be used even more widely too. In the future, this could be a module under `cpp/common/src/codingstandards/cpp/alertreporting` with a parameterized module e.g.

```codeql
signature class ElementSig = Element;
module TemplatableElement<ElementSig Elem> {
  newtype TTemplatableElement = TElementOutsideTemplate(Elem elem) { ... }
  or
  TElementInsideTemplate(TemplateInstantiation templateInstantiation, Elem elem) { ... };
  ...
```

cpp/autosar/src/rules/M5-0-12/SignedCharAndUnsignedCharTypeShallOnlyBeUsedForTheStorageAndUseOfNumericValues.ql
```diff
@@ -16,23 +16,237 @@
...
+  ImplicitConversionFromPlainCharType getImplicitConversion() {
+    result = this.asImplicitConversionOutsideTemplate() or
+    exists(TemplateInstantiation templateInstantiation |
+      this = TInstantiationOfImplicitConversionTemplate(templateInstantiation, result)
+    )
+  }
+
+  string toString() {
+    result = this.asImplicitConversionOutsideTemplate().toString() or
+    exists(ImplicitConversionFromPlainCharType implicitConversion |
```
MichaelRFairhurst: can just be `result = this.asInstantiationOfImplicitConversionTemplate(_).toString()`, both here and in a few other spots

cpp/autosar/src/rules/M5-0-12/SignedCharAndUnsignedCharTypeShallOnlyBeUsedForTheStorageAndUseOfNumericValues.ql
```diff
@@ -16,23 +16,237 @@
...
+  }
+
+  Location getLocation() {
+    result = this.asImplicitConversionOutsideTemplate().getLocation() or
+    exists(ImplicitConversionFromPlainCharType implicitConversion |
+      result = this.asInstantiationOfImplicitConversionTemplate(implicitConversion).getLocation()
+    )
+  }
+
+  Element asElement() {
```
MichaelRFairhurst: I think this name is a bit confusing, since I wouldn't expect it to call `getAUse` in the templated case. I might just name this `getAUse()` or maybe `getAnOccurrence()` or something like that?

A comment would suffice too!

cpp/autosar/src/rules/M5-0-12/SignedCharAndUnsignedCharTypeShallOnlyBeUsedForTheStorageAndUseOfNumericValues.ql
```diff
@@ -16,23 +16,237 @@
...
-  ) and
-  // Identify places where these explicitly signed types are being assigned to a
-  // non-numeric type.
-  aexp = v.getAnAssignedValue() and
-  aexp.getUnspecifiedType() instanceof CharType
-select aexp,
-  "Assignment of an non-integer type to variable $@ which is a variable with an explicitly signed char type",
-  v, v.getName()
+  implicitConversion = implicitConversionLocation.getImplicitConversion()
+select implicitConversionLocation.asElement(), getMessageTemplate(implicitConversionLocation),
```
MichaelRFairhurst: `select implicitConversionLocation, ...` might also work, since you define `getLocation()`

MichaelRFairhurst:
Looks great!

If you're so inclined, it could be a good idea to add tests for const types and typedefs:

```cpp
typedef char char_t;
typedef unsigned char uchar_t;

const char x1 = 'a'; // compliant
const signed char x2 = x1; // non-compliant
char_t x3 = x1; // compliant
uchar_t x4 = x3; // non-compliant
// etc
```

It looks like the code already handles these cases correctly, since it uses `getUnspecifiedType`, so I don't think it should block the PR.

MichaelRFairhurst (approved):

PR #938, opened 2025-08-01
Implement "Statements" package

jeongsoolee09:
## Description

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation
 - [ ] Query files (`.ql`, `.qll`, `.qls` or unit tests)

 - [ ] No rules added
     - _rule number here_
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
```diff
@@ -0,0 +1,558 @@
...
+/**
+ * Holds if either of the following holds for the given variable access:
+ * 1. Another variable access of the same variable as the given variable access is taken an
+ * address and is assigned to a non-const pointer variable, i.e. initialization, assignment,
+ * and pass-by-value.
+ * 2. Another variable access of the same variable as the given variable access is assigned
+ * to a non-const reference variable (thus constituting a `T` -> `&T` conversion.), i.e.
+ * initialization and assignment.
+ */
+/*
```
MichaelRFairhurst: Simple comment formatting, unnecessary split

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
```diff
@@ -0,0 +1,558 @@
...
+ * Note that pass-by-reference is dealt with in a different predicate named
+ * `loopVariablePassedAsArgumentToNonConstReferenceParameter`, due to implementation
+ * limitations.
+ */
+
+predicate loopVariableAssignedToNonConstPointerOrReferenceType(
+  ForStmt forLoop, VariableAccess loopVariableAccessInCondition
+) {
+  exists(Expr assignmentRhs, DerivedType targetType |
```
MichaelRFairhurst: Likely want to test that this works for a `int * const x`:

```cpp
void f(int * const x) {
    (*x)++;
}

int main() {
    for (int i = 0; i < 10; ++i) {
        f(&i);
        std::cout << i << std::endl;
    }
}
```

I believe what will happen is that `int * const x` will be a `DerivedType` of type `SpecifiedType` with a const specifier. A `SpecifiedType` is not `instanceof PointerType` or `instanceof ReferenceType` and so this predicate will not hold, even though the value of `i` is modifiable within `f`.

You may also have problems with typedefs, such as `typedef int *int_ptr_t` for the same reason.

The solution here I believe will be to call `.getUnderlyingType()`. Another option frequently used for this is `.stripSpecifiers()`. Each of these will remove the const and resolve the typedef. I think `.stripSpecifiers()` may remove the const in `const int*`, though, which would make it unsuitable here.

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
```diff
@@ -0,0 +1,558 @@
...
+      targetType instanceof ReferenceType
+    )
+  |
+    assignmentRhs.getEnclosingStmt().getParent*() = forLoop.getStmt() and
+    (
+      /* 1. The address is taken: A loop variable access */
+      assignmentRhs.(AddressOfExpr).getOperand() =
+        loopVariableAccessInCondition.getTarget().getAnAccess()
+      or
+      /* 2. The address is taken: A loop variable access */
```
MichaelRFairhurst: Should this say a reference is taken?

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
```diff
@@ -0,0 +1,558 @@
...
+ * This predicate adds two constraints to the target type, as compared to the original
+ * portion of the predicate:
+ *
+ * 1. This predicate adds type constraint that the target type is a `ReferenceType`.
+ * 2. This predicate adds the constraint that the target type is not `const`.
+ *
+ * Also, this predicate requires that the call is the body of the given for-loop.
+ */
+
+/**
```
MichaelRFairhurst: I would attach these two doc block comments together so that we don't lose the connection if adding code or reordering predicates etc.

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
```diff
@@ -0,0 +1,558 @@
...
+   * since the update expression may be a compound one that embeds the four aforementioned
+   * expression types, such as a comma expression (e.g. `i += 1, E` where `E` is an
+   * arbitrary expression).
+   *
+   * This may be detrimental to performance, but we keep it for soundness. A possible
+   * alternative is an IR-based solution.
+   */
+
+  /* 1. Get the expression `E` when the update expression is `i += E` or `i -= E`. */
+  result = forLoop.getUpdate().getAChild*().(AssignAddExpr).getRValue()
```
MichaelRFairhurst: You can use `AssignAddOrSubExpr` and `.getLoopStep()`!

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
```diff
@@ -0,0 +1,558 @@
...
+import codingstandards.cpp.Call
+import codingstandards.cpp.Loops
+import codingstandards.cpp.misra.BuiltInTypeRules::MisraCpp23BuiltInTypes
+
+/**
+ * A comparison expression that has the minimum qualification as being a valid termination
+ * condition of a legacy for-loop. It is characterized by a value read from a variable being
+ * compared to a value, which is supposed to be the loop bound.
+ */
+class LegacyForLoopCondition extends RelationalOperation {
```
MichaelRFairhurst: Consider moving this into `codingstandards.cpp.Loops`

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
```diff
@@ -0,0 +1,558 @@
...
+
+abstract class LegacyForLoopUpdateExpression extends Expr {
+  ForStmt forLoop;
+
+  LegacyForLoopUpdateExpression() { this = forLoop.getUpdate().getAChild*() }
+
+  abstract Expr getLoopStep();
+}
+
+class CrementLegacyForLoopUpdateExpression extends LegacyForLoopUpdateExpression {
```
MichaelRFairhurst: Consider taking this class, and `AssignAddOrSubExpr` and `AddOrSubThenAssignExpr` and moving them to a file such as `cpp/common/src/codingstandards/cpp/ast/Increment.qll`.

I'd decouple it from loops, for instance, rename `getLoopStep` to `getIncremented` or something like that.

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
```diff
@@ -0,0 +1,558 @@
...
+) {
+  exists(Expr assignmentRhs, DerivedType targetType |
+    isAssignment(assignmentRhs, targetType, _) and
+    not targetType.getBaseType().isConst() and
+    (
+      targetType instanceof PointerType or
+      targetType instanceof ReferenceType
+    )
+  |
+    assignmentRhs.getEnclosingStmt().getParent*() = forLoop.getStmt() and
```
MichaelRFairhurst, line 97: You may want to add `not assignmentRhs.isInUnevaluatedContext()` for safety.

That would prevent reporting cases like `sizeof(g(&i))` or `decltype(g(&i))`.

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
```diff
@@ -0,0 +1,558 @@
...
+    none() // TODO
+  } or
+  /* 4. The type size of the loop counter is smaller than that of the loop bound. */
+  TLoopCounterSmallerThanLoopBound(ForStmt forLoop, LegacyForLoopCondition forLoopCondition) {
+    forLoopCondition = forLoop.getCondition() and
+    exists(Type loopCounterType, Type loopBoundType |
+      loopCounterType = forLoopCondition.getLoopCounter().getType() and
+      loopBoundType = forLoopCondition.getLoopBound().getType()
+    |
+      loopCounterType.getSize() < loopBoundType.getSize()
```
MichaelRFairhurst: Two missed cases here:
- Mixing signed/unsigned types, they may have the same size but they'll hold different ranges.
- The type and runtime value may lead to different conclusions.

I think you may be able to get away with `upperBound(loopCounter) < upperBound(loopBound)`. That would handle signedness, constants (like `x < 10ull`), and dynamic ranges (like `unsigned long long bound = 10; ... x < bound`).

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
```diff
@@ -0,0 +1,558 @@
...
+ *
+ * Also, this predicate requires that the call is the body of the given for-loop.
+ */
+
+/**
+ * Holds if the given variable access has another variable access with the same target
+ * variable that is passed as reference to a non-const reference parameter of a function,
+ * constituting a `T` -> `&T` conversion.
+ */
+predicate loopVariablePassedAsArgumentToNonConstReferenceParameter(
```
MichaelRFairhurst, line 126: A thought on simplifying these names.

Maybe instead of naming them `loopVariablePassedAs...` you can rename them to `passedAsNonConstReference`/`passedAsNonConstPointer` and remove the `ForStmt` argument.

 Then at the call sites you can change

```
loopVariablePassedAsArgumentToNonConstReferenceParameter(loop, va)
// becomes
exists(VariableAccess other |
  passedAsNonConstReference(other) and
  other.getVariable() = loop.getBound().(VariableAccess).getVariable() and
  other.getEnclosingStmt().getParent*() = loop.getStmt()
)
```

MichaelRFairhurst (commented):
This is really coming along and looking really good!!

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
jeongsoolee09: Ouch!

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
jeongsoolee09: Good call. The intention was to split the documentation and the meta-level comment (explaining how this predicate came to be). But like you said it can be disconnected easily, so I'll merge the meta-level comment into the docstring first.

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
jeongsoolee09: Good call. The intention was to split the documentation and the meta-level comment (explaining how this predicate came to be). But like you said it can be disconnected easily, so I'll merge the meta-level comment into the docstring first.

jeongsoolee09: Addressed in [c8c0770](https://github.com/github/codeql-coding-standards/pull/938/commits/c8c0770d19aafe77ec70ff1c246cecd7d9766510).

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
jeongsoolee09: Addressed in [c8c0770](https://github.com/github/codeql-coding-standards/pull/938/commits/c8c0770d19aafe77ec70ff1c246cecd7d9766510).

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
jeongsoolee09: Addressed in [c8c0770](https://github.com/github/codeql-coding-standards/pull/938/commits/c8c0770d19aafe77ec70ff1c246cecd7d9766510).

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
jeongsoolee09: Addressed in [c8c0770](https://github.com/github/codeql-coding-standards/pull/938/commits/c8c0770d19aafe77ec70ff1c246cecd7d9766510).

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
jeongsoolee09: Addressed in [2227255](https://github.com/github/codeql-coding-standards/pull/938/commits/2227255825358b3cec67478284b682037361ad54).

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
jeongsoolee09: Moved to a new file with the suggested path and name in [2227255](https://github.com/github/codeql-coding-standards/pull/938/commits/2227255825358b3cec67478284b682037361ad54).

jeongsoolee09: The library as a whole is no longer tied to `ForStmt` and is used to rewrite `getLoopStepOfForStmt` in the query, in [38b5fbc](https://github.com/github/codeql-coding-standards/pull/938/commits/38b5fbcaaa61b8860de5cc3619c89ff43294bf0e).

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
jeongsoolee09: I'm not sure this is a good idea because
- `TLoopCounterIsTakenNonConstAddress` and friends have to have an identical body, and repeating these lines three times will increase the verbosity.
- The predicates `loopVariablePassedAs...` and the other one are highly specialized to this query alone, so it doesn't hurt much to keep them amalgamated (if that's a word!) and not break them down.

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
jeongsoolee09: You're right; the predicate does not catch this example. 🤔 I guess a clever use of one or more of `isDeeplyConst`, or `isDeeplyConstBelow` will do the trick.

MichaelRFairhurst: Forgetting to handle typedefs or meaningless consts is a very common bug. But you'll (mostly) get in the habit soon enough of always calling one of these four member predicates on the `Type`s you handle in your queries:

- `getUnderlyingType()`
- `resolveTypedefs()`
- `stripSpecifiers()`
- `stripTopLevelSpecifiers()`

Each one does subtly different things.

In this case, I believe the fix is to do:

```ql
  exists(..., Type targetType, DerivedType strippedType |
    isAssignment(assignmentRhs, targetType, _) and
    strippedType = targetType.stripTopLevelSpecifiers()
    not strippedType.getBaseType().isConst() and
    (
      strippedType instanceof PointerType or
      strippedType instanceof ReferenceType
    )
```

The documentation for `stripTopLevelSpecifiers` says:

> Get this type after any top-level specifiers and typedefs have been stripped.
>
> For example, starting with `const i64* const`, this predicate will return `const i64*`.

**which is actually wrong**, as it ignores the fact that `i64` is a `TypeDefType`, so it _actually_ will result in `const long long*`. Which is what you want!

The TLDR of the other options:

- `getUnderlyingType()` -- resolves `TypdefType`s and `DeclType`s, but won't drop the outer specifer in `const i64* const`. Stops at the first non-TypedefType/non-DeclType.
- `stripType()` -- resolves all typedefs and decltypes and removes all `const`/`volatile` specifiers recursively all the way down the type chain -- not what you want.
- `resolveTypedefs` -- resolves all typedefs and decltypes all the way down the type chain without removing `const` or `volatile` specifiers. That would handle typedefs but not `int const *`.

Note that these predicates can have no result. Only a limited set of types are in the database, and these operations just assume that the type you want is one of those types. `resolveTypedefs` is also bugged and doesn't recurse into `ArrayType`.

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
MichaelRFairhurst: Feel free to resolve this.

I mostly wanted to point this out because this is a common trap case in coding standards query writing. In C/C++ you can write any expression inside of a `sizeof()` check, as well as other exprs like `decltype()`/`alignof()`/constexpr()`/`requires()`, and these aren't evaluated.

So if you're looking at a rule that says something like, "Never pass `nullptr` into `std::some_function(x)`", then that's the kind of rule where we may want to be careful that `sizeof(std::some_function(nullptr))` isn't flagged, because it won't actually execute `std::some_function()`.

In this case, I've convinced myself this isn't something we have to worry about in this case. Hopefully I'm not wrong about that! :)

<details>
<summary>A really complicated example with TLDR, no need to worry about this.</summary>

Here's the most reasonable example that Copilot and I came up with:

```
template<typename IndexType, typename Observer>
void process_with_observer(std::vector<int>& data) {
    Observer observer;
    
    for (IndexType i = 0; i < data.size(); ++i) {
        // Determine observer's interface at compile time
        using observer_result = decltype(observer(&i));
        
        if constexpr (std::is_void_v<observer_result>) {
            observer(&i);
            data[i] = default_transform(data[i]);
        } else {
            auto metadata = observer(&i);
            data[i] = complex_transform(data[i], metadata);
        }
    }
}
```

This code is _almost_ reasonable. It would allow you to customize the loop behavior by writing a class with an overloaded `()` operator......but it's really really strange, and the only "problem" we'd have analyzing this code is that it would flag all three sites to `observer(&i)` when we would "ideally?" only flag two.

So yeah. No need to worry about it :)
</detail>

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
MichaelRFairhurst: Also almost forgot

Another trap case is that when doing `upperBound(e)` / `lowerBound(e)` you usually want `upperBound(e.getFullyConverted())`. Because conversions on `e` will change the bound.

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
MichaelRFairhurst: I wouldn't worry about this as extra verbosity because the extra code is easy to read. In general, I'd worry more about clear/concise naming.

If verbosity is an issue you can add helper methods to `LegacyForLoopCondition`

```
class LegacyForLoopCondition {
  ...
  predicate isBoundAccess(VariableAccess va) {
    va.getVariable() = getBound().(VariableAccess).getVariable() and
    other.getEnclosingStmt().getParent*() = loop.getStmt()
  }
  ...
}

...
  exists(VariableAccess other
    | passedAsNonConstReference(other) and loopCondition.isBoundAccess(other))
 ```

You can definitely keep it as is if you prefer, and you may think of a better way to abstract the duplicated code than the `isBoundAccess()` example here.

Definitely your call!

File: cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql (hunk 2 above)
Comment (jeongsoolee09) [reply cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql]: Thank you for the detailed breakdown of the related predicates. What I want to express here is definitely "The type we get _after we strip all the typedefs and the specifiers_ is const". I've come to believe `stripTopLevelSpecifiers` is the one I should use, and swapped the portion with your suggestion.

I also patched an equivalent part in `loopVariablePassedAsArgumentToNonConstReferenceParameter`, in [7d5f08b](https://github.com/github/codeql-coding-standards/pull/938/commits/7d5f08b3e788afbe46fe865f4243205263eb3cd0).

File: cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql (hunk 1 above)
Comment (jeongsoolee09) [reply cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql]: Somehow this change didn't make it to c8c0770; it did to a recent commit.

File: cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql (hunk 9 above)
Comment (jeongsoolee09) [reply cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql]: This eliminated _a lot_ of false positives where the counter variable is `int` and the loop bound is `size_t`. Thank you!

File: cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql (hunk 8 above)
Comment (jeongsoolee09) [reply cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql:97]: Since it's nonetheless a valid use case, I might record this in an issue. Thank you for drawing my attention to these special operators!

File: cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql (hunk 10 above)
Comment (jeongsoolee09) [reply cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql:126]: I can't think of a clean way to cut it up until I run into a potential case that would benefit from reusing parts of it. Until then I'd keep it and the other one as they are.

File: cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql (hunk 9 above)
Comment (jeongsoolee09) [reply cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql]: Changed the `upperBound(loopCounter.getFullyConverted())` to `typeUpperBound(loopCounter.getType())`. As `typeUpperBound` resolves references, I didn't have to use `getBaseType()` on it.

File: cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql [hunk 11]
```diff
@@ -0,0 +1,478 @@
...
+   */
+
+  TNoRelationalOperatorInLoopCondition(ForStmt forLoop, Expr condition) {
+    condition = forLoop.getCondition() and
+    not condition instanceof LegacyForLoopCondition
+  } or
+  /* 3-1. The loop counter is mutated somewhere other than its update expression. */
+  TLoopCounterMutatedInLoopBody(ForStmt forLoop, Variable loopCounterVariable) {
+    loopCounterVariable = getDeclaredVariableInForLoop(forLoop) and
+    variableModifiedInExpression(forLoop.getStmt().getChildStmt().getAChild*(),
```
MichaelRFairhurst: Do we need to worry about updates in the condition?

```
for(int i = 0; i++ < 10;) {...}
```

cpp/misra/test/rules/RULE-9-5-1/test.cpp
```diff
@@ -0,0 +1,293 @@
...
+
+  for (int i = 0; i < 10;
+       i += j, j++) { // NON_COMPLIANT: The loop step `j` is mutated in the loop
+  }
+
+  for (int i = 0; i < j; i++) { // COMPLIANT: The loop bound `j` is not mutated
+                                // anywhere in the loop
+  }
+
+  for (int i = 0; i < j; i++) { // COMPLIANT: The loop bound `j` is not mutated
```
MichaelRFairhurst, line 166: Case to test: `for (int i = 0; i < j++; i++)`, and maybe `for (int i = 0; i++ < j++; i++)` while we're at it.

cpp/common/src/codingstandards/cpp/Loops.qll
```diff
@@ -374,3 +374,43 @@ predicate isInvalidLoop(ForStmt forLoop, string reason, Locatable reasonLocation
...
+  ForStmt forLoop;
+  VariableAccess loopCounter;
+  Expr loopBound;
+
+  LegacyForLoopCondition() {
+    this = forLoop.getCondition() and
+    exists(Expr loopCounterExpr |
+      loopCounterExpr = this.getAnOperand() and
+      loopBound = this.getAnOperand() and
+      loopCounter = loopCounterExpr.getAChild*() and
```
MichaelRFairhurst: `getAChild*()` is the right tool here, but it must be coupled with an allow-list or we'll have FNs, because it's still casting a very very wide net.

We should ensure (either here, or reported as an error in the query) that `loopCounterExpr` is not any arbitrary type of expression.

The following should be non-compliant:

```
for (int i = 0; f(i) < 10; ++i) {}
for (int i = 0; i * i < 10; ++i) {}
for (int i = 0; i + f() < 10; ++i) {}
for (int i = 0; (i > other_var) < 1; ++i) {}
// etc
```

Basically, we probably just want an allow-list where every expr from loopCounterExpr to loopCounter is either loopCounter itself or an addition/subtraction with only constant values on one side and an allow-listed expression on the other.

```
for (int i = 0; i + 10 < 20; ++i) {} // OK, `i` is allowed and `ALLOWED + 10` is allowed
for (int i = 0; 10 - i < 20; ++i) {} // OK, `i` is allowed and  `10 - ALLOWED` is allowed
for (int i = 0; -i < 20; ++i) {} // OK, `i` is allowed and `-ALLOWED` is allowed
for (int i = 0; -i + 10 - < 20; ++i) {} // OK, `i` is allowed and  -ALLOWED is allowed
for (int i = 0; (i + 5) + 3 < 20; ++i) {} // OK, `i` is allowed, `ALLOWED + 5`, and `ALLOWED + 3` is allowed
for (int i = 0; i + (5 + 3) < 20; ++i) {} // OK, `i` is allowed, `ALLOWED + (5 + 3)` is allowed

for (int i = 0; i + i < 20; ++i) {} // BAD, `i` is allowed but `ALLOWED + ALLOWED` is not allowed
for (int i = 0; i + j < 20; ++j) {} // BAD, 'j' is not allowed
for (int i = 0; (i + 10) + (i + 10) < 20; ++i) {} // BAD, `i` and `ALLOWED + 10` is allowed, but `ALLOWED + ALLOWED` is not allowed
```

Hopefully that mostly makes sense.

cpp/misra/test/rules/RULE-9-5-1/test.cpp
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  for (int i = j; i < k; i += l) { // COMPLIANT: The loop step is taken as
+                                   // a const reference
+    const int &m = l;
+  }
+
+  for (int i = j; i < k; i += l) { // COMPLIANT: The loop step is taken as
+                                   // a const pointer
+    const int *m = &l;
```
MichaelRFairhurst, line 285: Add tests that `int * const m` is non compliant

cpp/misra/test/rules/RULE-9-5-1/test.cpp
jeongsoolee09: Sounds good. Two alerts should appear in the second example:
- The loop bound is mutated (the rule in question here).
- The loop counter is mutated outside of the update expression.

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
jeongsoolee09: I changed the case body to better match the comment `/* 3-1. ... */`, it now finds all cases where the mutating expression is *not in the update expression* ([25e29e4](https://github.com/github/codeql-coding-standards/pull/938/commits/25e29e48e4daf0bd97c3059dff85861784b6ef3c)).

cpp/misra/test/rules/RULE-9-5-1/test.cpp
jeongsoolee09: Added in [0f998ea](https://github.com/github/codeql-coding-standards/pull/938/commits/0f998ea7b1ec8087c12d001693c0d4cd34288f8c) and covered in [01216fa](https://github.com/github/codeql-coding-standards/pull/938/commits/01216fa1cdd0b4e8a01bedd09b8ab52e53453c61).

cpp/misra/test/rules/RULE-9-5-1/test.cpp
jeongsoolee09: Added in [2aed0f1](https://github.com/github/codeql-coding-standards/pull/938/commits/2aed0f125fcc52d81a55285a8271c69fd82eaf74).

jeongsoolee09: Covered in [33222ae](https://github.com/github/codeql-coding-standards/pull/938/commits/33222aef74c5fe67d70c7b685f3033cff04f8a21).

cpp/common/src/codingstandards/cpp/Loops.qll
jeongsoolee09: I understand what you mean, I guess what you suggest is to define a allowed class of ASTs where:
- The base case is a VariableAccess or a unary operator application on it; and
- The recursive case is an application of addition or subtraction to the allowed class.

However, I think even these are non-compliant since the rules dictate that (paraphrased by me): 

> The comparison between the loop-counter and the loop-step is done in the condition by only using a relational operator

which is preceded by (also paraphrased and highlighted by me):

> The only thing the init-statement does is declaring and initializing a *loop-counter of integer type*

From these two, we can infer that the rule implicitly wants (1) the loop counter to be a *variable*, and (2) the loop condition to be a *direct comparison* between the loop condition variable and some loop-bound expression. This reflected in the alert message we emit in case of a violation of this sub-rule: `"The [loop condition] does not determine termination based only on a comparison against the value of the counter variable"`.

Then, one might ask, why do we distinguish `loopCounterExpr` from `loopCounter` in the characteristic predicate? That's for internal purposes only; we somehow need to distinguish which of the operands has `loopCounter` in it, so we dig into one of the expressions to find a variable whose target is the initialized counter variable. I'd say it's unfortunate but necessary.

Then, a follow-up objection might be: Why don't we demand that one of the operands to the relational operator is a variable access in the first place? That's because we don't want to tailor this code specifically to this rule, because it's library code. `9-5-1` wants one of the things compared is an access to the initialized variable, but other rules might not, and in those cases this class (`LegacyForLoopCondition`) can be used in other ways.

MichaelRFairhurst: OK, I thought that this code was intended to handle the following case, based on real code examples in open pilot:

```
for (int i = 0; i + 10 < j; ++i) {}
```

I believe we have leeway to be able to interpret 9-5-1 to accept or reject this. It's trivially clear that `i + 10 < j` is equivalent to `i < j + 10`. It's fair to say that the MISRA team doesn't think of every edge case and that we have some discretion here. It's also fair to say that this violates the plain text of the rule.

I'm happy with us rejecting the above case if you think so. However, I don't think the query _currently_ rejects it because of the `.getAChild*()`. Quickly looking for calls to `getLoopCounter`, I don't see anything that later checks that `getLoopCounter()` is a direct operand of the `LegacyForLoopCondition`.

I believe that 9-5-1 will also currently not flag

```
for (int i = 0; f(i) < j; ++i) {}
```

which is clearly intended to be non-compliant.

So we should add tests for these cases. We should test `f(i) < j` case is flagged as non compliant, and we should add a test for `i + 10 < j` with a comment about how we came to decide that it should or shouldn't be considered compliant.

My preference would be to remove `getAChild*()`, and either replace it with `getAnOperand()` or a predicate that finds a restricted set of operations. Currently `LegacyForLoopCondition` is only used here, and we can document its restrictions. If you would rather leave it as is and add checks in 9-5-1 that `getLoopCounter() = loopCondition.getAnOperand()`, that is reasonable too. But `getAChild*()` casts a very wide net that might not be expected in future queries, so I wouldn't say we need to have that behavior now.

jeongsoolee09: Okay, I believe it is reasonable to do so. Thank you for your second opinion!

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
```diff
@@ -0,0 +1,505 @@
...
+      strippedReferenceType = targetType.stripTopLevelSpecifiers() and
+      not strippedReferenceType.getBaseType().isConst()
+    |
+      /* A regular function call */
+      targetType = call.getTarget().getParameter(i).getType()
+      or
+      /* A function call where the argument is passed as varargs */
+      call.getTarget().getNumberOfParameters() <= i and
+      /* The rule states that the type should match the "adjusted" type of the argument */
+      targetType = loopVariableAccessInCondition.getFullyConverted().getType()
```
MichaelRFairhurst, line 143: I believe this is incorrect. Take the following example:

```
01  void f(int &args...);
02  void g(int args...);
03  void main() {
04    for(int i = 0; i < 10; ++i) {
05     f(i);
06     g(i);
07   }
08  }
```

In this example, we want to report `f(i)` as non-compliant while `g(i)` is OK. This code is looking to see if `i` is converted to a const or a non-const reference type. However, it's looking at the wrong `i`.

`loopVariableAccessInCondition` refers to the i on line 3. But what we want to see is if the `i`s on lines 4 and 5 are converted to `const int&` or `int&`.

We should add tests for this case as well.

cpp/common/src/codingstandards/cpp/Loops.qll
jeongsoolee09: Added in [15be010](https://github.com/github/codeql-coding-standards/pull/938/commits/15be0109ebb4fa1721e7ba362359f9f56b8a306b).

cpp/misra/src/rules/RULE-9-5-1/LegacyForStatementsShouldBeSimple.ql
jeongsoolee09: Aha, I see where the confusion is coming from. The link targets were not meant to be the mutating expressions; rather they point to the loop counter variables themselves that are mutated somewhere in the loop. I solved this issue by exposing the mutating expressions to the query message in [6ffd229](https://github.com/github/codeql-coding-standards/pull/938/commits/6ffd2296d662c4e27d885589e21d51892cec55b5).

jeongsoolee09: Also added the test cases in [15be010](https://github.com/github/codeql-coding-standards/pull/938/commits/15be0109ebb4fa1721e7ba362359f9f56b8a306b).

PR #967, opened 2025-11-18
Implement `Memory1` (`RULE-8-7-1`)

jeongsoolee09:
## Description

Implement Memory1 (`RULE-8-7-1`).

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - `RULE-8-7-1`
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


cpp/common/src/codingstandards/cpp/OutOfBounds.qll
```diff
@@ -0,0 +1,1358 @@
+/**
+ * This module provides classes and predicates for analyzing the size of buffers
+ * or objects from their base or a byte-offset, and identifying the potential for
+ * expressions accessing those buffers to overflow.
```
MichaelRFairhurst, line 10: Is this a direct copy? We should probably state that, and/or, list modifications that have been made.

cpp/misra/src/rules/RULE-8-7-1/PointerArgumentToCstringFunctionIsInvalid.ql
```diff
@@ -0,0 +1,26 @@
...
+ * @tags external/misra/id/rule-8-7-1
+ *       scope/system
+ *       external/misra/enforcement/undecidable
+ *       external/misra/obligation/required
+ */
+
+import cpp
+import codingstandards.cpp.OutOfBounds // for OOB::problems
+import codingstandards.cpp.Exclusions // for isExcluded(Element, Query)
+import codingstandards.cpp.exclusions.c.RuleMetadata
```
MichaelRFairhurst: I think this should be `codingstandards.cpp.exclusions.cpp.RuleMetadata`, or deleted?

cpp/misra/src/rules/RULE-8-7-1/PointerArgumentToCstringFunctionIsInvalid.ql
```diff
@@ -0,0 +1,26 @@
...
+
+from
+  OOB::BufferAccessLibraryFunctionCall fc, string message, Expr bufferArg, string bufferArgStr,
+  Expr sizeOrOtherBufferArg, string otherStr
+where
+  not isExcluded(fc, OutOfBoundsPackage::libraryFunctionArgumentOutOfBoundsQuery()) and
```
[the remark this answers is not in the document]
MichaelRFairhurst: I believe this should be `not isExcluded(Memory1Package::theQueryName())`

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
```diff
@@ -0,0 +1,301 @@
...
+  /**
+   * Gets the declared length of this array.
+   */
+  int getLength() { result = length }
+}
+
+/**
+ * A call to a function that dynamically allocates memory on the heap.
+ */
+class HeapAllocationFunctionCall extends FunctionCall {
```
MichaelRFairhurst, line 48: Maybe we should put this logic in `common` somewhere, maybe `standardlibrary/memory`?

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
```diff
@@ -0,0 +1,301 @@
...
+  MallocFunctionCall() { this.isMallocCall() }
+
+  override int getMinNumBytes() { result = lowerBound(this.getArgument(0)) }
+}
+
+class CallocFunctionCall extends HeapAllocationFunctionCall {
+  CallocFunctionCall() { this.isCallocCall() }
+
+  override int getMinNumBytes() {
+    result = lowerBound(this.getArgument(0)) * lowerBound(this.getArgument(1))
```
MichaelRFairhurst, line 75: Do we want to use the minimum, or the maximum?

I'd suggest we run this on MRVA and see how many false positives we get. If it is a lot, I'd suggest using `upperBound()`.

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
```diff
@@ -0,0 +1,301 @@
...
+  string toString() {
+    result = this.asArrayExpr().toString() or
+    result = this.asPointerArithmetic().toString()
+  }
+
+  /**
+   * Gets the offset of this pointer formation as calculated in relation to the base pointer.
+   */
+  int getOffset() {
+    result = this.asArrayExpr().getArrayOffset().getValue().toInt()
```
MichaelRFairhurst: instead of `getArrayOffset().getValue().toInt()`, which only handles constants, we likely want to use range analysis, either `upperBound` (noisiest) or `lowerBound` (quietest).

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
```diff
@@ -0,0 +1,301 @@
...
+    or
+    exists(PointerAddExpr pointerAddition | pointerAddition = this.asPointerArithmetic() |
+      result = pointerAddition.getAnOperand().getValue().toInt() // TODO: only get the number being added
+    )
+    or
+    exists(PointerSubExpr pointerSubtraction | pointerSubtraction = this.asPointerArithmetic() |
+      result = -pointerSubtraction.getAnOperand().getValue().toInt()
```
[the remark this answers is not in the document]
MichaelRFairhurst: Let's handle this todo, but not how copilot is suggesting, so that we can handle both `p + n` and `n + p`.

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
```diff
@@ -0,0 +1,301 @@
...
+ * NOTE The code in the below module is copied from
+ * `cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/TaintTrackingUtil.qll` in `github/codeql`, commit hash
+ * `960e990`. This commit hash is the latest of the ones with tag  `codeql-cli-2.21.4` which is the CLI version
+ * compatible with `codeql/cpp-all: 5.0.0` that this query depends on.
+ */
+module Copied {
+  import semmle.code.cpp.ir.IR
+  import semmle.code.cpp.ir.dataflow.internal.SsaInternals as Ssa
+
+  predicate operandToInstructionTaintStep(Operand opFrom, Instruction instrTo) {
```
MichaelRFairhurst: Unfortunately, as I run this on some c++ databases, it seems that this is adding too many edges, and we may need to try a different approach :(

For instance, I'm getting an error in the following matrix math code on opencv

```
double A[2][9];
A[k][4] = ...;
```

The alert is "this pointer has offset 4 when the minimum possible length of the object might be 2."

I think the issue may be related to this:

```
double _Qy[2][4] /* Node 1 */
  = { /* Node 2 */
       {c, 0, -s}, /* Node 3 */
       {0, 1, 0}, /* Node 4 */
       {s, 0, c} /* Node 5 */
  };
```

This taint predicate causes Node 1 to taint Node 2, and Node 2 taints nodes 3, 4, and 5. This is a pretty big problem, it means that any path to `_Qy[x]` will also have a path to `_Qy[0][x]` -- basically, we won't be able to distinguish multidimensional arrays at all.

There are many other edges being added that also seem concerning. In the results for opencv, many of the paths are nearly 100 nodes long. Another example is the following path:

```
// 1.
uint8_t m[1][6] = { ... }; // &m has size 1
// 2.
cv::Mat images_u = Mat(1, 6, CV_8UC1, m); // &images_u.data has size 1
// 3.
const Mat* images = &images_u; // images->data has size 1

// Note, the following code executes between 3. and 4.
std::vector<uchar*>& _ptrs; // declare vector _ptrs
// some amount of elements get put in _ptrs

// 4.
_ptrs[i] = images[j].data + c*esz1; // _ptrs[i] has unknown size
// A few things happen here:
// - An alert should probably occur, because images[j].data + c*esz1 might produce an invalid pointer
// - The size of _ptrs[i] is now unknown.
// - The size of _ptrs is not known, and not based on step 3

// 5.
T** ptrs = (T**)&_ptrs[0];
// Two things happen in the above line
// - _ptrs[0] is a pointer of unknown size that flows from step 4
// - &_ptrs[0] takes the address of that pointer, producing a new pointer to a buffer of unknown size

// 6.
ptrs[2];
// Here we get an alert: minimum possible length of object might be 1. This alert flows from node 1
// However, the size of `ptrs` is not based on node 1.
// The size of `ptrs` is based on how many elements get put in `_ptrs` between steps 3 and 4
```

The above logic looks exactly like what we would expect from taint analysis, but isn't what we want for precise buffer size tracing. This makes me think that the edges being added via taint analysis are not the right edges.

Let me know if you want to pair program on this. I know these indirect data flow nodes have been nasty, so if you want to let me bash my head in for a bit on it we can see what are other options may be together.

MichaelRFairhurst (changes requested):
Very nice work, Jeongsoo! The work you've put into this really shows, its elegant and focused. In terms of the taint tracking edge cases, I think we can handle those cleanly on your foundation here if we focus on the array-to-pointer conversion cases, which we can talk more about later! Also, I have to say the tests you've made are awesomely comprehensive, nicely done. That is huge and really shows all your attention to detail!

cpp/misra/src/rules/RULE-8-7-1/PointerArgumentToCstringFunctionIsInvalid.ql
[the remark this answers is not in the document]
jeongsoolee09: Fixed in [8435f3c](https://github.com/github/codeql-coding-standards/pull/967/commits/8435f3cc1f1b65a7257e024b0d038841b42d637b)!

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
jeongsoolee09: Good thing we've come up with a better solution than this!

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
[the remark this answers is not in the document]
jeongsoolee09: I believe this is no longer an issue because of the new `srcSinkLengthMap`: The analysis is bound to track `p` if it was in the middle of tracking `p`.

cpp/common/src/codingstandards/cpp/OutOfBounds.qll
jeongsoolee09: It is indeed a copy of `c/common/src/codingstandards/c/OutOfBounds.qll` and it needed to be copy over because `cpp` does not depend on `c` but it's the other way around. The only modification made is to remove `import codingstandards.c.Variable` because of the same reason: the import would fail.

That means that line is not needed in the original file as well, but that's for another refactor.

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
```diff
@@ -0,0 +1,410 @@
...
+import semmle.code.cpp.ir.IR
+import semmle.code.cpp.dataflow.new.DataFlow
+import semmle.code.cpp.ir.dataflow.internal.DataFlowUtil
+import semmle.code.cpp.security.BufferAccess
+
+/**
+ * A declaration of a variable that is of an array type.
+ */
+class ArrayDeclaration extends VariableDeclarationEntry {
+  int length;
```
MichaelRFairhurst: We can probably delete this field

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
MichaelRFairhurst: Resolving -- let's focus on shipping it

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
MichaelRFairhurst: Perhaps we should file a bug to come back to this.

In theory, it would be great to have two versions of the query: one where we know with certainty that the resulting pointer is out of bounds if flow analysis is correct -- we assume the maximum allocation size and the smallest pointer offsets. Then another where we suspect a possible invalid pointer, where we assume the minimum allocation size and the largest pointer offsets. These could share most behavior and would have different precisions.

In the meantime, lets ship!

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
```diff
@@ -0,0 +1,410 @@
...
+       */
+
+      result = rawResult.maximum(1)
+    )
+  }
+}
+
+newtype TArrayAllocation =
+  TStackAllocation(ArrayDeclaration arrayDecl) or
+  TDynamicAllocation(NarrowedHeapAllocationFunctionCall narrowedAlloc)
```
MichaelRFairhurst: Let's file a bug to come back to the third kind of "allocation," which is just taking the address of a non-array variable or lvalue.

```
int x = 0;
int *p = &x; // p is essentially a buffer of size 1
```

Partly I say let's come back because we would need to be careful to distinguish:

```
int x = 0;
int arr[5] = {0};

int *p1 = &x; // generally, taking an address to anything should be a buffer of size 1
int *p2 = &arr[0]; // except this

// Note that any lvalue expression can create a "buffer" of size 1, not just variables:
int &f() { return x; }
int *p3 = &f(); // also a "buffer" of size 1
int *p4 = &*p3; // also a "buffer" of size 1
```

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
```diff
@@ -0,0 +1,410 @@
...
+      result = pointerArithmetic.getAnOperand() // TODO: only get the number being added
+    )
+  }
+
+  /**
+   * Gets the offset of this pointer formation as calculated in relation to the base pointer.
+   */
+  int getOffset() {
+    if this.asPointerArithmetic() instanceof PointerSubExpr
+    then result = -this.getOffsetExpr().getValue().toInt()
```
MichaelRFairhurst, line 280: Another thing to file is that this currently only works on constant values, but in the future we could extend this to use range analysis.

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
```diff
@@ -0,0 +1,410 @@
...
+  FatPointer start, FatPointer end, int srcOffset, int sinkOffset, int length
+) {
+  exists(TrackArray::PathNode src, TrackArray::PathNode sink |
+    TrackArray::flowPath(src, sink) and
+    /* Reiterate the data flow configuration here. */
+    src.getNode() = start.getNode() and
+    sink.getNode() = end.getBasePointerNode()
+  |
+    srcOffset = start.getOffset() and
+    sinkOffset = end.getOffset() and
```
MichaelRFairhurst, line 437: This overwrites the previous offset, but they should add up.

For example:

```cpp
int arr[5];
int *p = arr;
int p1 = p + 3; // offset: 3, length: 5
int p2 = p1 + 2; // offset: 5, length: 5
```

Currently, this will produce `sinkOffset = 2` for the last line

MichaelRFairhurst (changes requested):
So close to ready!

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
jeongsoolee09: It's used in `ArrayAllocation::getLength/0`, so I'll dismiss it (I know it's an old comment!).

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
jeongsoolee09: Implemented in [394b7ad](https://github.com/github/codeql-coding-standards/pull/967/commits/394b7ada0c36168d8ba8c71f0e7e0e3c24dc23c7) and documented in [2e4ace6](https://github.com/github/codeql-coding-standards/pull/967/commits/2e4ace6b834dbb0b20bfb7662f1365be0e662ec0).

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
jeongsoolee09: Implemented to some degree in [44ef266](https://github.com/github/codeql-coding-standards/pull/967/commits/44ef2668d99f96412412efc5a7c50302ffb197d9).

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
jeongsoolee09: Good point. Introducing range analysis should be careful, otherwise it might generate a lot of noise. This is out of scope of this PR and should be reserved for later.

jeongsoolee09:
Two things to note about the multidimensional arrays:
1. If a row access and the row element access are apart from each other with a function boundary in between, the query loses indirection information. We probably need to add a `level` column to simulate push-pop behavior: taking an indirection edge during initialization pushes and accesses pops them.
2. Currently, the relationship between `FatPointer` and `DataFlow::Node` is one-to-many, if we don't-care the level column. The multidimensional array accesses have duplicate alerts and this is probably what is causing it.

cpp/common/src/codingstandards/cpp/OutOfBounds.qll
MichaelRFairhurst: In this case, can we have `c/common/src/codingatandards/c/OutOfBounds.qll` import `cpp.OutOfBounds` ? A simple wrapper import file would be a reasonable easy refactor.

(or find each c query that imports OutOfBounds and update those)?

I think it's probably reasonable

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
```diff
@@ -0,0 +1,451 @@
+/**
+ * @id cpp/misra/pointer-arithmetic-forms-an-invalid-pointer
+ * @name RULE-8-7-1: Pointer arithmetic shall not form an invalid pointer
+ * @description Pointers obtained as result of performing arithmetic should point to an initialized
+ *              object, or an element right next to the last element of an array.
+ * @kind path-problem
+ * @precision medium
+ * @problem.severity error
+ * @tags external/misra/id/rule-8-7-1
+ *       scope/system
```
MichaelRFairhurst, line 10: looks like we're missing correctness

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
```diff
@@ -0,0 +1,451 @@
...
+ */
+class ArrayDeclaration extends VariableDeclarationEntry {
+  int length;
+
+  ArrayDeclaration() { this.getType().getUnderlyingType().(ArrayType).getArraySize() = length }
+
+  /**
+   * Gets the declared length of this array.
+   */
+  int getLength() { result = length }
```
MichaelRFairhurst: This and the `int length;` on line 26 can be deleted now, right?

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
MichaelRFairhurst: Nice!

Only thought now, do still need `srcOffset`, `sinkOffset` to be in the table / to be predicate parameters?

In the base case, the `srcOffset` and `sinkOffset` come from `start` and `end`, not from `srcSinkLengthMap`.

In the recursive case, the `srcOffset` from the previous iteration is unused `(srcSinkLengthMap(_, start, /*here -> */ _, ...)`. The `sinkOffset` from the previous iteration is only bound to be the new `srcOffset`, which we just determined wasn't used in the next iteration.

`srcOffset` and `sinkOffset` are then also not used by the select

cpp/misra/test/rules/RULE-8-7-1/test.cpp
```diff
@@ -0,0 +1,555 @@
...
+  int *valid1 = array;     // COMPLIANT: pointer is within boundary
+  int *valid2 = array + 1; // COMPLIANT: pointer is within boundary
+  int *valid3 = array + 2; // COMPLIANT: pointer is within boundary
+  int *valid4 =
+      array + 3; // COMPLIANT: pointer points one beyond the last element
+  int *invalid1 =
+      array +
+      4; // NON_COMPLIANT: pointer points more than one beyond the last element
+  int *invalid2 =
+      array - 1; // NON_COMPLIANT: pointer is outside boundary [FALSE_NEGATIVE]
```
MichaelRFairhurst: No longer a false negative! 🎉

cpp/misra/test/rules/RULE-8-7-1/test.cpp
```diff
@@ -0,0 +1,555 @@
...
+  /* 2. Array Access (entails pointer arithmetic) */
+  int valid1 = array[0];   // COMPLIANT: pointer is within boundary
+  int valid2 = array[1];   // COMPLIANT: pointer is within boundary
+  int valid3 = array[2];   // COMPLIANT: pointer is within boundary
+  int valid4 = array[3];   // COMPLIANT: pointer points one beyond the last
+                           // element, but non-compliant to Rule 4.1.3
+  int invalid1 = array[4]; // NON_COMPLIANT: pointer points more than one beyond
+                           // the last element
+  int invalid2 =
+      array[-1]; // NON_COMPLIANT: pointer is outside boundary [FALSE_NEGATIVE]
```
MichaelRFairhurst: (same)

cpp/misra/test/rules/RULE-8-7-1/test.cpp
```diff
@@ -0,0 +1,555 @@
...
+    strcat(buf0, " "); // NON_COMPLIANT[FALSE_NEGATIVE] - not null terminated at
+                       // initialization
+
+    memset(buf0, 0, sizeof(buf0)); // COMPLIANT
+    memset(buf2, 0, sizeof(buf2)); // COMPLIANT
+
+    strcat(buf1, " ");     // NON_COMPLIANT - not null terminated
+    strcat(buf2, " ");     // COMPLIANT
+    strcat(buf3, " ");     // COMPLIANT
+    strcat(buf4, "12345"); // NON_COMPLIANT
```
MichaelRFairhurst: Hmm, this is actually a FN

rule_packages/cpp/Memory1.json
```diff
@@ -0,0 +1,35 @@
...
+      "properties": {
+        "enforcement": "undecidable",
+        "obligation": "required"
+      },
+      "queries": [
+        {
+          "description": "Pointers obtained as result of performing arithmetic should point to an initialized object, or an element right next to the last element of an array.",
+          "kind": "path-problem",
+          "name": "Pointer arithmetic shall not form an invalid pointer",
+          "precision": "medium",
```
MichaelRFairhurst, line 13: I'd be very tempted to say "high" precision.

Nicely done :)

rule_packages/cpp/Memory1.json
```diff
@@ -0,0 +1,35 @@
...
+          "severity": "error",
+          "short_name": "PointerArithmeticFormsAnInvalidPointer",
+          "tags": [
+            "scope/system"
```
MichaelRFairhurst: tags should probably include correctness and security, for this and below

rule_packages/cpp/Memory1.json
```diff
@@ -0,0 +1,35 @@
...
+          ]
+        },
```
MichaelRFairhurst, line 24: Consider adding an implementation scope that we only handle constant offsets for increased precision

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
```diff
@@ -0,0 +1,451 @@
...
+  |
+    srcSinkLengthMap(start, end, srcOffset, sinkOffset, totalOffset, length) and
+    (
+      totalOffset < 0 or // Underflow detection
+      totalOffset > length // Overflow detection
+    )
+  )
+select end, src, sink,
+  "This pointer accesses element at index " + totalOffset +
+    " while the underlying object has length " + length + "."
```
MichaelRFairhurst, line 488: One small thought on verbiage.

While "Object" is the right term here, generally speaking, I'm not sure devs think of objects as something that has a "length"... I'd say objects have a "size," and that size is in bytes, making "length" perhaps doubly confusing.

If you want to go with "length," I'd probably suggest "array" or "buffer." Though it could be confusing to say that `&x` is either.

Maybe something like, "Pointer formed that points to element X of an object contains Y elements" ?

MichaelRFairhurst (changes requested):
literally just minor tweaks!! This looks 🔥 great!

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
jeongsoolee09: It's pretty tricky because "length" denotes the count of the elements while "size" is measured in bytes. The more relevant notion to this rule is length that goes together with "indices".

But again, "elements" belongs to containers and not simply any objects. (the thing that contains Y elements is a _container_!)

All in all, I don't think there's a sensible alternative that can satisfy everything. The rule itself is conducive to conflating object with an array.

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
jeongsoolee09: No, but we can have `getLength` compute the `length` from the charpred, thereby removing the class field `int length` and `getLength/0`. That is done in `e8cb6b5`.

cpp/misra/test/rules/RULE-8-7-1/test.cpp
jeongsoolee09: Removed the label in `ebd1e58`!

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
jeongsoolee09: Yes, that's obviously cleaner. 🙂 But I'd like to leave it this way to inspect all "local variables" from simply quick-evaluating the predicate.

cpp/misra/test/rules/RULE-8-7-1/test.cpp
jeongsoolee09: Flagged as an FN at [a21c8ba](https://github.com/github/codeql-coding-standards/pull/967/commits/a21c8ba120d7a24e5fb4c7b7039decd04570ebeb). As I wrote in the commit message, the C queries that depend on `OutOfBounds.qll` consistently miss this case:

- c/misra/src/rules/RULE-21-17/StringFunctionPointerArgumentOutOfBounds.ql
- c/misra/src/rules/RULE-21-18/StringLibrarySizeArgumentOutOfBounds.ql
- c/cert/src/rules/ARR30-C/DoNotFormOutOfBoundsPointersOrArraySubscripts.ql
- c/cert/src/rules/ARR38-C/LibraryFunctionArgumentOutOfBounds.ql

Something to get back to later.

cpp/misra/test/rules/RULE-8-7-1/test.cpp
jeongsoolee09: Removed the label in `ebd1e58` 🙂

rule_packages/cpp/Memory1.json
jeongsoolee09: Added in [60ddfaa](https://github.com/github/codeql-coding-standards/pull/967/commits/60ddfaafc88a440410494a5b4fed2ac68ba24f6f).

rule_packages/cpp/Memory1.json
jeongsoolee09: Added in [60ddfaa](https://github.com/github/codeql-coding-standards/pull/967/commits/60ddfaafc88a440410494a5b4fed2ac68ba24f6f).

cpp/misra/src/rules/RULE-8-7-1/PointerArithmeticFormsAnInvalidPointer.ql
```diff
@@ -206,7 +220,24 @@ class IndirectUninitializedNode extends Node {
...
+     *
+     * ``` C++
+     * int arr[4];
+     * int *third_pos = &arr[2];              // current offset is 2.
+     * int *one_beyond_last = third_pos + 2;  // current offset is 4 (which is just one beyond the last).
+     * ```
+     *
+     * &
+     *
+     * c.f. `TAddressOfIndex` branch of `PointerFormation`.
```
MichaelRFairhurst, line 238: Nice!

MichaelRFairhurst:
LGTM once the build passes!

MichaelRFairhurst (approved):

cpp/common/src/codingstandards/cpp/OutOfBounds.qll
jeongsoolee09: I think a more fundamental solution is to have a radical refactor (that is of enormous magnitude and has to be surgically precise) that makes `cpp` depend on `c`, as C++ is (almost, but 100% in practice) a superset of C.

Then we won't have to keep a copy here which is awkward.

PR #992, opened 2025-12-04
Implement rule 28-6-1, only move non-const lvalues

MichaelRFairhurst:
## Description

Implement 28-6-1, std::move with const will not move, and std::move with a non-lvalue is redundant.

Unfortunately, `expr.getLvalueCategory()` did not work out of the box, as it generates extra rvalue for load steps that are transparent to the user, so we have to navigate lvalue to rvalue conversion etc to handle properly. `isLvalue()` also exists, but its an incomplete syntactic match. Further, the definition of an lvalue/xvalue/etc changes across C++ versions, which that function doesn't handle. While MISRA C++ 2023 currently requires C++17, we want our queries to work for users that don't comply with that restriction, for instance, users using the next version of MISRA C++ :)

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - `RULE-28-6-1`
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


cpp/misra/test/rules/RULE-28-6-1/test.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+void f3() {
+  std::string l1{"hello"};
+  std::string &l2 = l1;
+  std::string const &l3 = l1;
+  std::string const l4{"hello"};
+  std::string &&l5 = std::string{"hello"};
+  const std::string &&l6 = std::string{"hello"};
+
+  func(std::move(l1)); // COMPLIANT - non-const lvalue
```
jeongsoolee09: I wonder why wrapping `std::move` calls with `func(...)` is needed?

cpp/common/src/codingstandards/cpp/ast/ValueCategory.qll
```diff
@@ -0,0 +1,60 @@
...
+
+  predicate isLValue() { this instanceof LValue }
+
+  predicate isPRValue() { this instanceof PRValue }
+
+  predicate isXValue() { this instanceof XValue }
+
+  predicate isRValue() { this instanceof PRValue or this instanceof XValue }
+
+  predicate isGlvalue() { this instanceof LValue or this instanceof XValue }
```
jeongsoolee09: A more consistent pascal-casing would be `isGLValue`.

cpp/common/src/codingstandards/cpp/ast/ValueCategory.qll
```diff
@@ -0,0 +1,60 @@
...
+  if e.isLValueCategory()
+  then result = LValue(e.getValueCategoryString())
+  else
+    if e.isPRValueCategory()
+    then result = PRValue(e.getValueCategoryString())
+    else
+      if e.isXValueCategory()
+      then result = XValue(e.getValueCategoryString())
+      else none()
+}
```
jeongsoolee09, line 92: I have noticed that `getValueCategory(expr)` yields the same answer as `Expr.is___ValueCategory(expr)` that the argument `x` of `f(x)` is a `prvalue(load)` in the code below:

``` C++
void f(int) {}
void g(int &) {}
template <typename T> void h(T &&) {}

typedef int &IntRef;

IntRef getRef(IntRef x) { return x; }

int main() {
  /* 1. `int x` as arguments to parameter, lvalue ref */
  int x = 42;
  f(x);    // <= HERE!
  g(x);
  h(x);

  /* 2. `IntRef y` as arguments to parameter, lvalue ref */
  IntRef y = x;
  f(getRef(y));
  g(getRef(y));
  h(getRef(y));

  return 0;
}
```
This is because of this line here:
``` C++
  not exists(e.getConversion()) and
  result = getDirectValueCategory(e)
```
Surprisingly, the argument `x` to the call `f(x)` does not have a conversion. I have dived more into the issue because it does not match common sense; it turns out `Expr::getConversion/0` and `Expr::hasLValueToRValueConversion/0` depend on different dbscheme tables.

``` codeql
  predicate hasLValueToRValueConversion() { expr_isload(underlyingElement(this)) }

  Expr getConversion() { exprconv(underlyingElement(this), unresolveElement(result)) }
```
(`exprconv/2` is what `Expr::hasConversion/0`, `Expr::hasImplicitConversion/0`, `Expr::hasExplicitConversion/0` depend on as well. `hasLValueToRValueConversion` is an outlier; it is the _only_ predicate in all of `cpp-all` that uses `expr_isload`.)

The result is that the argument of `f(x)` is still detected as `prvalue(load)`. All the other arguments `x` in the example is detected correctly as `lvalue` and `glvalue` because those have conversions. The `f(x)` problem may have not surfaced during development against `test.cpp`, because `std::move`'s parameter is a forwarding reference, similar to `h` above.

So, to make this predicate more perfect (to at least cover the `f(x)` case above, we need an additional check on `expr.hasLValueToRValueConversion`:

```suggestion
ValueCategory getValueCategory(Expr e) {
  not (exists(e.getConversion()) or e.hasLValueToRValueConversion()) and
  result = getDirectValueCategory(e)
  or
  (
    if e.getConversion() instanceof ReferenceToExpr and not e.hasLValueToRValueConversion()
    then result = getDirectValueCategory(e)
    else result = getDirectValueCategory(e.getConversion())
  )
  or
  (
    if e.hasLValueToRValueConversion()
    then result = LValue("lvalue")
    else result = PRValue("prvalue")
  )
}
```
(The code is not elegant; but you get the idea.)

jeongsoolee09: The A side of the diff includes `getDirectValueCategory`; that's not intended.

jeongsoolee09 (changes requested):

MichaelRFairhurst:
Rebased to remove dependency on https://github.com/github/codeql-coding-standards/pull/991

cpp/common/src/codingstandards/cpp/ast/ValueCategory.qll
MichaelRFairhurst: This should be done.

I beefed up the tests for 28-6-1, and found other issues, and I think I found the underlying cause of all of this headache (CodeQL conversions vs spec behavior).

I also added tests for ValueCategory.qll to test the cases you found!

cpp/misra/test/rules/RULE-28-6-1/test.cpp
MichaelRFairhurst: Tests in general have been both cleaned up, and expanded.

cpp/misra/test/rules/RULE-28-6-1/test.cpp
```diff
@@ -0,0 +1,305 @@
...
+    f1();
+    x = other.x;
+    return *this;
+  }
+};
+
+void fstr(std::string param);
+void fint(int param);
+void fmovable(TrivMovable param);
+void fnontriv(NonTrivMovable param);
```
jeongsoolee09: I see several sets of test cases that use these four functions, what do they intend to test?

MichaelRFairhurst: Simplified in 51fc895c6 !

jeongsoolee09 (approved):
LGTM!

PR #1003, opened 2025-12-18
First implementation of Rule-0-0-2, invariant conditions.

MichaelRFairhurst:
## Description

Implements Rule 0-0-2, invariant conditions. Similar to autosar's M0-1-2 but with different exceptional cases.

I tried using modules as a new form of query behavior sharing here. At some point in the future, we may want the shared query generator to generate this template instead of the existing one that uses abstract classes and an unparameterized problems query.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - `RULE-0-0-2`
     - `M-0-1-2`

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 

### DIFF
--- a/rule_packages/cpp/DeadCode4.json
+++ b/rule_packages/cpp/DeadCode4.json
@@ -10,7 +10,7 @@
           "description": "Invariant expressions in controlling statements can indicate logic errors or redundant code.",
           "kind": "problem",
           "name": "Controlling expressions should not be invariant",
-          "precision": "very-high",
+          "precision": "high",
           "severity": "error",
           "short_name": "InvariantCondition",
           "tags": [
--- a/cpp/misra/src/rules/RULE-0-0-2/InvariantCondition.ql
+++ b/cpp/misra/src/rules/RULE-0-0-2/InvariantCondition.ql
@@ -4,7 +4,7 @@
  * @description Invariant expressions in controlling statements can indicate logic errors or
  *              redundant code.
  * @kind problem
- * @precision very-high
+ * @precision high
  * @problem.severity error
  * @tags external/misra/id/rule-0-0-2
  *       scope/system


rule_packages/cpp/DeadCode4.json
```diff
@@ -0,0 +1,26 @@
...
+      "properties": {
+        "enforcement": "undecidable",
+        "obligation": "advisory"
+      },
+      "queries": [
+        {
+          "description": "Invariant expressions in controlling statements can indicate logic errors or redundant code.",
+          "kind": "problem",
+          "name": "Controlling expressions should not be invariant",
+          "precision": "very-high",
```
mbaluda: Should we reduce the precision to `medium`, considering:
- known limitations in `SimpleRangeAnalysis ` 
- the MISRA rule is stated in a general way? (e.g. it includes the case "sum of six, consecutive, non-negative integers is always an odd number")

MichaelRFairhurst: Good question! In this case, we're using SimpleRangeAnalysis conservatively, so we'll get more FNs than FPs, and perhaps we should go with "High"?

I reran this query against various open source repos, and they're certainly pretty noisy. The biggest issue was caused by macros, I had an "and" in the exclusion predicate that should have been an "or," and I updated that.

The other common source of issues is in templates. But I think these are TPs, because the rule implies that these cases should be replaced with `constexpr if`.

What do you think?

MichaelRFairhurst: Edit: I forgot that the rule explicitly states that macros can produce do-while-false, so we shouldn't categorically exclude macros even though they're extremely noisy.

This has me considering medium again because the macros produce so much noise, however, this strongly suggests that these alerts are from the query working as intended...

mbaluda (approved):

PR #1016, opened 2026-01-15
Add linkage1 package

knewbury01:
## Description

adds a package for linkage1 - contains 2 rules which are shared implementations

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - RULE-6-5-1
     - RULE-6-0-2
     - A3-3-1

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [x] Confirmed

### Author

### Reviewer 

### DIFF
--- a/cpp/common/src/codingstandards/cpp/rules/externallinkagearraywithoutexplicitsize/ExternalLinkageArrayWithoutExplicitSize.qll
+++ b/cpp/common/src/codingstandards/cpp/rules/externallinkagearraywithoutexplicitsize/ExternalLinkageArrayWithoutExplicitSize.qll
@@ -1,7 +1,6 @@
 /**
  * Provides a library with a `problems` predicate for the following issue:
- * Introducing a function or object with external linkage outside of a header file can
- * cause developer confusion about its translation unit access semantics.
+ * Declaring an array with external linkage without its size being explicitly specified can disallow consistency and range checks on the array size and usage.
  */
 
 import cpp
@@ -21,9 +20,9 @@ query predicate problems(DeclarationEntry declEntry, string message) {
         "' with external linkage doesn't specify the size explicitly." and
     hasExternalLinkage(v) and
     not arrayType.hasArraySize() and
-    // Holds is if declEntry is an array variable declaration (not a definition)
+    // Holds if declEntry is an array variable declaration (not a definition)
     v.getADeclarationEntry() = declEntry and
     not declEntry.isDefinition() and
-    arrayType = v.getType().stripTopLevelSpecifiers()
+    arrayType = declEntry.getType().stripTopLevelSpecifiers()
   )
 }
--- a/cpp/common/test/rules/externallinkagearraywithoutexplicitsize/test.cpp
+++ b/cpp/common/test/rules/externallinkagearraywithoutexplicitsize/test.cpp
@@ -24,6 +24,13 @@ struct s {
   int foo;
 
   // static data members have external linkage - but not currently detected in
-  // our external linkage lib - also FAMs are not in scope for this rule
-  static const int flexibleArrayMember[]; // COMPLIANT
+  // our external linkage lib - also FAMs are expected to be detected
+  // specifically in RULE-18-7
+  static const int flexibleArrayMember[]; // NON_COMPLIANT[FALSE_NEGATIVE]
+  static int flexibleArrayMember2[];      // NON_COMPLIANT[FALSE_NEGATIVE]
 };
+
+// test.cpp
+#include "test.hpp"
+// definition associated with a declaration from test.hpp
+int header_and_cpp[1] = {1};
--- a/cpp/common/test/rules/externallinkagearraywithoutexplicitsize/ExternalLinkageArrayWithoutExplicitSize.expected
+++ b/cpp/common/test/rules/externallinkagearraywithoutexplicitsize/ExternalLinkageArrayWithoutExplicitSize.expected
@@ -1,1 +1,2 @@
 | test.cpp:19:14:19:15 | declaration of e1 | The declared array 'e1' with external linkage doesn't specify the size explicitly. |
+| test.hpp:2:13:2:26 | declaration of header_and_cpp | The declared array 'header_and_cpp' with external linkage doesn't specify the size explicitly. |
--- a/cpp/common/test/rules/externallinkagearraywithoutexplicitsize/test.hpp
+++ b/cpp/common/test/rules/externallinkagearraywithoutexplicitsize/test.hpp
@@ -1,0 +1,2 @@
+// test.hpp
+ extern int header_and_cpp[]; // NON_COMPLIANT


change_notes/2026-01-15-a3-1-4-extern-to-full.md
```diff
@@ -0,0 +1,2 @@
+- `A3-1-4` - `ExternalLinkageArrayWithoutExplicitSizeAutosar.ql`:
+   - `ExternalLinkageArrayWithoutExplicitSize.ql` has been renamed to `ExternalLinkageArrayWithoutExplicitSizeAutosar.ql` to reflect shared query implementation. Additionally the query previously only detected explicit uses of `extern` to determine external linkage, and now would catch other cases that are possible where it is external linkage and an array is declared without an explicit size.
```
MichaelRFairhurst, line 2: Very nice to catch and fix this!

cpp/common/src/codingstandards/cpp/rules/externallinkagearraywithoutexplicitsize/ExternalLinkageArrayWithoutExplicitSize.qll
```diff
@@ -0,0 +1,29 @@
+/**
+ * Provides a library with a `problems` predicate for the following issue:
+ * Introducing a function or object with external linkage outside of a header file can
+ * cause developer confusion about its translation unit access semantics.
```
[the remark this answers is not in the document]
MichaelRFairhurst: Seems copilot is right here

cpp/common/test/rules/externallinkagearraywithoutexplicitsize/test.cpp
```diff
@@ -0,0 +1,29 @@
...
+  int a2[1];      // COMPLIANT - not external linkage
+
+  extern int e1[]; // NON_COMPLIANT
+}
+
+struct s {
+  // Structs must have at least one non-flexible array member.
+  int foo;
+
+  // static data members have external linkage - but not currently detected in
```
MichaelRFairhurst, line 26: Can you add a case for this?

```
static int[] x; // NON_COMPLIANT[False negative]
```

MichaelRFairhurst (changes requested):
Just small suggestions!

This looks great!

And ❗ great find ❗  that, not only could this logic be shared, but also, it could be improved!! 🎉

cpp/common/src/codingstandards/cpp/rules/externallinkagearraywithoutexplicitsize/ExternalLinkageArrayWithoutExplicitSize.qll
```diff
@@ -0,0 +1,29 @@
...
+Query getQuery() { result instanceof ExternalLinkageArrayWithoutExplicitSizeSharedQuery }
+
+query predicate problems(DeclarationEntry declEntry, string message) {
+  exists(Variable v, ArrayType arrayType |
+    not isExcluded(declEntry, getQuery()) and
+    message =
+      "The declared array '" + declEntry.getName() +
+        "' with external linkage doesn't specify the size explicitly." and
+    hasExternalLinkage(v) and
+    not arrayType.hasArraySize() and
```
MichaelRFairhurst, line 22: I'm surprised that `hasArraySize` was working before..!

_Below are a bunch of thoughts that kind of go down a rabbit hole._ Consider it 100% up to you how you want to proceed -- you could file a bug and copy my comment, or you can explore it, add tests, etc.

We may not be able to do much about it, but it may be worth adding some `.hpp` test cases to see if the following are FNs:

```
// test.hpp
int header_only[]; // NON_COMPLIANT
int header_and_cpp[]; // NON_COMPLIANT

// test.cpp
#include "test.hpp"
// definition associated with a declaration from test.hpp
int header_and_cpp[10] = ...
```

Basically, I fear that `header_and_cpp[]` will be a false negative. It should be extracted as one `Variable`, with one `Type` -- that type probably being an int array of size 10.

If this is a FN we can write `NON_COMPLIANT[False negative]` along with an explanatory comment, and I'm not sure if we have any recourse to fix it if so. However, we likely don't _have to_ fix it.....if it _is_ a FN, we probably _also_ want to add a test like:

```
// test.hpp
// an array declaration that's defined in test1.cpp but not test2.cpp
int defined_in_test1_cpp[]; // NON_COMPLIANT

// test1.cpp
int defined_in_test1_cpp[10] = ...;

// test2.cpp
// This file exists to _not_ do the following:
// int defined_in_test1_cpp[10] = ...;
// So that we can test that a violation of the One Definition Rule is caught in this case
```

My guess is that this will be correctly reported as non compliant because we would have two variables (a violation of ODR), and one would have no size. This would means the `NON_COMPLIANT[False negative]` is low impact.

Let me know if any of this is unclear, feel free to reach out via chat etc. Up to you what next steps you think may be best!

cpp/common/test/rules/externallinkagearraywithoutexplicitsize/test.cpp
knewbury01: sure! originally I had wrote in the comment that I think this is maybe COMPLIANT for this rule? it is external linkage, but also its a FAM, and there is already a specific rule dedicated to detecting FAM usage (RULE-18-7) 

but I think that this rule can also superset those results, and a FAM would be included, in which case I added the testcase! thanks!

cpp/common/src/codingstandards/cpp/rules/externallinkagearraywithoutexplicitsize/ExternalLinkageArrayWithoutExplicitSize.qll
knewbury01: I think that `int header_and_cpp[]; // NON_COMPLIANT` is a compiler checked case, see the other global (therefore external linkage) testcases similar to this in that test.cpp file, they are commented out bc the compiler needs the size in these declarations. so technically they violate the rule but they are also already compiler checked (ie this cant compile) ... I am not aware of a compiler flag that allows for ignoring this , but lmk if I am missing something!

knewbury01: oh! this definitely compiles with explicit `extern` usage though, and I think would still be a violation of the rule.

I think your notes are correct, it shouldnt be the variable that is used to check if the size is at all specified, that was a logic error on my part and I have now changed it so that the external linkage is still overall checked on the variable but the presence of any decl without a size is a problem (ie we now catch the testcase above, and have added it to the testcase also)

thanks!

cpp/common/src/codingstandards/cpp/rules/externallinkagearraywithoutexplicitsize/ExternalLinkageArrayWithoutExplicitSize.qll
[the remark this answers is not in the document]
knewbury01: oops , yeah I did a package file gen wrong at one point. good catch , thanks!

cpp/common/src/codingstandards/cpp/rules/externallinkagearraywithoutexplicitsize/ExternalLinkageArrayWithoutExplicitSize.qll
MichaelRFairhurst: Nice job finding the fix!!

PR #1024, opened 2026-02-03
Add lifetime package

knewbury01:
## Description

add rules for lifetime package, two shared rules, one with improvements upon previous, one as simple shared

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - RULE-11-6-2
     - RULE-6-8-3
     - _rule number here_

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


cpp/common/test/rules/readofuninitializedmemory/test.cpp
```diff
@@ -123,4 +123,75 @@ void test_non_default_init() {
...
+void extra_test() {
+  int i;
+  int j = i + 1; // NON_COMPLIANT
+
+  int *i1 = new int;
+  int i2 = *i1; // NON_COMPLIANT
+
+  int *i3;
+
+  if (i3 = i1) { // NON_COMPLIANT
```
MichaelRFairhurst: Single `=` should be `==`

MichaelRFairhurst (commented):
The test cases look great Kristen, thank you!

cpp/common/src/codingstandards/cpp/lifetimes/CppObjects.qll
```diff
@@ -262,6 +262,8 @@ class AggregateLiteralObjectIdentity extends AggregateLiteral, ObjectIdentityBas
 class AllocatedObjectIdentity extends AllocationExpr, ObjectIdentityBase {
   AllocatedObjectIdentity() {
     this.(FunctionCall).getTarget().(AllocationFunction).requiresDealloc()
+    or
+    this = any(NewOrNewArrayExpr new | not exists(new.getPlacementPointer()))
```
knewbury01, line 266: borrowed this from [here](https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/commons/Alloc.qll#L21)

cpp/common/src/codingstandards/cpp/rules/readofuninitializedmemory/ReadOfUninitializedMemory.qll
```diff
@@ -121,13 +122,39 @@ class InitializationContext extends TInitializationContext {
...
+    (
+      not exists(getInitializer().getExpr())
+      or
+      //or is a builtin type used with new operator but there is no value initialization as far as we can see
+      exists(Initializer i, NewExpr n |
+        i = getInitializer() and
+        n = i.getExpr() and
+        this.getUnspecifiedType().stripType() instanceof BuiltInType and
+        //ignore value init
+        not exists(n.getAChild())
```
MichaelRFairhurst: Would this work?

```suggestion
      //or is a builtin type used with new operator but there is no value initialization as far as we can see
      getInitializer().getExpr() instanceof NewNotInit
```

cpp/common/src/codingstandards/cpp/rules/readofuninitializedmemory/ReadOfUninitializedMemory.qll
```diff
@@ -121,13 +122,39 @@ class InitializationContext extends TInitializationContext {
   }
 }
 
+class NewNotInit extends NewExpr {
```
MichaelRFairhurst, line 129: Can you add a comment explaining the limitations somewhere in this file, maybe here?

Something like, "Catches `new int;` as an expression that doesn't initialize its value. Note that the pointer returned has been initialized, but the pointee has not. In our analysis, we simply count x as uninitialized in `x = new int` for now, though a more thorough analysis might track the initialization of `x` is and `*x` separately."

cpp/common/src/codingstandards/cpp/rules/readofuninitializedmemory/ReadOfUninitializedMemory.qll
```diff
@@ -176,16 +203,33 @@ class UninitializedVariable extends LocalVariable {
...
+   * This approximates that any access in the lvalue category may be a definition.
+   */
   VariableAccess getADefinitionAccess() {
     result = getAnAccess() and
-    result.isLValueCategory()
+    result.isLValueCategory() and
+    // Not a pointless read
+    not result = any(ExprStmt es).getExpr() and
+    // not involved in a new expr assignment since that does not define
+    not result instanceof VariableAccessOnLHSOfNonInitAssignment
```
MichaelRFairhurst:
```suggestion
    not result = any(NonInitAssignment a).getLValue()
```

This class name is pretty complicated and makes the code seem maybe more challenging than it is.

cpp/common/src/codingstandards/cpp/rules/readofuninitializedmemory/ReadOfUninitializedMemory.qll
```diff
@@ -176,16 +203,33 @@ class UninitializedVariable extends LocalVariable {
...
+   */
   VariableAccess getAUse() {
-    result = getAnAccess() and
-    result.isRValueCategory() and
+    result = this.getAnAccess() and
+    // Not used as an lvalue
+    not result = any(AssignExpr a).getLValue() and
+    // Not passed to another initialization function
+    not exists(Call c, int j | j = c.getTarget().(InitializationFunction).initializedParameter() |
+      result = c.getArgument(j).(AddressOfExpr).getOperand()
```
MichaelRFairhurst, line 236: Consider adding `or result = c.getArgument(j)`, that would handle:

```
void f(int *p) {
  *p = 0;
}

void g() {
  int *p = new int;
  f(p); // compliant
  use(*p); // compliant
}
```

In fact we might want to rely on `.isRValueCategory() and not exists(Call ... | ... result = c.getArgument(j))`.

In general, if `x` appears in `&x`, that should mean imply not `.isRValueCategory()` for `x`. So technically, `int *p = &x` would previously have not been considered a use and now will be.

Overall, `isRValueCategory()` should find only cases where `x` could _not_ be initialized (only reads + copies). So we were already very conservative in what we consider a use.

The catch is that copying or reading `x` may be how someone initializes `*x`, such as in the above example with `f(p)`. So with our new `x`+`*x` analysis, we can't guarantee `isRValueCategory()` for `x` means we don't initialize `*x`. Really _any_ rvalue use could lead to initialization of `*x`.

We also don't want to lose the ability to analyze uninitialized pointers more generally (as in, `int *p; f(p);` which is never valid and can segfault, and the current query should catch).

Ultimately, for `*x` vs `x`:
- if `*x` is used as an rvalue, that's problematic if `x` had a non-init new
- otherwise, all uses of `x` as an rvalue require that `x` had _any_ initializer
- If `x` is used as an lvalue in `*x`, that's problematic if `x` had _no_ initializer
- All other lvalue uses of `x` are valid
- It's always valid if `*x` is used as an lvalue if x had _any_ initializer

Ultimately, I think it would probably be fine to say:
- track `x = new int` as an "uninitialized variable"
- count rvalue `x` as a use if **not** `new int`
- count lvalue `x` as a use if used in `*x` and **not** `new int`
- count rvalue `*x` as a use if **has** `new int`
- don't count lvalue `*x` as a use in and of itself

change_notes/2026-02-03-uninitialized-mem-improve.md
```diff
@@ -0,0 +1,2 @@
+- `A8-5-0`, `EXP53-CPP`, `EXP33-C`, `RULE-9-1` - `MemoryNotInitializedBeforeItIsRead.ql`, `DoNotReadUninitializedMemory.ql`, `DoNotReadUninitializedMemory.ql`, `ObjectWithAutoStorageDurationReadBeforeInit.ql`:
+  - The queries listed now find uses of the operator 'new' where there is no value initialization provided.
```
MichaelRFairhurst: can you note that we don't yet finely track the initialization/use of `p` vs `*p`

And add a line that we're now using an existing library to more finely track initialization through functions?

cpp/common/test/rules/readofuninitializedmemory/test.cpp
```diff
@@ -123,4 +123,101 @@ void test_non_default_init() {
...
+
+  C c2(5);
+  if (c2.getm2() > 0) { // NON_COMPLIANT[FALSE_NEGATIVE] - rule currently is not
+                        // field sensitive
+  }
+}
+
+void extra_extra_test() {
+  int *p0 = new int;
+  p0; // COMPLIANT -- this is uninitialized but its access not a meaningful read
```
MichaelRFairhurst: add cases for `use(p0)` and `use(*p0)`

cpp/common/test/rules/readofuninitializedmemory/test.cpp
```diff
@@ -123,4 +123,101 @@ void test_non_default_init() {
...
+      // to this rule
+
+  int *p1 = new int;
+  *p1 = 0; // COMPLIANT[FALSE_POSITIVE] -- this is not found bc this is not an
+           // lvalue access
+  *p1; // COMPLIANT[FALSE_POSITIVE] -- the pointee of p1 has been initialized
```
MichaelRFairhurst:
```suggestion
  use(*p1); // COMPLIANT[FALSE_POSITIVE] -- the pointee of p1 has been initialized
```

cpp/common/test/rules/readofuninitializedmemory/test.cpp
```diff
@@ -123,4 +123,101 @@ void test_non_default_init() {
...
+
+  int *p2 = new int;
+  p2 = new int;
+  p2; // COMPLIANT -- this is uninitialized but its access not a meaningful read
```
MichaelRFairhurst:
```suggestion
  p2; // COMPLIANT -- this is uninitialized but its access not a meaningful read
  use(p2);
  use(*p2);
```

cpp/common/test/rules/readofuninitializedmemory/test.cpp
```diff
@@ -123,4 +123,101 @@ void test_non_default_init() {
...
+      // to this rule
+
+  int *p3 = new int(1);
+  *p3 = *p2; // NON_COMPLIANT -- the pointee of p2 has not been
+             // initialized
+  *p3; // NON_COMPLIANT[FALSE_NEGATIVE] -- the pointee of p3 has be overridden
```
MichaelRFairhurst:
```suggestion
  use(p3);
  use(*p3); // NON_COMPLIANT[FALSE_NEGATIVE] -- the pointee of p3 has be overridden
```

cpp/common/test/rules/readofuninitializedmemory/test.cpp
```diff
@@ -123,4 +123,101 @@ void test_non_default_init() {
...
+
+  int *p4;
+  p4 = new int;
+  *p4; // NON_COMPLIANT -- the pointee of p4 has not been
```
MichaelRFairhurst:
```suggestion
  use(*p4); // NON_COMPLIANT -- the pointee of p4 has not been
  use(p4)
```

rule_packages/cpp/Lifetime.json
```diff
@@ -0,0 +1,53 @@
...
+      "queries": [
+        {
+          "description": "Reading from uninitialized indeterminate values may produce undefined behavior.",
+          "kind": "problem",
+          "name": "The value of an object must not be read before it has been set",
+          "precision": "medium",
+          "severity": "error",
+          "short_name": "ValueOfAnObjectMustNotBeReadBeforeItHasBeenSet",
+          "shared_implementation_short_name": "ReadOfUninitializedMemory",
+          "tags": [
```
MichaelRFairhurst, line 17: Can you add an implementation_scope note that we don't track class member initialization?

rule_packages/cpp/Lifetime.json
```diff
@@ -0,0 +1,53 @@
...
+    "RULE-6-8-3": {
+      "properties": {
+        "enforcement": "decidable",
+        "obligation": "required"
+      },
+      "queries": [
+        {
+          "description": "When storage durations are not compatible between assigned pointers it can lead to referring to objects outside of their lifetime, which is undefined behaviour.",
+          "kind": "problem",
+          "name": "Declare objects with appropriate storage durations",
```
MichaelRFairhurst: Should the name be about assignments of different lifetimes instead of declarations?

cpp/common/test/rules/readofuninitializedmemory/test.cpp
```diff
@@ -123,4 +123,101 @@ void test_non_default_init() {
...
+       // initialized
```
MichaelRFairhurst: Can we add a test for initialization via a function?

```
int *p5;
initialize(p5); // NON_COMPLIANT -- segfault

int *p6 = new int;
initialize(p6) // COMPLIANT
```

cpp/common/src/codingstandards/cpp/rules/readofuninitializedmemory/ReadOfUninitializedMemory.qll
knewbury01: yep! excellent catch on this simplification

cpp/common/src/codingstandards/cpp/rules/readofuninitializedmemory/ReadOfUninitializedMemory.qll
knewbury01: yep agree much more readable

cpp/common/test/rules/readofuninitializedmemory/test.cpp
knewbury01: I think this is the outdated file, so these are already updated

cpp/common/test/rules/readofuninitializedmemory/test.cpp
knewbury01: same as above

cpp/common/test/rules/readofuninitializedmemory/test.cpp
knewbury01: same as above

cpp/common/test/rules/readofuninitializedmemory/test.cpp
knewbury01: ohh. you mean have both, yep sure!

rule_packages/cpp/Lifetime.json
knewbury01: yep makes sense to me!

cpp/common/test/rules/readofuninitializedmemory/test.cpp
knewbury01: added this case as well

rule_packages/cpp/Lifetime.json
knewbury01: yeah that is much better, 
I kinda think it makes a lot of sense for the shared queries to have the same description/name/severities etc

but we havent been totally uniform about these all of the time. I chose to copy the name from the declarations package version, but that one has the least accurate name for what they all intend to do. changed it now to align with the Freed pkg one

cpp/common/test/rules/readofuninitializedmemory/test.cpp
knewbury01: yep! added one for the valid case too, where it actually is initialized within the initialize function

knewbury01: I think 
```
int *p6 = new int;
initialize(p6) // COMPLIANT
```
is still NON_COMPLIANT though , right? its allocated but not initialized

knewbury01: ah right nvm I was thinking in terms of the rule thinking that this is technically a use... even though its not a read. righto! it is a compliant FP rn

cpp/common/src/codingstandards/cpp/rules/readofuninitializedmemory/ReadOfUninitializedMemory.qll
knewbury01: I think I have addressed this as best as I can, please carefully double check my logic/annotations in the testcases, there were a few points where I switched the expectations on some cases a lot depending on what we decided a use was

this all makes sense enough to me though to do so I have added the adjustments!

cpp/common/test/rules/readofuninitializedmemory/test.cpp
```diff
@@ -123,4 +123,120 @@ void test_non_default_init() {
...
+                        // field sensitive
+  }
+}
+
+void initialize(int *p) { *p = 0; }
+
+void extra_extra_test() {
+  int *p0 = new int;
+  use(p0);  // COMPLIANT -- the pointer is valid
+  use(*p0); // COMPLIANT[FALSE_POSITIVE] -- the pointer is valid
```
MichaelRFairhurst: I believe this is correctly NON_COMPLIANT, because this copies `*p`, and we haven't assigned to `*p` yet! Nice!

cpp/common/test/rules/readofuninitializedmemory/test.cpp
```diff
@@ -123,4 +123,120 @@ void test_non_default_init() {
...
+
+  int *p2 = new int;
+  p2 = new int;
+  use(p2);  // COMPLIANT -- the pointer is valid
+  use(*p2); // NON_COMPLIANT -- the value may be read
+
+  int *p3 = new int(1);
+  *p3 = *p2; // NON_COMPLIANT -- the pointee of p2 has not been
+             // initialized
+  use(p3);   // NON_COMPLIANT[FALSE_NEGATIVE] -- the pointee of p3 has been
```
MichaelRFairhurst: This is correctly compliant, as `p3` itself is still the same pointer value returned by `new int` and therefore initialized.

cpp/common/src/codingstandards/cpp/rules/readofuninitializedmemory/ReadOfUninitializedMemory.qll
```diff
@@ -176,20 +196,72 @@ class UninitializedVariable extends LocalVariable {
...
+  //   not exists(Call c, int j | j = c.getTarget().(InitializationFunction).initializedParameter() |
+  //     result = c.getArgument(j).(AddressOfExpr).getOperand()
+  //     or
+  //     result.isRValue() and result = c.getArgument(j)
+  //   ) and
+  //   // Not a pointless read
+  //   not result = any(ExprStmt es).getExpr() and
+  //   // sizeof operators are not real uses
+  //   not result.getParent+() instanceof SizeofOperator
+  // }
```
MichaelRFairhurst: dead code

cpp/common/test/rules/donotcopyaddressofautostorageobjecttootherobject/test.cpp
```diff
@@ -6,4 +6,23 @@ void test_simple_aliasing() {
     int c;
     a = &c; // NON_COMPLIANT - different scope
   }
+}
+
+void extra_test_simple_aliasing() {
+  int *p;
+  {
+    int l_a[1];
+    p = l_a; // NON_COMPLIANT
```
MichaelRFairhurst, line 15: No action to take, just commenting that it's great that the query catches this! I did not realize, awesome!

cpp/common/src/codingstandards/cpp/rules/readofuninitializedmemory/ReadOfUninitializedMemory.qll
```diff
@@ -176,20 +196,72 @@ class UninitializedVariable extends LocalVariable {
...
+   * Gets an access of the this variable which is not used as an lvalue, and not used as an argument
+   * to an initialization function.
+   */
   VariableAccess getAUse() {
-    result = getAnAccess() and
-    result.isRValueCategory() and
+    result = this.getAnAccess() and
+    (
+      //count rvalue x as a use if not new int
+      result.isRValue() and
```
MichaelRFairhurst, line 220: Does `isRValue` -> `isRValueCategory`, `isLValue` -> `isLValueCategory` fix the FP on 207?

cpp/common/src/codingstandards/cpp/rules/readofuninitializedmemory/ReadOfUninitializedMemory.qll
```diff
@@ -176,20 +196,72 @@ class UninitializedVariable extends LocalVariable {
...
+      not exists(PointerDereferenceExpr e | result = e.getAChild()) and
```
MichaelRFairhurst: I believe this condition is unnecessary.

Any time x has no initializer (not even initialized with `new int;`) then any rvalue use will be a random pointer address. It's uninitialized to read (as in line 141 `p1 == p2`), and it would also be invalid to dereference it with `*`.

cpp/common/src/codingstandards/cpp/rules/readofuninitializedmemory/ReadOfUninitializedMemory.qll
knewbury01: oops

cpp/common/test/rules/readofuninitializedmemory/test.cpp
knewbury01: very true, my bad!

cpp/common/src/codingstandards/cpp/rules/readofuninitializedmemory/ReadOfUninitializedMemory.qll
knewbury01: agree! any reference to or deref on a pointer that is not allocated should be a use, good catch

this fixes that line 230 ` use(*p4);` is technically a use

cpp/common/src/codingstandards/cpp/rules/readofuninitializedmemory/ReadOfUninitializedMemory.qll
knewbury01: these dont change anything on the unit test, (overall result or getAUse wise) - on openpilot the overall results are also same but the set of `isRValue()` (not `isRValueCategory`) uses means `getAUse` gets more results... seems broader. the results of uses in the broader set, on openpilot seem to make sense to me (from just sampling)
but I do feel I dont quite know what the categories of difference are

knewbury01: additionally I sortof followed suit from use definitions from [here ](https://github.com/github/codeql/blob/d52e9bc18c9895f43d8855ad22bfa360d04b1932/cpp/ql/src/Critical/NotInitialised.ql#L31)and [here](https://github.com/github/codeql/blob/d52e9bc18c9895f43d8855ad22bfa360d04b1932/cpp/ql/src/Critical/GlobalUseBeforeInit.ql#L42) and I dont see any uses of `isRValueCategory` in codeql to learn from

PR #1037, opened 2026-02-12
First pass implementing RULE-0-2-1, unused variables.

MichaelRFairhurst:
Implementation relies on shared code with autosar that was previously declared in the ql files. This behavior has been moved into UnusedObjects.qll.

The previous code seemed like a very disorganized set of exceptional cases, which was going to be worse with the new code. I made a minimal attempt to better organize it by organizing it into modules containing individual filter passes.

Filtering out `.getAnAccess()` was always the last step. From a performance perspective, this is a bad order. We start with all variables, and for each one we join it against some set of rare exceptional cases, like functions with assembly, that only minimally reduce the set of variables we want to query. Changed to use a better order, where we begin with the set of variables that have no accesses, which should be a very small set relatively speaking, and then perform the odds-and-ends filters on top of that set. Small performance improvements have been measured across the queries.

## Description

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - `RULE-0-2-1`
     - `M0-1-3`
     - `RULE-2-8`

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 

### DIFF
--- a/cpp/misra/test/rules/RULE-0-2-1/test_header.hpp
+++ b/cpp/misra/test/rules/RULE-0-2-1/test_header.hpp
@@ -19,6 +19,13 @@ const int cg8 = 8; // COMPLIANT - unused constant variable defined in header
 }
 
 class HeaderClass {
-  const int m1 = 9;            // NON_COMPLIANT -- not namespace scope
-  static const int m2 = 30234; // NON_COMPLIANT -- not namespace scope
+  const int m1 = 9;            // NON_COMPLIANT -- no linkage
+  static const int m2 = 30234; // COMPLIANT -- external linkage
 };
+
+namespace {
+class HeaderClass2 {
+  static const int m1 =
+      9; // NON_COMPLIANT -- internal linkage, not namespace scope
+};
+} // namespace
--- a/cpp/misra/src/rules/RULE-0-2-1/UnusedLimitedVisibilityVariable.ql
+++ b/cpp/misra/src/rules/RULE-0-2-1/UnusedLimitedVisibilityVariable.ql
@@ -33,7 +33,7 @@ class UnusedVariable extends Variable {
       description = "Variable '" + this.getName() + "' is unused."
       or
       this instanceof FullyUnusedMemberVariable and
-      description = "Member variable '" + this.getQualifiedName() + "' is unused."
+      description = "Member variable '" + this.getName() + "' is unused."
     ) and
     not (isConstant(this) and this.getFile() instanceof HeaderFile and hasNamespaceScope(this))
   }
--- a/cpp/misra/test/rules/RULE-0-2-1/UnusedLimitedVisibilityVariable.expected
+++ b/cpp/misra/test/rules/RULE-0-2-1/UnusedLimitedVisibilityVariable.expected
@@ -10,6 +10,9 @@
 | test_global_or_namespace.cpp:29:4:29:4 | a | Variable '(unnamed namespace)::a' is unused. |
 | test_global_or_namespace.cpp:37:5:37:6 | x3 | Variable '(unnamed namespace)::N1::x3' is unused. |
 | test_global_or_namespace.cpp:47:5:47:5 | a | Variable '(unnamed namespace)::N1::a' is unused. |
-| test_header.hpp:22:13:22:14 | m1 | Member variable 'HeaderClass::m1' is unused. |
-| test_member.cpp:4:7:4:8 | m1 | Member variable 'test::A::m1' is unused. |
-| test_member.cpp:31:7:31:8 | m1 | Member variable 'test::C::m1' is unused. |
+| test_header.hpp:22:13:22:14 | m1 | Member variable 'm1' is unused. |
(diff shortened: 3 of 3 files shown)


cpp/common/src/codingstandards/cpp/deadcode/UnusedVariables.qll
```diff
@@ -93,97 +87,250 @@ class FullyDefinedClass extends Class {
...
+      // Exclude members whose value is compile time and is potentially used to inintialize a template
+      not maybeACompileTimeTemplateArgument(this)
+    }
+  }
+
+  class UnusedMemberVariable extends SecondPassMemberVariable {
+    UnusedMemberVariable() {
+      // No explicit initialization in a constructor
+      not exists(UserProvidedConstructorFieldInit cfi | cfi.getTarget() = this) and
+      // Exclude members whose value is compile time and is potentially used to inintialize a template
```
mbaluda:
```suggestion
      // Exclude members whose value is compile time and is potentially used to initialize a template
      not maybeACompileTimeTemplateArgument(this)
    }
  }

  class UnusedMemberVariable extends SecondPassMemberVariable {
    UnusedMemberVariable() {
      // No explicit initialization in a constructor
      not exists(UserProvidedConstructorFieldInit cfi | cfi.getTarget() = this) and
      // Exclude members whose value is compile time and is potentially used to initialize a template
```

cpp/common/test/Linkage/test.cpp
mbaluda:
```suggestion
extern const int g4 = 0; // external linkage
```

cpp/misra/test/rules/RULE-0-2-1/test_header.hpp
```diff
@@ -0,0 +1,24 @@
...
+                       // file, with template.
+
+namespace cg8ns {
+const int cg8 = 8; // COMPLIANT - unused constant variable defined in header
+                   // file, in a namespace.
+}
+
+class HeaderClass {
+  const int m1 = 9;            // NON_COMPLIANT -- not namespace scope
+  static const int m2 = 30234; // NON_COMPLIANT -- not namespace scope
```
mbaluda:
```suggestion
  static const int m2 = 30234; // NON_COMPLIANT[FALSE_NEGATIVE] -- not namespace scope
```

cpp/misra/test/rules/RULE-0-2-1/test_member.cpp
```diff
@@ -0,0 +1,64 @@
...
+};
+
+void setm4(A *a) { a->m4 = 0; }
+
+void test_nested_struct() {
+  struct s1 {
+    int sm1 : 7; // COMPLIANT - used
+    int pad : 1; // NON_COMPLIANT - padding bytes, so not used, but is not
+                 // unnamed
+    int sm2 : 6; // NON_COMPLIANT - unused
```
mbaluda, line 19:
```suggestion
    int pad : 1; // NON_COMPLIANT[FALSE_NEGATIVE] - padding bytes, so not used, but is not
                 // unnamed
    int sm2 : 6; // NON_COMPLIANT[FALSE_NEGATIVE] - unused
```

cpp/common/src/codingstandards/cpp/deadcode/UnusedVariables.qll
MichaelRFairhurst: Thank you!

cpp/misra/test/rules/RULE-0-2-1/test_header.hpp
MichaelRFairhurst: Good catch!

Had to reread the spec here. By spec, "static variables at namespace scope" are internal linkage -- but `m2` is not namespace scope. "static data members" inherit the class linkage. In this case, "HeaderClass" has external linkage since its in the global namespace. This should have been COMPLIANT, because `m2` has external linkage, and the rule only applies to variables with internal linkage or no linkage.

I also added a new test case, a static data member of a class in an anonymous namespace. Namespace has internal linkage -> class has internal linkage -> static data member has internal linkage. This new static member is unused by basic rule definition, since it has internal linkage, but still fits some of the criteria of Exception 2: it is constant, and in a header file. However, exception 3 only applies to variables with namespace scope, and this static data member has class scope, so the exception doesn't apply and it is NON_COMPLIANT.

This is a lot but I think that's what my brain was going for when I added this test case originally...?

cpp/misra/test/rules/RULE-0-2-1/test_member.cpp
MichaelRFairhurst: Oof! Another good catch!

I completely missed this, but looks like this is a "regression" (of sorts) since I used `.getQualifiedName()` and `M0-1-3` uses `getName()`.

Changing to `getName()` on `UnusedMemberVariable` in `0-2-1` to match fixed these FNs!

PR #1101, opened 2026-04-02
Declarations5 cpp misra 2023

knewbury01:
## Description

Declarations 5

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - RULE-6-9-1
     - RULE-6-8-4
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


cpp/misra/src/rules/RULE-6-8-4/MemberFunctionsRefqualified.ql
```diff
@@ -0,0 +1,117 @@
...
+abstract class MembersReturningObjectOrSubobject extends MembersReturningPointerOrRef {
+  string toString() { result = "Members returning object or subobject" }
+}
+
+class MembersReturningObject extends MembersReturningObjectOrSubobject {
+  MembersReturningObject() {
+    exists(ReturnStmt r, ThisExpr t |
+      r.getEnclosingFunction() = this and
+      (
+        r.getAChild*() = t
```
MichaelRFairhurst: I'm usually not a fan of `getAChild*()`, because it is typically a bigger hammer than we want. For instance, this will return FPs for `return ref ? 1 : 2`, or `return ref_to_vector.get_size()`, etc.

Can we use `localFlow` instead?

cpp/misra/src/rules/RULE-6-8-4/MemberFunctionsRefqualified.ql
```diff
@@ -0,0 +1,117 @@
...
+    )
+  }
+}
+
+/**
+ * Fields that are not reference type can be subobjects
+ */
+class FieldSubObjectDeclaration extends Declaration {
+  FieldSubObjectDeclaration() {
+    not this.getADeclarationEntry().getType() instanceof ReferenceType and
```
MichaelRFairhurst: This isn't going to do exactly what we probably want, because we have to worry again about pointers and pointees.

Returning a pointer by value is allowed, and dereferencing that pointer should also be allowed because the result could be any object -- not guaranteed to be a subobject of `this`. And returning an address or reference to a field that is a pointer should be flagged as well.

```
class C {
  int i;
  int *ip;
  int **ip2;

  int *f() {
    return &i; // non compliant
    return ip; // compliant -- copies pointer
    return *ip2; // compliant -- reads pointer
  }

  int **f2() {
    return &ip; // non compliant
    return ip2; // compliant -- copies pointer
  }

  int &f3() {
    return i; // non compliant
    return *ip; // compliant -- reads pointer
    return **ip2; // compliant -- reads pointer
  }

  int *&f4() {
    return &p; // won't compile
    return ip; // non compliant
    return *ip2; // compliant -- reads pointer
  }

  int **&f5() {
    return &ip; // won't compile
    return ip; // non compliant
  }
}
```

MichaelRFairhurst (commented):
There are definitely some tricky nuanced scenarios here, but the good news is I think we have some prior art that can handle them, and is still fairly simple. Here's my thoughts.

Consider making `getASubobjectAccessOf` public in `CppObjects.qll` and using that here, so that we can essentially write:

```
predicate isSubobjectReferenceToThis(ThisExpr root, Expr access) {
  // something like this:
  access.(AddressOfExpr).getAnOperand() = getASubobjectAccessOf(root)
  or
  // ... handle the case where we take a reference to a subobject instead of its address
}
```

We likely want to extend `getASubobjectAccessOf` in two ways for C++ (it's still a copy of the c version right now):
- A. to unwrap the `this` pointer, and
- B. to exclude reference accesses

A. would be something like:
```
Expr getASubobjectAccessOfPointee(Expr e) {
  e.getParent() instanceof AddressOfExpr and
  result = getASubobjectAccessOf(e.getParent())
  or
  e.getParent() instanceof PointerFieldAcess() and
  ....
}
```

B. excluding reference accesses, would be something like:
```
Expr getASubobjectAccessOf(Expr e) {
  ...
  or
  result.(DotFieldAccess).getQualifier() = getASubobjectAccessOf(e) and
  // new part:
  not result.(FieldAccess).getField().getUnderlyingType() instanceof ReferenceType
  or
  ...
}
```

Note that there's overlap between A. and B., because `this->x.y` may be a subobject (if `x` and `y` are not references) or not (if either x or y is a reference).

Something beyond A. and B. that we probably don't want to handle right now, but rather file an issue for, would be to add support for simple getter functions, especially for pointer-like types such as `unique_ptr<T>` and iterator types. Definitely reasonable to file in the issue tracker as a TODO.

```
class C {
  Member m;
  Member &get_member() { return m; }
  Member& operator*() { return m; }
  Member *operator->() { return &m; }
}

void f() {
  C c;
  c.get_member(); // accesses member
  *c; // accesses member
  c->bar; // accesses member
```

Hopefully this makes some sense, feel free to ping questions of course!

cpp/misra/src/rules/RULE-6-8-4/MemberFunctionsRefqualified.ql
knewbury01: I have now opted to direct explicit references only... I do agree localFlow within the statement might make the most sense, but it would maybe be overkill? let me know though if that seems like it would miss reasonable things and I will add it!

cpp/misra/src/rules/RULE-6-8-4/MemberFunctionsRefqualified.ql
knewbury01: added this test, as well as incorporated feedback from overall review comment , some FPs remain, I think I need a second set of eyes on my expansion of cppobjects at this point

cpp/common/src/codingstandards/cpp/lifetimes/CppObjects.qll
```diff
@@ -136,19 +137,39 @@ abstract class ObjectIdentityBase extends Element {
...
-  result = e
-  or
-  result.(DotFieldAccess).getQualifier() = getASubobjectAccessOf(e)
+Expr getASubobjectAccessOf(Expr e) {
+  exists(Field f |
+    not f.getUnderlyingType() instanceof ReferenceType and
+    f.getAnAccess().(FieldAccess) = result.getAChild*()
+  ) and
+  (
+    result = e
```
MichaelRFairhurst: Hmm....I think a field of reference type should still be a subobject of itself. So this should be outside of the above condition.

cpp/common/src/codingstandards/cpp/lifetimes/CppObjects.qll
```diff
@@ -136,19 +137,39 @@ abstract class ObjectIdentityBase extends Element {
...
  * Note that this does not hold for `e->x` or `e[x]` where `e` is a pointer.
  */
-private Expr getASubobjectAccessOf(Expr e) {
-  result = e
-  or
-  result.(DotFieldAccess).getQualifier() = getASubobjectAccessOf(e)
+Expr getASubobjectAccessOf(Expr e) {
+  exists(Field f |
+    not f.getUnderlyingType() instanceof ReferenceType and
+    f.getAnAccess().(FieldAccess) = result.getAChild*()
```
MichaelRFairhurst: It looks like using `getAChild*()` finds too many results here.

For example:
```
my_var                         // 1
  .get_a_field_of_non_ref_type // 2
  .get_a_field_of_ref_type     // 3
```

Here, `3` is not a subobject of `1`. However, `2` is a child of `3`, and `2` is a field access of non-ref type, so it satisfies `f`. So this will give the wrong answer here.

I think the correct implementation is simpler:
```
result = e
or
result.(DotFieldAccess).getQualifier = getASubobjectAccessOf(e) and
not result.(DotFieldAccess).getTarget().getUnderlyingType() instanceof ReferenceType
or
....
```

cpp/common/src/codingstandards/cpp/lifetimes/CppObjects.qll
```diff
@@ -136,19 +137,39 @@ abstract class ObjectIdentityBase extends Element {
...
+    or
+    result.(DotFieldAccess).getQualifier() = getASubobjectAccessOf(e)
+  )
   or
   result.(ArrayExpr).getArrayBase() = getASubobjectAccessOf(e) and
   not result.(ArrayExpr).getArrayBase().getUnspecifiedType() instanceof PointerType
 }
 
+/**
+ * gets an access where the pointee is the subobject
```
MichaelRFairhurst: Might be clearer to write with a more detailed comment, which I'm happy to help fill in!

Based on the name, I'd say this predicate takes an expression `e` with pointer type pointer, and finds subobjects of the thing that `e` points to. Doing so is just a matter of dereferencing `e`, and then potentially getting subobject accesses of that result.

The expression `e` may be dereferenced either with `*e`, getting the pointee itself, or with `e->x`, immediately getting a subobject of the pointee. Any subobject access of these expressions (e.g. `(*e).x`, or `e->x.y[0].b...` are subobject accesses of the pointee of expression `e`.

This predicate is especially useful for finding subobject accesses of `*this`.

cpp/common/src/codingstandards/cpp/lifetimes/CppObjects.qll
```diff
@@ -136,19 +137,39 @@ abstract class ObjectIdentityBase extends Element {
...
+ */
+Expr getASubobjectAccessOfPointee(Expr e) {
+  e.getParent() instanceof AddressOfExpr and
```
MichaelRFairhurst: I may have put the wrong code here in my earlier review, and if so I'm sorry!

We don't want to find `&x`, we want to find `*x`, which is a `PointerDereferenceExpr`

cpp/misra/src/rules/RULE-6-8-4/MemberFunctionsRefqualified.ql
```diff
@@ -0,0 +1,119 @@
...
+  MembersReturningObject() {
+    exists(ReturnStmt r, ThisExpr t |
+      r.getEnclosingFunction() = this and
+      (
+        //direct access only
+        r.getAChild() = t
+        or
+        //or one level of indirection
+        exists(PointerDereferenceExpr p |
+          p.getAChild() = t and
```
MichaelRFairhurst: Here we can say

`r.getAChild() = getASubobjectAccessOf(t)`.

That will handle: `return *this`, `return this->x`, `return (*this)[4]`, etc.

We'll need to say `or r.getAChild() = t` to handle `return this`.

cpp/common/src/codingstandards/cpp/lifetimes/CppObjects.qll
```diff
@@ -136,19 +137,39 @@ abstract class ObjectIdentityBase extends Element {
...
+  result = getASubobjectAccessOf(e.getParent())
+  or
+  // the accessed field is a pointer to a subobject
+  e.getParent().(PointerFieldAccess).getTarget().getUnspecifiedType() instanceof PointerType and
+  result = getASubobjectAccessOf(e.getParent())
```
MichaelRFairhurst: Mostly right but subtle difference!

Given `e1->e2`, we want the result to be `e2` when the parameter is `e1`. (or a subobject access of e2).

cpp/misra/src/rules/RULE-6-8-4/MemberFunctionsRefqualified.ql
```diff
@@ -0,0 +1,119 @@
...
+ */
+class MembersReturningSubObject extends MembersReturningObjectOrSubobject {
+  MembersReturningSubObject() {
+    exists(ReturnStmt r, FieldAccess access, Expr e |
+      r.getEnclosingFunction() = this and
+      //direct access only
+      r.getAChild() = e and
+      (
+        //pointer or reference to pointer subobject returned
+        e = getASubobjectAccessOfPointee(access) and
```
MichaelRFairhurst: In this case we don't want to use `getASubobjectAccessOfPointee` on the `access`.

It looks like this case is handling:
```
return my_member;
return my_member.subobject;
return my_member.subobject[n];
// etc
```

So in this case, we can say `e = getASubobjectAccessOf(access)`. But we should also handle `return &my_member` too.

cpp/misra/src/rules/RULE-6-8-4/MemberFunctionsRefqualified.ql
```diff
@@ -0,0 +1,119 @@
...
+        (e.getType() instanceof PointerType or e.getType() instanceof ReferenceType)
```
MichaelRFairhurst: So, in these two branches here, we don't necessarily care if `e` has a pointer type or `e` has a reference type so long as `e = getASubobjectAccessOf(access)`.

The two cases that we need to handle, though, are:
- pointers to subobjects, and
- references to subobjects

So if `e` is a subobject, then:
- returning the address of `e` is not allowed (as in `return &...)`
- returning `e` directly is not allowed if the method returns by reference (as in `int &get_ref() { return e; }`)

I do have to apologize though because there's one more caveat that makes this confusing 😅. I think it's the last thing. We have to not only check that `e = getASubobjectAccessOf(access)`, but we also have to check if `access` is not a reference type. That's just because an access of reference type isn't a subobject in the first place, if that makes sense.

MichaelRFairhurst (commented):
This is really not far off! It's using basically all the right pieces in all the right places, just looks like a couple cases are subtly flipped around I'd say!

cpp/common/src/codingstandards/cpp/lifetimes/CppObjects.qll
knewbury01: ah I see now why it was suggested to be added to the dotfieldaccess condition only, done!

cpp/common/src/codingstandards/cpp/lifetimes/CppObjects.qll
knewbury01: fixed!

cpp/common/src/codingstandards/cpp/lifetimes/CppObjects.qll
knewbury01: when `e` is e1 then  `e.getParent().(PointerFieldAccess)` is e2

so should it be ?
```
result = e.getParent()
or 
result = getASubobjectAccessOf(e.getParent())
```

but since getASubobjectAccessOf(e) = e then we dont actually need the first one?

cpp/misra/src/rules/RULE-6-8-4/MemberFunctionsRefqualified.ql
knewbury01: handled now! this really helped, thanks

MichaelRFairhurst (commented):
Fully approved after one last case is handled!!!

We just need to check in `MembersReturningSubobject` that the `FieldAccess` has no qualifier, or the qualifier is `this`.

MichaelRFairhurst (approved):
🍾 LGTM!

PR #1105, opened 2026-04-07
Add Banned7 package

jeongsoolee09:
## Description

Add rule package `Banned7` that contains Rule 21.6.1.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
 - [X] Queries have been added for the following rules:
     - `RULE-21-6-1`
     - `RULE-21-6-2`

## Release change checklist

  - [ ] Yes
  - [X] No

🚨🚨🚨
  - [ ] Confirmed

  - [x] Confirmed

### Author

 - [X] Have you run the queries on OpenPilot and verified that the performance and results are acceptable?<br />_As a rule of thumb, predicates specific to the query should take no more than 1 minute, and for simple queries be under 10 seconds. If this is not the case, this should be highlighted and agreed in the code review process._

### Reviewer 


cpp/common/src/codingstandards/cpp/DynamicMemory.qll
```diff
@@ -0,0 +1,32 @@
...
+
+import cpp
+
+/**
+ * An `operator new` or `operator new[]` allocation function called by a placement-new expression.
+ *
+ * The operator functions have a `std::size_t` as their first parameter and a
+ * `void*` parameter somewhere in the rest of the parameter list.
+ */
+class PlacementNewOrNewArrayAllocationFunction extends AllocationFunction {
```
MichaelRFairhurst: Can we combine this with cpp/common/src/allocations/CustomOperatorNewDelete ?

cpp/misra/src/rules/RULE-21-6-1/DynamicMemoryShouldNotBeUsed.ql
```diff
@@ -0,0 +1,246 @@
...
+ * and C++ allocation functions (operator new, operator new[]).
+ *
+ * This excludes placement-new operators, as they do not allocate memory themselves.
+ */
+class DirectDynamicMemoryAllocatingFunction extends DynamicMemoryAllocatingFunction {
+  DirectDynamicMemoryAllocatingFunction() {
+    this instanceof AllocationFunction and
+    not this instanceof PlacementNewOrNewArrayAllocationFunction
+    or
+    this.hasGlobalOrStdName("aligned_alloc")
```
MichaelRFairhurst, line 38: Can you add a comment for why aligned_alloc has to be special cased here, or, can you add aligned_alloc to AllocationFunction?

cpp/misra/src/rules/RULE-21-6-1/DynamicMemoryShouldNotBeUsed.ql
```diff
@@ -0,0 +1,246 @@
...
+ */
+abstract class IndirectDynamicMemoryAllocatingFunction extends DynamicMemoryAllocatingFunction { }
+
+/**
+ * A constructor of a standard library container that uses `std::allocator` directly
+ * as template argument or under the hood as the default value of the template argument.
+ * Includes `vector`, `deque`, `list`, `forward_list`, `set`, `map`, `multiset`, `multimap`,
+ * `unordered_set`, `unordered_map`, `unordered_multiset`, `unordered_multimap`, and `valarray`.
+ */
+class AllocatorContainerConstructor extends IndirectDynamicMemoryAllocatingFunction {
```
MichaelRFairhurst, line 56: I think we need to implement this via a very precise and subtle reading of the specified hueristic in the standard:

> any instantiation of a C++ standard library entity having a template argument that is a specialization of `std::allocator` is a violation of this rule

This is actually important because:

```
// Non compliant, because it uses std::allocator<int> which is a specialization of std::allocator
std::vector<int, std::allocator<int>> non_compliant_vec1;

// Non compliant, because the 2nd type parameter is defaulted to std::allocator<T>
// so std::vector<int> and std::vector<int, std::allocator<int>> are the same type.
std::vector<int> non_compliant_vec2;

// Some setup code that uses static storage instead of the heap
char buf[128];
struct nonallocating {
  void *allocate(int size) { return buf; }
  void deallocate(void *p) { }
};

// Compliant: this vector does _not_ use a specialization of std::allocator<T> ! 
std::vector<int, nonallocating> my_nonallocating_vec;
```

The above pattern is I believe pretty reasonable and/or common in the embedded space.

Many of the cases you found here are awesome and will actually still be necessary. For instance, vector and string only use dynamic memory if their `allocator` uses dynamic memory, but `std::ifstream` always allocates and is non compliant by the 2nd part of the rule's hueristic: `... as is any call to a C++ standard library function that may use dynamic memory`.

Happy to talk more and clarify later.

cpp/misra/src/rules/RULE-21-6-2/DynamicMemoryManagedManually.ql
```diff
@@ -17,20 +17,7 @@
...
- * `void*` parameter somewhere in the rest of the parameter list.
- */
-class PlacementNewOrNewArrayAllocationFunction extends AllocationFunction {
-  PlacementNewOrNewArrayAllocationFunction() {
-    this.getName() in ["operator new", "operator new[]"] and
-    this.getParameter(0).getType().resolveTypedefs*() instanceof Size_t and
-    this.getAParameter().getUnderlyingType() instanceof VoidPointerType
-  }
-}
+import codingstandards.cpp.DynamicMemory
```
MichaelRFairhurst, line 20: Nice code reuse!

cpp/misra/test/rules/RULE-21-6-1/test.cpp
```diff
@@ -0,0 +1,352 @@
...
+#include <string>
+#include <string_view>
+#include <thread>
+#include <tuple>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+#include <valarray>
+#include <variant>
+#include <vector>
```
MichaelRFairhurst, line 31: excellent test coverage as well!

MichaelRFairhurst (commented):
Let's explore how complex it may be to implement the std::allocator based logic.

If it's really complex to implement, I think it's fair to say that we don't support non-allocating specializations of allocating types in this rule. In that case, users can simply file the extra deviations when they do that kind of stuff, and we can lower the precision and add it to the implementation_scope.

cpp/misra/src/rules/RULE-21-6-1/DynamicMemoryShouldNotBeUsed.ql
jeongsoolee09: `std::aligned_alloc` nor `aligned_alloc` is an `AllocationFunction`. `AllocationFunction` is standard library class, I'll open a PR in github/codeql.

cpp/common/src/codingstandards/cpp/DynamicMemory.qll
jeongsoolee09: Yes, we can. The blast radius would have Rule 21.6.2 in it though.

cpp/misra/src/rules/RULE-21-6-1/DynamicMemoryShouldNotBeUsed.ql
jeongsoolee09: [PR in github/codeql](https://github.com/github/codeql/pull/21725) merged successfully. Upgrading `codeql/cpp-all` to the next version should remove this handler.

cpp/misra/src/rules/RULE-21-6-1/DynamicMemoryShouldNotBeUsed.ql
jeongsoolee09: My opinion on this is that it is enough to report cases that explicitly passes a specialized `std::allocator<T>` or omits it. It is already a noisy rule due to its scope, so I don't see there's a good reason to find more cases (unless we hit an embedded C++ code that does pass a non-allocating allocator).

cpp/common/src/codingstandards/cpp/DynamicMemory.qll
jeongsoolee09: Addressed in [6bbd610](https://github.com/github/codeql-coding-standards/pull/1105/commits/6bbd610df75fca3afaa6c532d57ec5df10cd34a2).

cpp/misra/src/rules/RULE-21-6-1/DynamicMemoryShouldNotBeUsed.ql
jeongsoolee09: All addressed in the latest revision.

cpp/misra/src/rules/RULE-21-6-1/DynamicMemoryShouldNotBeUsed.ql
MichaelRFairhurst: Awesome! I think just add an explanatory comment here, it will be a while before we can upgrade cpp-all to the newest version.

cpp/misra/test/rules/RULE-21-6-1/DynamicMemoryShouldNotBeUsed.expected
```diff
@@ -0,0 +1,160 @@
...
+| test.cpp:95:23:95:30 | call to vector | Call to 'vector' that uses 'std::allocator<T>'. |
+| test.cpp:96:23:96:25 | call to vector | Call to 'vector' that uses 'std::allocator<T>'. |
+| test.cpp:97:23:97:29 | call to vector | Call to 'vector' that uses 'std::allocator<T>'. |
+| test.cpp:99:19:99:20 | call to deque | Call to 'deque' that uses 'std::allocator<T>'. |
+| test.cpp:100:22:100:29 | call to deque | Call to 'deque' that uses 'std::allocator<T>'. |
+| test.cpp:102:18:102:19 | call to list | Call to 'list' that uses 'std::allocator<T>'. |
+| test.cpp:103:21:103:28 | call to list | Call to 'list' that uses 'std::allocator<T>'. |
+| test.cpp:105:26:105:28 | call to forward_list | Call to 'forward_list' that uses 'std::allocator<T>'. |
+| test.cpp:106:30:106:37 | call to forward_list | Call to 'forward_list' that uses 'std::allocator<T>'. |
+| test.cpp:107:1:107:1 | call to ~vector | Call to '~vector' that uses 'std::allocator<T>'. |
```
MichaelRFairhurst: Consider not reporting destructors.

Reading the rule text, it first says "any instantiation of a C++ Standard Library entity having a template argument that is a specialization of std::allocator is a violation of this rule." I think if the intent was to flag destructors, they would have said "any instantiation or deletion of ..."

Reading more of the rule text, we can definitely justify flagging destructors. For example, after talking about instantiations, the rule says, "as is any call to a C++ Standard Library function that may use dynamic memory." Which could be interpreted to mean destructors, for sure. It also talks about "free" specifically, and how that can occur implicitly.

So there's definitely a case to be made for reporting `~vector()`. And we should definitely definitely flag `delete vec` for some `std::vector<T> *vec`.

That said, I'm not sure RAII implicit deletions will be useful to report. If a user deviates this rule to declare `std::vector<T> vec;` then it's implied that they're OK with that destructor being called. It's also slightly weird because there's no explicit location in the code to deviate.

cpp/misra/src/rules/RULE-21-6-1/DynamicMemoryShouldNotBeUsed.ql
```diff
@@ -0,0 +1,257 @@
...
+ * This class can be divided into big categories:
+ *
+ * 1. A constructor of standard library containers such as `vector`, `deque`, `unordered_set`.
+ * 2. A constructor of standard library strings such as `string`, `wstring` that derives from
+ *    `std::basic_string`.
+ */
+class AllocatorContructor extends IndirectDynamicMemoryAllocatingFunction {
+  AllocatorContructor() {
+    /* Ensure that the constructor accepts a `std::allocator`. */
+    this.getDeclaringType()
```
MichaelRFairhurst, line 70: It looks like destructors are flagged here because it doesn't check `this instanceof Constructor`

MichaelRFairhurst (approved):
Looks great! Two suggestions if you want to implement them.

cpp/misra/src/rules/RULE-21-6-1/DynamicMemoryShouldNotBeUsed.ql
jeongsoolee09: Done in the latest revision.

cpp/misra/test/rules/RULE-21-6-1/DynamicMemoryShouldNotBeUsed.expected
jeongsoolee09: I believe there's a middle ground to be taken here. The call detected above is obviously hidden away from the user because it's the RAII at action and not a manual invocation from the user. I am sure that the rule only applies to the latter and obviously not the former.

The case for reporting such RAII destructors being called can be countered by the argument that report on these calls very likely pollute the results, and these are implied by a more fundamental cause that is the use of `std::vector` that allocates.

So, a manual invocation of the destructor is what we want to report here. I am confident that calls to destructors, although not mentioned directly, count as dynamic memory management considering the intention of the rule.

All in all, the manual destructor calls are what we want to report along the same lines as we report calls to  `free` and `operator delete`.

MichaelRFairhurst: Seems reasonable to me.

Basically, the two cases left over to consider are:

```
vec->~vector(); // explicit call
// and
delete vec;
```

I'm not sure the first one is common enough to be worth tracking and reporting, but we could definitely track it and flag it. And luckily enough, the 2nd one will already be reported.

jeongsoolee09: Addressed in [508020e](https://github.com/github/codeql-coding-standards/pull/1105/commits/508020ebdd9a66bf1488c61176dcd54f8d8a024c).

cpp/misra/src/rules/RULE-21-6-1/DynamicMemoryShouldNotBeUsed.ql
jeongsoolee09: Addressed in [234ca96](https://github.com/github/codeql-coding-standards/pull/1105/commits/234ca9644b0bb5dc3daa2f94049fada0426b2697).

cpp/misra/test/rules/RULE-21-6-1/DynamicMemoryShouldNotBeUsed.expected
jeongsoolee09: Rolled back in [0e85f55](https://github.com/github/codeql-coding-standards/pull/1105/commits/0e85f553447a509e938d2246ed4d68e9ea74ff2c) and added justification.

PR #1112, opened 2026-04-20
Michaelrfairhurst/declarations8 rule 6-2-3 do not duplicate source code

MichaelRFairhurst:
## Description

This rule is somewhat interesting because 6-2-1 already mostly does things that 6-2-3 does. We can't expressly detect ODR violations, so 6-2-1 looks for repeated names across files. However, there's still room to make 6-2-1 stricter and that's what this query does. Essentially, 6-2-3 gives us permission to flag cases that we considered likely false positives in 6-2-1.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - `RULE-6-2-3`
     - _rule number here_

## Release change checklist

  - [x] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 


cpp/misra/test/rules/RULE-6-2-3/test.cpp
```diff
@@ -0,0 +1,52 @@
+#include <cstdint>
+
+inline int16_t global_redefined = 0;  // NON_COMPLIANT[False negative]
+inline int16_t global_unique = 0;     // COMPLIANT
+inline int16_t global_redeclared = 0; // COMPLIANT
+inline void func_redefined() {}       // NON_COMPLIANT
```
knewbury01, line 6: is this actually technically compliant?
the rule says inline entities can be defined once per each translation unit, test and test2 are separate translation units? (as long as the definition is the same... however maybe we cant really check that so we should always flag, is that the logic?)

cpp/misra/test/rules/RULE-6-2-3/test.cpp
```diff
@@ -0,0 +1,52 @@
+#include <cstdint>
+
+inline int16_t global_redefined = 0;  // NON_COMPLIANT[False negative]
```
knewbury01: minor adjustment to NON_COMPLIANT[FALSE_NEGATIVE]

rule_packages/cpp/Declarations8.json
```diff
@@ -0,0 +1,56 @@
...
+          "precision": "very-high",
+          "severity": "error",
+          "short_name": "DuplicateInlineFunctionDefinitions",
+          "tags": [
+            "correctness",
+            "maintainability",
+            "scope/system"
+          ],
+          "implementation_scope": {
+            "description": "This query does not detect duplicated definitions of inline variables.",
```
knewbury01: maybe note here that the rest of duplicated entities (noninline, vars etc) will be found in some other rules (list those ones)?

cpp/misra/src/rules/RULE-6-2-3/DuplicateInlineFunctionDefinitions.ql
```diff
@@ -0,0 +1,44 @@
...
+ *       external/misra/obligation/required
+ */
+
+import cpp
+import codingstandards.cpp.misra
+import codingstandards.cpp.types.Compatible
+
+predicate isInline(FunctionDeclarationEntry d) { d.getDeclaration().isInline() }
+
+predicate interestedInFunctions(FunctionDeclarationEntry f1, FunctionDeclarationEntry f2) {
```
knewbury01, line 33: classic case, I forget how to solve this , but this predicate seems to take a long time on opencv

knewbury01 (commented):
completed partial review, will come back for the rest at a later time, mostly minor comments so far only!

rule_packages/cpp/Declarations8.json
knewbury01: and maybe also elaborate here on which precise case this rule does detect (fairly obvious from the name but maybe only bc I have also read the implementation)

rule_packages/cpp/Declarations8.json
```diff
@@ -0,0 +1,56 @@
...
+            "correctness",
+            "maintainability",
+            "scope/system"
+          ]
+        },
+        {
+          "description": "Defining a type with the same fully qualified name in multiple files increases the risk of ODR violations and undefined behavior.",
+          "kind": "problem",
+          "name": "Duplicate type definitions across files",
+          "precision": "very-high",
```
knewbury01, line 49: this query finds pairs that are sometimes related, in openpilot the first result I see is one typedef and then a second def where the second one is only defined (in a header) ifndef the other one, which makes a lot of sense I think? maybe we can just lower the precision to high for this one as a result?

cpp/misra/src/rules/RULE-6-2-3/DuplicateTypeDefinitions.ql
```diff
@@ -0,0 +1,41 @@
...
+ *       external/misra/obligation/required
+ */
+
+import cpp
+import codingstandards.cpp.misra
+import codingstandards.cpp.Linkage
+
+class UserTypeDefinition extends TypeDeclarationEntry {
+  UserTypeDefinition() {
+    (isDefinition() or getDeclaration() instanceof TypedefType) and
```
knewbury01: ok, I am not sure of this myself but wanted to check

should this rule actually include typedefs at all?
[typedefs can be redeclared ](https://en.cppreference.com/cpp/language/typedef)if they refer to the same type so technically just the presence of 2 with the same name is not an ODR? also , if a typedef is a declaration not a defn, maybe incorrect duplicated typedefs (if the types are not the same) is another rule (not sure?) and wouldnt really be ODR? though I am having a hard time finding a resource on that, other than [this](https://en.cppreference.com/cpp/language/definition) which for types just says enum and class types are relevant for ODR

cpp/misra/src/rules/RULE-6-2-3/DuplicateTypeDefinitions.ql
```diff
@@ -0,0 +1,41 @@
...
+    not getDeclaration() instanceof ClassTemplateSpecialization and
+    not getDeclaration().getNamespace() instanceof WithinAnonymousNamespace
+  }
+
+  UserType getUserType() { result = getDeclaration() }
+}
+
+from UserTypeDefinition t1, UserTypeDefinition t2
+where
+  not isExcluded(t1, Declarations8Package::duplicateTypeDefinitionsQuery()) and
```
knewbury01: I think this should have (maybe , double check, bc this then does cut out dups of things in /usr/include/ headers so not totally sure)
```suggestion
  not isExcluded(t1, Declarations8Package::duplicateTypeDefinitionsQuery()) and
  not isExcluded(t2, Declarations8Package::duplicateTypeDefinitionsQuery()) and
```

knewbury01 (commented):
another partial review - for duplicate type definitions portion

cpp/misra/test/rules/RULE-6-2-3/test.cpp
MichaelRFairhurst: It's a good question. Here's my interpretation, please double check me on this!

Under amplification, it says:

> This rule requires that the source code to implement an entity should only appear once ... If the entity is inline, it can be implemented within a header file ... [and included in] multiple translation units

And rationale:

> inline entities can be defined once for each translation unit, but the definitions shall be identical. This is known as the one-definition rule. ... Requiring that the [definition] only appears once reduces the risk of violating the one-definition rule.

Rule 6-2-1 says that the one-definition-rule shall not be violated. My interpretation is that this rule is intended to be stricter than 6-2-1. So multiple definitions of an inline entity in different header files is disallowed, and if that inline entity is used it multiple translation units, it should be implemented in a single header and included in those translation units rather than copied.

wdyt?

cpp/misra/src/rules/RULE-6-2-3/DuplicateInlineFunctionDefinitions.ql
MichaelRFairhurst: Fixed! Good catch, it didn't have an issue on the dbs I tried. Glad you tried more than I did!

cpp/misra/src/rules/RULE-6-2-3/DuplicateTypeDefinitions.ql
MichaelRFairhurst: I wondered the same thing!

I eventually figured that typedefs can increase ODR violations like so:

```
// a.h
typedef int32_t my_int;
// b.h
typedef int64_t my_int;
// c.h
void f() {
  my_int x = ...;
  ...
}

// d.cpp
#include "a.h"
#include "c.h"

// e.cpp
#include "b.h"
#include "c.h"
```

I'd be happy to remove typedef checking, because that's a pretty rare kind of scenario.

cpp/misra/src/rules/RULE-6-2-3/DuplicateTypeDefinitions.ql
MichaelRFairhurst: good catch, I agree.

rule_packages/cpp/Declarations8.json
MichaelRFairhurst: I elaborated, feedback welcome!

rule_packages/cpp/Declarations8.json
MichaelRFairhurst: This an interesting case. The `#ifndef` pattern here is unusual. It's checking if `sighandler_t` is defined as a preprocessor symbol, not as a typedef...

`sighandler_t` is supposed to be exposed by the `#include <csignal>`. It looks like maybe this is some weird portability hack, or attempted portability hack.

Perhaps some runtimes don't expose `sighandler_t`, and so this file redefines it. In this case, when `sighandler_t` is already defined as a `typedef`, and the types are compatible, then the repeated declaration still compiles. But maybe the author was worried some runtimes might expose `sighandler_t` as a macro, e.g.:

```
typedef void (*__sighandler_t)(int sig);
#define sighandler_t __sighandler_t
```

But even in this case, the `#ifndef` seems unnecessary because expanding `sighandler_t` to `__sighandler_t` would still expand to a valid typedef (`typedef void (*__sighandler_t)(int sig);`).

Asking copilot for thoughts, it says:

> The #ifndef guard is effectively a no-op in every realistic scenario. It doesn't hurt anything, but it doesn't actually protect against anything either. It's cargo-cult defensiveness.

I don't know if I'm missing something special here.

I'd say at the end of the day:
- The `#ifndef` is probably not doing anything -- especially since it's only used for one of the two cases
- The two definitions should probably be merged into one `csignal_compat.h` type file
- If the `#ifndef` hack is required for portability, that's all the more reason to unify the two `sighandler_t` definitions

But, uh, yeah, this case is weird...

cpp/misra/test/rules/RULE-6-2-3/test.cpp
knewbury01: sure, I am happy to leave it with that (stricter) interpretation

cpp/misra/src/rules/RULE-6-2-3/DuplicateTypeDefinitions.ql
knewbury01: hm , yeah I think still removing typedefs would make sense to me (and make a note in the package json for that it doesnt cover them)

right now there are 200 results on open pilot for them, many seem to be fine redeclarations with the same type, and many more are matches of unnamed struct and unnamed struct (for 2 diff typedef structs) which these result categories are not too useful atm without more work to maybe separate them out further

rule_packages/cpp/Declarations8.json
knewbury01: will be handled by the comment below

cpp/misra/src/rules/RULE-6-2-3/DuplicateTypeDefinitions.ql
MichaelRFairhurst: Done!

knewbury01 (approved):
Thanks for addressing all of the comments! lgtm!

PR #1114, opened 2026-04-21
Add "misra_help" query docs generator

data-douser:
Introduces `scripts/generate_rules/misra_help/`, a two-stage pipeline for (mostly) idempotent generation of per-query `.md` help files. It uses MISRA rule text as input and creates (or updates) documentation for `codeql-coding-standards` queries in C and C++.

Initial supported standards:

- MISRA C 2012 / 2023
- MISRA C++ 2023

**Stage 1** — deterministic, [docling](https://github.com/DS4SD/docling)-based extraction and rendering, with a JSON sidecar for downstream consumption.

**Stage 2** — a headless Python driver for the Copilot SDK that rewrites each help file from the JSON sidecar against a fixed Markdown schema, normalized to American English.

See [`scripts/generate_rules/misra_help/README.md`](scripts/generate_rules/misra_help/README.md) for usage, architecture, and operational notes.

## Description

Adds a new internal tooling package under `scripts/generate_rules/misra_help/` that automates generation of per-query Markdown help files for MISRA C/C++ queries. No query files, query metadata, rule packages, shared libraries, tests, `.expected` files, or release artifacts are modified by this PR — it is purely additive tooling (7 new files, ~2.1k lines, all under `scripts/generate_rules/misra_help/`).

The pipeline is split so that the deterministic extraction stage can be re-run cheaply and audited independently of the LLM-driven rewrite stage. The JSON sidecar is the contract between the two stages, which keeps Stage 2 reproducible against a pinned input.

## Change request type

 - [x] Internal documentation
 - [ ] External documentation
 - [ ] Query files (`.ql`, `.qll`, `.qls` or unit tests)

 - [x] No rules added
     - _rule number here_
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No — tooling-only change under `scripts/generate_rules/misra_help/`; no release artifacts, query results, or query performance are affected.

🚨🚨🚨
  - [x] N/A — no `.ql`/`.qll` files modified.

  - [ ] Confirmed

> N/A for this PR — no queries are added or modified. This section is left as-is for the reviewer to confirm.

### Author

### Reviewer 


scripts/generate_rules/misra_help/populate_help.py
```diff
@@ -0,0 +1,361 @@
...
+    return f"{action} {rel} ({len(body)} bytes)"
+
+
+def main() -> int:
+    ap = argparse.ArgumentParser(description=__doc__,
+                                 formatter_class=argparse.RawDescriptionHelpFormatter)
+    ap.add_argument("--standard", required=True, choices=SUPPORTED_STANDARDS,
+                    help="MISRA standard to populate (the source language is "
+                         "derived from this)")
+    ap.add_argument("--query-repo", type=Path, default=DEFAULT_QUERY_REPO,
```
MichaelRFairhurst, line 255: IMO some of this stuff should be deleted as YAGNI.

I think it's totally fine to either assume that the working directory is the project root, or to find the project root via relative path to `__FILE__`. We already have other scripts that assume the help repo can be found via `../codeql-coding-standards-help`.

Given the size of this PR, I'd rather not add too many bells and whistles

scripts/generate_rules/misra_help/populate_help.py
```diff
@@ -0,0 +1,361 @@
...
+# per-licensee suffix (e.g. "MISRA-C-2023-XXXXXX.pdf"). We do not hard-code the
+# filename. The PDF location is resolved in this order:
+#
+#   1. --pdf CLI flag
+#   2. environment variable named in PDF_ENV_VARS for the standard
+#   3. a glob of PDF_FILE_GLOBS within --help-repo
+#
+# If none of those resolve to exactly one file, we abort with a clear message.
+PDF_ENV_VARS = {
+    "MISRA-C-2023":   "MISRA_C_PDF",
```
MichaelRFairhurst, line 54: So just a clarification here. We should not differentiate C-2023 and C-2012 at all.

Every rule we have is both a part of MISRA C 2012, and a part of MISRA C 2023, there isn't an actual distinction. MISRA C 2012 with all amendments included = MISRA C 2023

scripts/generate_rules/misra_help/populate_help.py
```diff
@@ -0,0 +1,361 @@
...
+    "MISRA-C++-2023": "MISRA_CPP_PDF",
+}
+PDF_FILE_GLOBS = {
+    "MISRA-C-2023":   ["MISRA-C-2023*.pdf", "MISRA-C-2012*.pdf"],
+    "MISRA-C-2012":   ["MISRA-C-2023*.pdf", "MISRA-C-2012*.pdf"],
+    "MISRA-C++-2023": ["MISRA-CPP-2023*.pdf", "MISRA-C++-2023*.pdf"],
+}
+
+RULE_DIR_RE = re.compile(r"^(?:RULE|DIR)-\d+(?:-\d+){1,2}$")
+QL_NAME_RE = re.compile(r"@name\s+(?:RULE|DIR)-\d+(?:-\d+){1,2}:\s+(?P<title>.+?)\s*$")
```
MichaelRFairhurst, line 65: This is reproducing some prior art and needs to be consolidated.

We have normalized titles etc already in `rule_packages/*.json`. The script `scripts/generate_rules/generate_package_files.py` already takes the parsed rule_package data which is organized per rule and per query, and that's what's used to fill in [the existing help template](https://github.com/github/codeql-coding-standards/blob/main/scripts/generate_rules/templates/misra-c%2B%2B-2023-help.md.template).

What's especially important is that some rule_package.json entries have an `implementation_scope` property ([see here ](https://github.com/github/codeql-coding-standards/blob/main/rule_packages/cpp/Macros.json#L48)) that's added to the query help. This is critical, because it is the only part of our query help that isn't a direct copy of the misra text, but rather describes expected FPs and FNs to the user.

scripts/generate_rules/misra_help/populate_help.py
```diff
@@ -0,0 +1,361 @@
...
+    return None
+
+
+def resolve_pdf(standard: str, cli_pdf: Path | None, help_repo: Path) -> Path:
+    """Locate the licensed PDF for a standard. Raises with a helpful message."""
+    if cli_pdf is not None:
+        if not cli_pdf.is_file():
+            raise SystemExit(f"error: --pdf {cli_pdf} does not exist")
+        return cli_pdf
+    env_var = PDF_ENV_VARS[standard]
```
MichaelRFairhurst, line 179: This is another thing we should cut via YAGNI -- no need to support setting the pdf path as an environment variable, just creates more code to have to maintain

scripts/generate_rules/misra_help/populate_help.py
```diff
@@ -0,0 +1,361 @@
...
+    env_val = os.environ.get(env_var)
+    if env_val:
+        p = Path(env_val).expanduser()
+        if not p.is_file():
+            raise SystemExit(
+                f"error: ${env_var} is set to {p} which does not exist")
+        return p
+    matches: list[Path] = []
+    for pattern in PDF_FILE_GLOBS[standard]:
```
MichaelRFairhurst, line 188: this is also unnecessary magic

scripts/generate_rules/misra_help/populate_help.py
```diff
@@ -0,0 +1,361 @@
...
+        edition), and
+      - docling rule-anchor detection failures that leave a rule with
+        an empty or garbled title.
+
+    Status is one of:
+        wrote-new        file did not exist, was written
+        wrote-changed    file existed, content differs, was rewritten
+        unchanged        file existed and render matches byte-for-byte
+        skip-existing    file existed and --no-overwrite was passed
+        title-mismatch   rule title was not verifiable from any query;
```
MichaelRFairhurst: I would also be inclined to say that some of this is borderline excessive. We can get similar data from git. I'd rather this script be simpler to maintain than have specialized output for each of these cases.

scripts/generate_rules/misra_help/populate_help.py
```diff
@@ -0,0 +1,361 @@
...
+                    help="path to codeql-coding-standards-help repo")
+    ap.add_argument("--pdf", type=Path, default=None,
+                    help="path to the licensed MISRA PDF (overrides env var "
+                         "and help-repo glob)")
+    ap.add_argument("--cache-dir", type=Path,
+                    default=Path("/tmp/misra-pdf-probe/repo-cache"),
+                    help="docling JSON cache dir (deterministic across runs)")
+    ap.add_argument("--rule", action="append", default=[],
+                    help="restrict to specific RULE-X-Y[-Z] (repeatable)")
+    ap.add_argument("--no-overwrite", action="store_true",
```
MichaelRFairhurst, line 267: I'd flip this. By default, script doesn't overwrite, unless you pass `--overwrite`

scripts/generate_rules/misra_help/populate_help.py
```diff
@@ -0,0 +1,361 @@
...
+                         "from the PDF-extracted title (by default we skip "
+                         "such files to avoid overwriting correct content "
+                         "with content from a renumbered rule or a broken "
+                         "PDF anchor)")
+    ap.add_argument("--dry-run", action="store_true",
+                    help="report what would be written without writing")
+    args = ap.parse_args()
+
+    pdf = resolve_pdf(args.standard, args.pdf, args.help_repo)
+    args.cache_dir.mkdir(parents=True, exist_ok=True)
```
MichaelRFairhurst, line 282: It might be best to hardcode `scripts/generate_rules/misra_help/cache` as a path. Then we can add that to our `.gitignore`, and all should be swell

scripts/generate_rules/misra_help/README.md
```diff
@@ -0,0 +1,348 @@
...
+into your checkout of the help repo (or pass `--pdf` on every
+invocation).
+
+Expected local layout (the licensee suffix on each PDF filename will
+vary):
+
+```text
+codeql-coding-standards-help/
+├── .gitignore                       # ignores MISRA-*.pdf
+├── MISRA-C-2023-<licensee>.pdf      # NOT in git; you place it here
```
MichaelRFairhurst: why does this require the licensee?

scripts/generate_rules/misra_help/README.md
```diff
@@ -0,0 +1,348 @@
+# MISRA help-file populator
```
MichaelRFairhurst, line 1: This file should be simplified down to maybe ~60 lines of md at most

scripts/generate_rules/misra_help/extract_rules.py
```diff
@@ -0,0 +1,666 @@
...
+    ap = argparse.ArgumentParser()
+    ap.add_argument("pdf")
+    ap.add_argument("--standard", required=True, choices=list(STD_DISPLAY))
+    ap.add_argument("--lang", default=None,
+                    help="override the language used to render code fences "
+                         "(default: derived from --standard)")
+    ap.add_argument("--cache-dir", default="/tmp/misra-pdf-probe")
+    ap.add_argument("--out-dir", default="/tmp/misra-pdf-probe/extracted")
+    ap.add_argument("--rule", action="append", help="only emit these rule IDs")
+    ap.add_argument("--check-csv", default="/Users/data-douser/Git/github/codeql-coding-standards/rules.csv",
```
MichaelRFairhurst: local path

scripts/generate_rules/misra_help/extract_rules.py
```diff
@@ -0,0 +1,666 @@
...
+    args = ap.parse_args()
+    rules = extract_rules(Path(args.pdf), args.standard, Path(args.cache_dir))
+    print(f"Extracted {len(rules)} rules from {args.pdf}")
+    out = Path(args.out_dir)
+    out.mkdir(parents=True, exist_ok=True)
+    (out / f"{args.standard}.json").write_text(
+        json.dumps([to_dict(r) for r in rules], indent=2),
+        encoding="utf-8",
+    )
+    if args.check_csv:
```
MichaelRFairhurst: This should probably be its own script. It'd be a good utility overall to check for queries with missing help files, across all standards

scripts/generate_rules/misra_help/extract_rules.py
```diff
@@ -0,0 +1,666 @@
...
+# ----------------------------------------------------------------------------
+
+STD_DISPLAY = {
+    "MISRA-C-2023": "MISRA C 2023",
+    "MISRA-C-2012": "MISRA C 2012",
+    "MISRA-C++-2023": "MISRA C++ 2023",
+}
+
+
+def render_help(rule: Rule, lang: str = "c") -> str:
```
MichaelRFairhurst, line 600: Should use [the templates](https://github.com/github/codeql-coding-standards/blob/main/scripts/generate_rules/templates/misra-c-2012-help.md.template)

scripts/generate_rules/misra_help/extract_rules.py
```diff
@@ -0,0 +1,666 @@
...
+        "",
+        "## References",
+        "",
+        f"* {STD_DISPLAY.get(rule.standard, rule.standard)}: {rule.raw_id}: {rule.title}",
+        "",
+    ]
+    return "\n".join(parts)
+
+
+def to_dict(rule: Rule) -> dict:
```
MichaelRFairhurst, line 655: unnecessary

scripts/generate_rules/misra_help/extract_rules.py
```diff
@@ -0,0 +1,666 @@
...
+# line breaks are collapsed to spaces, so examples would render as one
+# long line. We cannot losslessly recover the original line breaks without
+# re-reading layout boxes, but for C/C++ examples we can insert
+# statement-level breaks at the obvious boundaries: `;`, `{`, `}`, and
+# before `//` line comments. This is a deterministic, purely textual
+# transform — no parsing or formatting — and keeps the output readable.
+
+_CODE_FORMAT_STEPS = [
+    # Pull "// ..." comments onto their own line.
+    (re.compile(r"\s+//"),          "\n//"),
```
MichaelRFairhurst: Most examples have trailing comments that we don't want to move to their own line

```
int x; // non-compliant
```

scripts/generate_rules/misra_help/extract_rules.py
```diff
@@ -0,0 +1,666 @@
...
+from __future__ import annotations
+import json
+import re
+from dataclasses import dataclass, field, asdict
+from pathlib import Path
+
+# ----------------------------------------------------------------------------
+# Wordlist-based ligature repair (deterministic)
+# ----------------------------------------------------------------------------
+_WORDLIST_PATHS = ["/usr/share/dict/words", "/usr/dict/words"]
```
MichaelRFairhurst, line 26: This won't be consistent across machines, which makes me slightly anxious

scripts/generate_rules/misra_help/extract_rules.py
```diff
@@ -0,0 +1,666 @@
...
+
+_LIGS = ("fi", "fl", "ff", "ffi", "ffl")
+# Suspect glyphs observed in the MISRA C++ PDF's font CMap:
+#   digits 0-9, capital `C`, caret `^`, percent `%`, and capital `A`
+#   all appear where a genuine ligature (fi/fl/ff/ffi/ffl) was
+#   originally rendered. The wordlist check in `repair_ligatures`
+#   prevents mis-substitution on legitimate CamelCase identifiers
+#   containing `A` or `C`.
+_SUSPECT_GLYPHS = set("0123456789CA^%")
+_SUSPECT_TOKEN_RE = re.compile(r"[A-Za-z0-9CA\^%]*[0-9CA\^%][A-Za-z0-9CA\^%]*")
```
MichaelRFairhurst: Some of these regexes seem unnecessarily complex and hard to read.

For example, the `CA`s in both `A-Za-z0-9CA`'s are redundant.

MichaelRFairhurst: This may also have catastrophic backtracking, I believe, that could be fixed by not allowing `\^%` in the first `[]` group. There's no reason, AFAICT, why the first group would match `\^%`

scripts/generate_rules/misra_help/extract_rules.py
```diff
@@ -0,0 +1,666 @@
...
+        # Only attempt repairs on tokens that already contain letters;
+        # pure-digit tokens like "4" or "10" must be left alone even
+        # though they start or end with a suspect glyph.
+        if not any(c.isalpha() for c in tok):
+            return tok
+        low = tok.lower()
+        if low.isalpha() and low in words:
+            return tok
+        out = tok
+        for _ in range(4):  # at most a few rounds for ffl etc.
```
MichaelRFairhurst: I can see why this would use a loop to a fixed point, generally speaking, but I don't think it will actually be used.

For example, if we take a word with one broken ligature, such as `efficiency` being `e^ciency`, then the candidate "ffi" produces "efficiency" which is a valid word match, and there's no need to repeat.

Alternatively, if we take a real word with multiple broken ligatures, such as "fluffy" which was corrupted to "%u^y", then none of the possible replacements ('fi' 'fl', 'ff', ...) at either of the positions ('%' or '^') will produce a valid word, and the loop will break because `changed = False`.

The only time we'd get value from repeating this loop is if the mangled tokens have lucky placement of `C` or `A`, and those tokens happen to produce a valid word...like, "fluffy/CuAy" which could produce "Cuffy" on the first iteration. But A. cuffy isn't a word, and B, if it was a word, how would we know that it wasn't the intended word?

scripts/generate_rules/misra_help/rewrite_help.py
```diff
@@ -0,0 +1,425 @@
...
+    return text
+
+
+# ---------------------------------------------------------------------------
+# Main rewrite loop
+# ---------------------------------------------------------------------------
+
+
+def load_cache(help_repo: Path, standard: str) -> dict[str, Any]:
+    cache_path = help_repo / ".misra-rule-cache" / f"{standard}.json"
```
MichaelRFairhurst: maybe make a `cache.py` that exposes helpers for finding & reading from the cache, so that we aren't repeating this kind of logic across each step with different default paths etc

scripts/generate_rules/misra_help/rewrite_help.py
```diff
@@ -0,0 +1,425 @@
...
+    p = argparse.ArgumentParser(description=__doc__.split("\n\n", 1)[0])
+    p.add_argument("--standard", required=True, choices=SUPPORTED_STANDARDS)
+    p.add_argument("--help-repo", type=Path, default=DEFAULT_HELP_REPO,
+                   help=f"Path to codeql-coding-standards-help (default: {DEFAULT_HELP_REPO}).")
+    p.add_argument("--rule", action="append", default=[],
+                   help="Restrict to specific rule IDs (e.g. RULE-6-7-1). Repeatable.")
+    p.add_argument("--model", default=DEFAULT_MODEL,
+                   help=f"Copilot model id. Default: {DEFAULT_MODEL}. "
+                        f"Known good: {', '.join(MODEL_FALLBACKS)}.")
+    p.add_argument("--no-overwrite", action="store_true",
```
MichaelRFairhurst, line 453: Again, I'd probably prefer default behavior doesn't overwrite and it requires `--overwrite` to do so.

scripts/generate_rules/misra_help/rewrite_help.py
```diff
@@ -0,0 +1,425 @@
...
+import sys
+import time
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Iterable
+
+import requests
+
+
+SUPPORTED_STANDARDS = ("MISRA-C-2012", "MISRA-C-2023", "MISRA-C++-2023")
```
MichaelRFairhurst, line 45: again, there aren't really two C standards

scripts/generate_rules/misra_help/rewrite_help.py
```diff
@@ -0,0 +1,425 @@
...
+@dataclass
+class CopilotToken:
+    token: str
+    expires_at: int  # unix seconds
+
+    def near_expiry(self, slack_seconds: int = 300) -> bool:
+        return time.time() + slack_seconds >= self.expires_at
+
+
+def fetch_copilot_token(oauth_token: str) -> CopilotToken:
```
MichaelRFairhurst, line 134: Should we be using the copilot API, or copilot CLI?

scripts/generate_rules/misra_help/rewrite_help.py
```diff
@@ -0,0 +1,425 @@
...
+                raise RuntimeError(
+                    f"Copilot chat failed: HTTP {resp.status_code} {resp.text[:500]}"
+                )
+            data = resp.json()
+            return data["choices"][0]["message"]["content"]
+        raise RuntimeError(f"Copilot chat failed after retries: {last_err}")
+
+
+# ---------------------------------------------------------------------------
+# Prompt construction (mirrors codeql-coding-standards-agent/src/rewriteHelp.ts)
```
MichaelRFairhurst, line 227: Note that this is a "nice to have."

The "must have" is that we have query help. As a first pass, this should be a word for word match to the MISRA documents.

scripts/generate_rules/misra_help/rewrite_help.py
```diff
@@ -0,0 +1,425 @@
...
+        "4. Then optional \"### Amplification\" and \"### Rationale\" sections, each as well-formed prose paragraphs. Collapse multi-space kerning runs (e.g. \"If   any   element\" -> \"If any element\"). Use straight quotes.",
+        "5. Then an optional \"### Exception\" section. If the source provides multiple exceptions, render them as a numbered list (1., 2., 3.) -- never as bullets.",
+        "6. Then an optional \"## Example\" section. Code goes inside a fenced block with the language tag (```c or ```cpp). REFORMAT the code so each statement is on its own line, braces are placed idiomatically, and `/* Compliant */` / `/* Non-compliant */` comments stay on the same line as the statement they annotate. If the docling extraction interleaved code and prose paragraphs (example_layout), preserve that interleaving with the prose between fenced code blocks.",
+        "7. Then optional \"## See also\" listing referenced rules.",
+        "8. End with these two sections verbatim, with the rule id and the short rule statement substituted in:",
+        "   \"## Implementation notes\"",
+        "   \"\"",
+        "   \"None\"",
```
MichaelRFairhurst: This is provided by the `implementation_scope` field in our rule_packages json files, and should not be `None`

MichaelRFairhurst (changes requested):

scripts/generate_rules/misra_help/populate_help.py
data-douser: Latest commits should address the `implementation_scope` requirement ^

scripts/generate_rules/misra_help/rewrite_help.py
data-douser: Either way it uses the Copilot API somewhere.

MichaelRFairhurst:
Closing in favor of committing to a different repo.

PR #1121, opened 2026-04-29
Narrowly focused implementation of RULE 15-0-1

MichaelRFairhurst:
## Description

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - `RULE-15-0-1`
     - _rule number here_

## Release change checklist

  - [ ] Yes
  - [x] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


cpp/misra/test/rules/RULE-15-0-1/test.cpp
```diff
@@ -275,28 +275,30 @@ class CopyEnabledCustomizedDtorCompliant1 { // COMPLIANT: copy-enabled (2)
...
 };
 
 // copy-assignable class with both move operations customized
 class CopyAssignableCustomizedDtorCompliant1 { // COMPLIANT
 public:
-  DEFINE_ALL_SPECIAL_MEMBERS(CopyAssignableCustomizedDtorCompliant1, DEFAULTED,
-                             CUSTOMIZED, CUSTOMIZED, CUSTOMIZED, CUSTOMIZED)
+  COPY_CTOR(CopyAssignableCustomizedDtorCompliant1) CUSTOMIZED;
+  MOVE_CTOR(CopyAssignableCustomizedDtorCompliant1) CUSTOMIZED;
+  // No move constructor declared
```
jeongsoolee09: This can now be deleted.

cpp/misra/test/rules/RULE-15-0-1/test.cpp
```diff
@@ -305,63 +307,60 @@ class CopyAssignableCustomizedDtorCompliant2 { // COMPLIANT
...
-  COPY_CTOR(CopyAssignableCustomizedDtorNonCompliant4)
-  CUSTOMIZED
+  COPY_CTOR(CopyAssignableCustomizedDtorNonCompliant4) CUSTOMIZED;
   // Move constructor is not declared
-  COPY_ASSIGN(CopyAssignableCustomizedDtorNonCompliant4)
-  CUSTOMIZED MOVE_ASSIGN(CopyAssignableCustomizedDtorNonCompliant4) CUSTOMIZED
-      DTOR(CopyAssignableCustomizedDtorNonCompliant4) CUSTOMIZED
+  COPY_ASSIGN(CopyAssignableCustomizedDtorNonCompliant4) CUSTOMIZED;
+  MOVE_ASSIGN(CopyAssignableCustomizedDtorNonCompliant4) CUSTOMIZED;
+  DTOR(CopyAssignableCustomizedDtorNonCompliant4) CUSTOMIZED
```
jeongsoolee09: Missing a trailing `;`.

cpp/misra/src/rules/RULE-15-0-1/AnalyzableClass.qll
```diff
@@ -0,0 +1,112 @@
...
+  else result = c.getAConstructor().(MoveConstructor)
+}
+
+Operator getMoveAssign(Class c) {
+  if
+    not exists(MoveAssignmentOperator mc | mc = c.getAMemberFunction() and isUserDeclared(mc)) and
+    implicitMoveIsSuppressed(c)
+  then result = c.getAMemberFunction().(CopyAssignmentOperator)
+  else result = c.getAMemberFunction().(MoveAssignmentOperator)
+}
```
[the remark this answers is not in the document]
jeongsoolee09: I think it's a good idea to add a motivating example to the QLDoc, respectively.

``` C++
class OnlyCopyCtor {
public:
  OnlyCopyCtor(const OnlyCopyCtor &) = default;
};

static_assert(std::is_copy_constructible_v<OnlyCopyCtor>); // Succeeds
static_assert(std::is_move_constructible_v<OnlyCopyCtor>); // Also succeeds
```

``` C++
class OnlyCopyAssign {
public:
  OnlyCopyAssign& operator=(const OnlyCopyAssign &) = default;
};

static_assert(std::is_copy_assignable_v<OnlyCopyAssign>); // Succeeds
static_assert(std::is_move_assignable_v<OnlyCopyAssign>); // Also succeeds
```

cpp/misra/test/rules/RULE-15-0-1/test.cpp
jeongsoolee09: (I mean the line 290.)

cpp/misra/test/rules/RULE-15-0-1/test.cpp
jeongsoolee09: (I mean the line 343.)

cpp/misra/src/rules/RULE-15-0-1/AnalyzableClass.qll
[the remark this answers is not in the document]
MichaelRFairhurst: Nice suggestion, done!

cpp/misra/test/rules/RULE-15-0-1/test.cpp
MichaelRFairhurst: Oops, I deleted the whole class!

But I think that's ok, it's covered by:

```cpp
class CustomizedCtorsCompliant { // COMPLIANT: copy-enabled (1)
public:
  DEFINE_ALL_SPECIAL_MEMBERS(CustomizedCtorsCompliant, CUSTOMIZED, CUSTOMIZED,
                             DELETED, DELETED, CUSTOMIZED)
};
```

(both are customized constructors, deleted assignments, and customized dtor)

cpp/misra/src/rules/RULE-15-0-1/AnalyzableClass.qll
```diff
@@ -0,0 +1,258 @@
...
+ * Note that `T(T&&) = default` and `T(T&&) = delete` are both user declared. This is not to be
+ * confused with "user defined."
+ */
+private predicate isUserDeclared(MemberFunction f) { not f.isCompilerGenerated() }
+
+/**
+ * Holds if the implicit move constructor or move assignment operator of the class `c` will not be
+ * declared.
+ *
+ * See [class.copy]/8 and [class.copy]
```
jeongsoolee09:
```suggestion
 * See [class.copy.ctor]/8 and [class.copy]
```

cpp/misra/src/rules/RULE-15-0-1/AnalyzableClass.qll
```diff
@@ -0,0 +1,258 @@
...
+ */
+private predicate implicitMoveIsSuppressed(Class c) {
+  isUserDeclared(c.getAConstructor().(CopyConstructor))
+  or
+  isUserDeclared(c.getAMemberFunction().(CopyAssignmentOperator))
+  or
+  isUserDeclared(c.getDestructor())
```
jeongsoolee09, line 62: It may be missing a branch.
```suggestion
  isUserDeclared(c.getDestructor())
  or
  isUserDeclared(c.getAMemberFunction().(MoveAssignmentOperator))
```

cpp/misra/src/rules/RULE-15-0-1/ImproperlyProvidedSpecialMemberFunctions.ql
```diff
@@ -0,0 +1,150 @@
...
+        specialMemberName(f), ", "
+      ) and
+    reason = "has customized " + concatenated + ", but does not customize the destructor."
+  )
+}
+
+private predicate undeclaredMoveException(AnalyzableClass c) {
+  // A copy-enabled class may have an undeclared move constructor.
+  isCopyEnabled(c) and
+  not c.moveAssignable() and
```
jeongsoolee09: Why is this condition needed?

cpp/misra/src/rules/RULE-15-0-1/ImproperlyProvidedSpecialMemberFunctions.ql
```diff
@@ -0,0 +1,150 @@
...
+  not c.declaresMoveConstructor()
+  or
+  // A copy-assignable class may leave both move operations undeclared.
+  isCopyEnabled(c) and
+  c.moveAssignable() and
```
jeongsoolee09: Shouldn't this be `c.copyAssignable()` instead?

cpp/misra/src/rules/RULE-15-0-1/ImproperlyProvidedSpecialMemberFunctions.ql
```diff
@@ -0,0 +1,150 @@
...
+    or
+    c.copyConstructible() and
+    not c.isCustomized(TCopyConstructor()) and
+    reason = "has customized the destructor, but does not customize the copy constructor."
+    or
+    c.copyAssignable() and
+    not c.isCustomized(TCopyAssignmentOperator()) and
+    reason = "has customized the destructor, but does not customize the copy assignment operator."
+  )
+}
```
jeongsoolee09, line 182: It was a bit hard for me to follow the logic of this requirement check. Since these come in the form of
```
isCopyEnabled(class) -> {some_condition}
```
wouldn't it be simpler to check it using by taking the negation of the above?
```
isCopyEnabled(class) and ~{some_condition}
```
For example, the last requirement translates to
``` codeql
c.copyAssignable() implies (
  c.isCustomized(TCopyAssignmentOperator()) and
  (
    c.isCustomized(TMoveConstructor()) and c.isCustomized(TMoveAssignmentOperator()) or
    not isUserDeclared(getMoveConstructor(c)) and not isUserDeclared(getMoveAssign(c))
  )
)
```
So the check becomes:
``` codeql
c.copyAssignable() and not (
  c.isCustomized(TCopyAssignmentOperator()) and
  (
    c.isCustomized(TMoveConstructor()) and c.isCustomized(TMoveAssignmentOperator()) or
    not isUserDeclared(getMoveConstructor(c)) and not isUserDeclared(getMoveAssign(c))
  )
)
```
Rewriting above gives an arguably more readable version:
``` codeql
c.copyAssignable() and (
  /* 1. The copy assignment operator is not customized. */
  not c.isCustomized(TCopyAssignmentOperator()) or
  (
    /* 2. Any one (or both) of the move operations is not customized. */
    not c.isCustomized(TMoveConstructor() or not c.isCustomized(TMoveAssignmentOperator()) and
    /* 3. Any one (or both) of the move operations are user-declared. */
    (isUserDeclared(getMoveConstructor(c)) or isUserDeclared(getMoveAssign(c)))
  )
)
```

cpp/misra/test/rules/RULE-15-0-1/test.cpp
```diff
@@ -273,7 +273,7 @@ class CopyEnabledCustomizedDtorCompliant1 { // COMPLIANT: copy-enabled (2)
                              CUSTOMIZED, CUSTOMIZED, CUSTOMIZED, CUSTOMIZED)
 };
 
-class CopyEnabledCustomizedDtorCompliant2 { // NON-COMPLIANT: copy-enabled (2)
+class CopyEnabledCustomizedDtorCompliant2 { // COMPLIANT: copy-enabled (2)
```
jeongsoolee09, line 276: Thanks for correcting the labels! Phew!

cpp/misra/test/rules/RULE-15-0-1/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+  int y;
+};
+
+struct TrivialClass { // NON_COMPLIANT - audit result
+  int x;
+  int y;
+  COPY_CTOR(TrivialClass) = default;
+};
+
+class NonTrivialClass { // NON_COMPLIANT - audit result
```
jeongsoolee09, line 429: Would there be value to adding a variety of this class that has a copy constructor?

``` C++
class NonTrivialClass {
  int x;
  int y;
public:
  COPY_CTOR(TrivialClass) = default;
  ~NonTrivialClass() { x = 1; }
};
```

jeongsoolee09 (changes requested):
Some thoughts, overall looks amazing!! Thank you so much for tackling this to an already great extent!

cpp/misra/src/rules/RULE-15-0-1/ImproperlyProvidedSpecialMemberFunctions.ql
MichaelRFairhurst: This (and the next comment) are attempting to check if the class is copy-assignable vs copy-enabled.

If a class is copyEnabled, we can check for the existence of either assignment operator to distinguish copy-enabled from copy-assignable, because a copy-assignable class will have both and a copy-enabled class will have neither.

It took me too long to remember that that's why I wrote it. I'll write a helper predicate that makes this clearer.

cpp/misra/src/rules/RULE-15-0-1/ImproperlyProvidedSpecialMemberFunctions.ql
MichaelRFairhurst: If I'm understanding correctly, I think this is simpler.

The bullet points 2 and 3 can essentially be rephrased as: 
- in the presence of a customized destructor
- if it's copy constructible the copy constructor must be customized, and
- if it's copy assignable the copy assignment operator must be customized, and
- if it's move constructible, the move constructor must be customized, and
- if it's move assignable, the the move assignment operator must be customized

The only exceptions are:
- copy enabled, not copy-assignable classes may leave the move constructor undeclared
- copy assignable classes may leave both move ctor & assign undeclared

This matches my implementation above.

We can prove each item:

> if it's copy constructible, the copy constructor must be customized

If it is copy constructible, it can only be copy enabled or copy assignable. Copy enabled classes must have a customized copy constructor, and copy assignable classes must have a customized copy constructor.

Erg, all copy constructible classes must have a customized copy constructor.

Inversely, if it must have a customized copy constructor, it must be either copy constructible or copy assignable -- the other three categories have no restrictions on the copy constructor.

> if it's copy assignable, the copy assignment operator must be customized

This one is a direct quote: "if it is copy-assignable it shall also have a customized copy assignment operator." No other categories place any requirements on the copy assignment operator.

> if it's move constructible, the move constructor must be customized

If it's move constructible, it can belong to any category except unmovable. If it is move-only or move-assignable, it must have a customized move constructor (bullet 2).

If it is copy-enabled, it may be undeclared (exception 1) or customized. If it is copy-assignable, both move operation may be undeclared (exception 2) or the the move constructor must be customized along with the copy constructor

Ergo, all move constructible classes must fulfill either exception 1 or exception 2 or have a customized move constructor.

Inversely, the unmovable category places no requirements on the move constructor.

> if it's move assignable, the move assignment operator must be customized

if its move assignable, it may be move-only. Move only and assignable classes must have a customized move assignment operator.

If it is copy-enabled, the move assignment operator may be undeclared if the move constructor is (exception 2). Otherwise, it has to be customized.

Ergo, all move assignable classes must fulfill exception 2 or have a customized move assignment operator.

Inversely, a class may be:
- unmovable, in which case there are no requirements on the move assignment operator, or
- move-only but not move assignable, in which case there are no requirements on the move assignment operator, or
- copy-enabled but not assignable, in which case there are no requirements on the move assignment operator.

*brief pause*

...To me, this is just much simpler overall. Not only is easier to remember that "x-operatable requires x operation is customized."

But it also makes pragmatic sense. A customized destructor indicates that you're managing a resource. Copying and moving a class requires care, during both construction and during assignment.

```
class Owner {
  Resource *resource;
  Owner() : resource(get_a_resource()) {}
  ~Owner() {
    release(resource);
  }
}
```

In the above code, the default copy operations will copy the handle, and the resource will be freed twice (once by the original, once by the copy). The default move operations will move the handle, which will typically have copy semantics.

It isn't enough to customize just the move constructor, or just the move assignment operator. Every available operation must be either customized or deleted....with one exception! If you leave the move operations undeclared, then moving will call your copy operations, so its ok to leave move operations undeclared and instead customize your copy operations.

MichaelRFairhurst: I think you're proposing something closer to this, right?

This is more verbose, but its absolutely possible to make the case that it has better error messages. It is also a pretty clear 1:1 mapping to the misra spec.

```
predicate violatesCustomizedDestructorRequirements(AnalyzableClass c, string reason) {
  c.isCustomized(TDestructor()) and
  (
    isMoveOnly(c) and
    (
      not c.isCustomized(TMoveConstructor()) and
      reason =
        "is move-only with a customized the destructor, but does not customize the move constructor."
      or
      c.moveAssignable() and
      not c.isCustomized(TMoveAssignmentOperator()) and
      reason =
        "is move-only and move-assignable with a customized the destructor, but does not customize the move assignment operator."
    )
    or
    isCopyEnabled(c) and
    (
      not c.isCustomized(TCopyConstructor()) and
      reason =
        "is copy-enabled with a customized the destructor, but does not customize the copy constructor."
      or
      not c.copyAssignable() and
      c.declaresMoveConstructor() and
      not c.isCustomized(TMoveConstructor()) and
      reason =
        "is copy-enabled with a customized the destructor, but declares a move constructor and does not customize it."
    )
    or
    c.copyAssignable() and
    (
      /* 1. The copy assignment operator is not customized. */
      not c.isCustomized(TCopyAssignmentOperator()) and
      reason =
        "is copy-assignable with a customized destructor but doesn't have a customized copy assignment operator"
      or
      /* 2. Only one of the move operations is customized. */
      c.declaresMoveConstructor() and
      not c.declaresMoveAssignmentOperator() and
      reason =
        "is copy assignable with a customized destructor, but declaring a move constructor implies that the move assignment operator should also be declared"
      or
      not c.declaresMoveConstructor() and
      c.declaresMoveAssignmentOperator() and
      reason =
        "is copy assignable with a customized destructor, but declaring a move assignment operator implies that the move constructor should also be declared"
      or
      /* 3. A move operation is declared but not customized. */
      c.declaresMoveConstructor() and
      not c.isCustomized(TMoveConstructor()) and
      reason =
        "has a customized destructor and declares a move constructor but does not customize it"
      or
      c.declaresMoveAssignmentOperator() and
      not c.isCustomized(TMoveAssignmentOperator()) and
      reason =
        "has a customized destructor and declares a move assignment operator but does not customize it"
    )
  )
}
```

The above version passes our current tests (just with different alert messages).

We shouldn't keep the old version unless we can clarify it. I don't know if its enough, but we could add a comment like this:
```
// If you follow the logic of each case in the rule specification, you'll find that all classes in a valid
// category that have a customized destructor, must customize every "x-operation" when the
// class is "x-operatable."
//
// This is logic makes perfect sense. A customized destructor indicates the class manages a
// resource, and such classes must customize every supported operation to prevent unwanted
// sharing, such as double frees.
//
// The only exception is that copy-enabled classes may sometimes omit move declarations. This
// also makes sense: when these operations aren't declared, the corresponding copy operation
// will be used instead.
```

Hopefully this comment:
- explains that this is intentionally formulated differently than the rule text
- explains the value of this formulation over the spec's formulation
- makes it easier for a reader verify that the code does what it intends to do
- makes the code easier to maintain

MichaelRFairhurst: Oh, one last thought, and I don't really know which method this supports.

The original/current version will report errors for cases that are not one of the valid categories (not unmovable, or move-only, or copy-enabled), while the rewrite I posted above will not.

Usually this is a bad thing. We don't want to give two alerts when we could give one.

In this case, you could argue the reverse, that a class with a customized destructor and an otherwise defaulted operation should always get a warning. If a user intentionally writes a class that doesn't fit one of the categories (unmovable, move-only, copy-enabled), they might ignore that alert, and then they'll never get a warning that they forgot to customize their move assignment constructor.

Interesting to consider. Maybe not an issue in practice. On qlx, these two versions give the exact same results.

cpp/misra/test/rules/RULE-15-0-1/test.cpp
MichaelRFairhurst: Done in 61a4dd9!

cpp/misra/src/rules/RULE-15-0-1/ImproperlyProvidedSpecialMemberFunctions.ql
MichaelRFairhurst: Ah, I see.

Yes, you're right. I forgot that `copy-assignable` was specifically declared to mean a class where `std::is_copy_assignable_v<T>` is true -- I thought it referred specifically to the assignable row of `copy-enabled`. (Which is only partly true).

Fixed in 61a4dd9!

cpp/misra/src/rules/RULE-15-0-1/ImproperlyProvidedSpecialMemberFunctions.ql
MichaelRFairhurst: Fixed in 61a4dd9!

cpp/misra/src/rules/RULE-15-0-1/AnalyzableClass.qll
MichaelRFairhurst: Fixed in 61a4dd9!

cpp/misra/src/rules/RULE-15-0-1/ImproperlyProvidedSpecialMemberFunctions.ql
```diff
@@ -0,0 +1,149 @@
...
+  // A copy-enabled class may have an undeclared move constructor.
+  isCopyEnabled(c) and
+  not c.copyAssignable() and
+  not c.declaresMoveConstructor()
+  or
+  // A copy-assignable class may leave both move operations undeclared.
+  c.copyAssignable() and
+  not c.declaresMoveConstructor() and
+  not c.declaresMoveAssignmentOperator()
+}
```
jeongsoolee09, line 94: Very, very small nitpick because it's about readability. I think you meant this, making the disjunction connected to `isCopyEnabled(c)` with `and`:
``` codeql
private predicate undeclaredMoveException(AnalyzableClass c) {
  isCopyEnabled(c) and (
    // A copy-enabled but non-copy-assignable class may have an undeclared
    // move constructor.
    not c.copyAssignable() and
    not c.declaresMoveConstructor()
    or
    // A copy-assignable class may leave both move operations undeclared.
    c.copyAssignable() and
    not c.declaresMoveConstructor() and
    not c.declaresMoveAssignmentOperator()
  )
}
```
but then an if-then-else pattern emerges:
``` codeql
private predicate undeclaredMoveException2(AnalyzableClass c) {
  isCopyEnabled(c) and
  (
    if c.copyAssignable()
    then
      /* A copy-assignable class may leave both move operations undeclared. */
      not c.declaresMoveConstructor() and
      not c.declaresMoveAssignmentOperator()
    else
      /*
       * A copy-enabled but non-copy-assignable class may only have an
       * undeclared move constructor.
       */

      not c.declaresMoveConstructor()
  )
}
```

jeongsoolee09 (approved):
Thank you!

PR #1169, opened 2026-08-04
Fix RULE-0-2-4/A0-1-3 false positives for private functions

### LINKED ISSUE #1168: `RULE-0-2-4 / A0-1-3`: false positives for friend-accessed members, direct static calls, and pure virtual private methods
### Affected rules
 - RULE-0-2-4
 - A0-1-3

### Description

The `cpp/misra/unused-limited-visibility-function` (RULE-0-2-4) and `cpp/autosar/unused-local-function` (A0-1-3) queries report false positives for private member functions that are genuinely used. We found 57 false-positive findings out of 58 total on the `eclipse-score/communication` codebase.

The false positives fall into three distinct categories.

---

### Category 1: Private methods called through `friend class` access

The query does not reliably resolve call sites that go through friend-class access.

#### Example
```cpp
class SharedMemoryResource {
    friend class SharedMemoryFactoryImpl;
  private:
    static cpp::expected<std::shared_ptr<SharedMemoryResource>, os::Error> Create(...) noexcept;
    static cpp::expected<std::shared_ptr<SharedMemoryResource>, os::Error> Open(...) noexcept;
    uid_t getOwnerUid() const noexcept;
};
```
`Create`, `Open`, and `getOwnerUid` are all called from `SharedMemoryFactoryImpl` (a declared friend), but the query flags them as unused.

---

### Category 2: Private template helpers / constructors only called via direct static dispatch

The query uses `getTarget(call)` from `DynamicCallGraph`, which resolves virtual dispatch. For non-virtual private methods or private template helpers called directly, the dynamic target may not be resolved, so the call goes unrecognised.

**Example 1 – private delegating constructor:**
```cpp
template <typename FieldType, typename... Tags>
class ProxyFieldImpl {
  public:
    ProxyFieldImpl(ProxyBase& proxy_base, std::string_view field_name)
        : ProxyFieldImpl{field_name,
                         MakeSetMethodDispatchIfEnabled(proxy_base, field_name),  // called here
                         MakeGetMethodDispatchIfEnabled(proxy_base, field_name)}  // called here
    {}
  private:
    static std::unique_ptr<ProxyMethod<SetMethodSignature<FieldType>>>
    MakeSetMethodDispatchIfEnabled(ProxyBase&, std::string_view);  // flagged as unused

    static std::unique_ptr<ProxyMethod<GetMethodSignature<FieldType>>>
    MakeGetMethodDispatchIfEnabled(ProxyBase&, std::string_view);  // flagged as unused
};
```

**Example 2 – singleton accessor pattern:**
```cpp
template <typename SampleType>
class ProxyEventBindingFactory {
  public:
    static auto Create(...) { return instance().Create(...); }  // calls instance()
  private:
    static IProxyEventBindingFactory<SampleType>& instance() noexcept;  // flagged as unused
};
```

---

### Category 3: Pure virtual private member declarations

Pure virtual functions (`= 0`) are interface contracts that derived classes must override. They have no definition body and cannot be "called" directly. The query incorrectly includes them as candidates.

**Example:**
```cpp
class ManagedMemoryResource {
  private:
    virtual const void* getEndAddress() const noexcept = 0;            // flagged
    virtual const MemoryResourceProxy* getMemoryResourceProxy() noexcept = 0;  // flagged
};

class SkeletonFieldBase {
  private:
    virtual Result<void> DoDeferredUpdate() noexcept = 0;  // flagged
};

class IMessagePassingService {
  private:
    virtual void UnregisterMethodCallHandler(QualityType, ProxyMethodInstanceIdentifier) = 0;  // flagged
};
```

---

## Expected behavior

The query should only flag private / limited-visibility functions that are genuinely never called. It should **not** flag:

- Private members invoked from friend classes.
- Private helper functions that appear as direct static call targets.
- Pure virtual member declarations (`= 0`).

---

### Affected queries

- `cpp/misra/unused-limited-visibility-function` (RULE-0-2-4)
- `cpp/autosar/unused-local-function` (A0-1-3)

Both share `cpp/common/src/codingstandards/cpp/rules/unusedlocalfunction/UnusedLocalFunction.qll`.

---

### Environment

- CodeQL bundle: `codeql-bundle-v2.21.4`
- Coding standards pack: `2.61.0`
- Downstream repo: `eclipse-score/communication`

castler:
## Description

Two categories of false positive are addressed:

1. **Pure virtual private member functions** — RULE-0-2-4 and A0-1-3
   require that private member functions be *used*.  A pure virtual
   function (`= 0`) is an interface contract that must be overridden
   by derived classes; it has no body of its own and cannot be
   "called" directly.  Excluding `PureVirtualFunction` from the
   `LocalFunction` class prevents spurious reports on these.

2. **Private members of never-instantiated class templates** — when a
   class template is never instantiated with a concrete type anywhere
   in the analyzed compilation units, Clang never elaborates a body for
   its member functions, so `Call`/`FunctionCall` targets within that
   pattern's own text cannot be resolved by
   `DynamicCallGraph::getTarget()` or `VirtualDispatch`, even for calls
   between sibling members of the very same class (e.g. a public entry
   point calling a private helper). This is common for CRTP-style wrappers
   that is only ever instantiated by
   downstream consumers outside of the analyzed codebase. The new
   `hasNoVisibleInstantiation(fn)` predicate conservatively treats such
   private members as "used" (out of scope for this analysis) rather
   than reporting them as dead code — but only when *no* sibling member
   of the same class-template pattern has any instantiation either, so
   genuinely dead private helpers in class templates that *are*
   instantiated elsewhere are still correctly reported.

Fixes #1168

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - RULE-0-2-4
     - A0-1-3

## Release change checklist

  - [X] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 

### DIFF
--- a/change_notes/2026-08-03-unused-local-function-false-positives.md
+++ b/change_notes/2026-08-03-unused-local-function-false-positives.md
@@ -1,0 +1,16 @@
+- `A0-1-3`, `RULE-0-2-4` - `UnusedLocalFunction.ql`:
+  - Fixed false positives for private pure virtual functions used through the non-virtual
+    interface (NVI) idiom. `DynamicCallGraph::getTarget()` resolves a virtual call to the
+    implementations that may actually run; a pure virtual function has no body, so it is
+    never a viable dispatch target and was reported as unused even when a sibling member
+    called it. A call is now also counted when the function is the statically named
+    callee. Pure virtual functions that are genuinely never called and never overridden
+    are still reported.
+  - Excluded private member functions of class templates that are never concretely
+    instantiated anywhere in the database (and where no sibling member of the same
+    class-template pattern is instantiated either). Clang never elaborates a body for the
+    members of such patterns, so calls between sibling members of the same
+    never-instantiated class (e.g. a public entry point calling a private helper) cannot be
+    resolved by the call graph. This is common for generic "plumbing" library code
+    (CRTP-style wrappers, etc.) that is only ever
+    instantiated by downstream consumers outside of the analyzed codebase.
--- a/cpp/common/src/codingstandards/cpp/rules/unusedlocalfunction/UnusedLocalFunction.qll
+++ b/cpp/common/src/codingstandards/cpp/rules/unusedlocalfunction/UnusedLocalFunction.qll
@@ -20,9 +20,82 @@ predicate overloadedFunctionIsCalled(Function unusedFunction) {
   exists(Function f | f = unusedFunction.getAnOverload() and f = getTarget(_))
 }
 
+/**
+ * Holds if `fn` is the target of some call, either statically or according to the
+ * dynamic call graph.
+ *
+ * `DynamicCallGraph::getTarget()` resolves a virtual call to the functions that may
+ * actually run, i.e. the overriding implementations. A pure virtual function has no
+ * body, so it is never a viable dispatch target and is therefore *never* returned by
+ * `getTarget()` -- even when it is unambiguously named by a call, as in the
+ * non-virtual interface (NVI) idiom where a public member calls a private pure
+ * virtual. The additional static `FunctionCall.getTarget()` disjunct recovers exactly
+ * that case: the callee as written in the source.
+ */
+predicate functionIsCalled(Function fn) {
+  fn = getTarget(_)
+  or
+  // The statically named callee, which the dynamic call graph drops for calls that
+  // dispatch to an override (notably pure virtual functions, which have no body).
+  exists(FunctionCall fc | fc.getTarget() = fn)
+}
+
 /** Checks if a Function's address was taken. */
 predicate addressBeenTaken(Function unusedFunction) {
   exists(FunctionAccess fa | fa.getTarget() = unusedFunction)
+}
+
+/**
+ * Holds if some member of the class `c` has at least one concrete instantiation anywhere in the
+ * database.
+ *
+ * If this holds for the declaring type of a member function `fn`, the class template is genuinely
+ * "alive" (used with a concrete type somewhere), and the fact that `fn` itself was never
+ * instantiated is real evidence that it is unused: for a member function to lack a concrete
+ * instantiation while sibling members do have one, it must never have been called from any of
+ * those sibling bodies.
+ *
+ * `pragma[noinline]` keeps this a standalone relation of arity one. Inlined into the caller, the
+ * join orderer loses the fact that `c` is functionally determined and materialises the full
+ * (member, sibling) cross product per class before projecting it away, which is quadratic in the
+ * size of the largest class.
+ */
+pragma[noinline]
+private predicate classHasAnyInstantiatedMember(Class c) {
+  exists(Function sibling, Function siblingInstantiation |
+    sibling.getDeclaringType() = c and
+    siblingInstantiation.isConstructedFrom(sibling)
+  )
+}
+
+/**
+ * Holds if `fn` is a function from an uninstantiated template for which no concrete
+ * instantiation exists anywhere in the database, and no other member of the same
+ * class-template pattern is instantiated either.
+ *
+ * When a class template is never instantiated with a concrete type in the analyzed
+ * compilation units, Clang never elaborates a body for its member functions, so
+ * `Call`/`FunctionCall` targets within that pattern's own text cannot be resolved by
+ * `DynamicCallGraph::getTarget()` or `VirtualDispatch`, even for calls between sibling members
+ * of the very same class (e.g. a constructor calling a private helper). This is common for
+ * generic "plumbing" library code (CRTP-style wrappers, etc.)
+ * that is only ever instantiated by downstream consumers outside of this codebase. In that
+ * situation we have no visibility at all into the call graph, so we conservatively treat the
+ * function as "used" (out of scope for this analysis) rather than report it as dead code.
+ *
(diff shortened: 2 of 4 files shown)


MichaelRFairhurst:
Mostly looks good! I just wanted to say I couldn't reproduce this issue:

```
1. **Static call targets not recognised** — the original check used
   only `getTarget(call)` from DynamicCallGraph, which resolves
   virtual dispatch.  For private (non-virtual) member functions and
   private static functions that are called directly, CodeQL sometimes
   cannot resolve a dynamic target and the function was incorrectly
   reported as unused.  The new `functionIsCalled(fn)` predicate
   combines the dynamic-graph result with a direct
   `FunctionCall.getTarget()` lookup so that statically-dispatched
   calls are always counted as a use.
 ```
 
and I couldn't get these `// COMPLIANT` test cases to fail when I commented out the change.

I tried to get the eclipse/communication project database to confirm the difference myself, but I ran into issues there that I'll have to figure out. It looked somewhat like the database I got was maybe built with the wrong CodeQL version, though I looked through the project and saw it looked like you're using the correct version everywhere, so I'm not totally sure why I was seeing issues on my side.

I also don't see exactly how the `FunctionCall::getTarget` will fix the issues with statics that you're seeing. If you look at `VirtualDispatch::getAViableTarget` you'll see that for non-virtual calls, it already uses `Call::getTarget()`, which `FunctionCall::getTarget()` overrides. The only difference I'm seeing on first look here is that `DynamicCallGraph::getTarget1()` requires the result of `VirtualDisaptch::getAViaibleTarget` to have a definition.

Thank you for the PR!

**Comment by castler** [comment]
> I tried to get the eclipse/communication project database to confirm the difference myself, but I ran into issues there that I'll have to figure out. It looked somewhat like the database I got was maybe built with the wrong CodeQL version, though I looked through the project and saw it looked like you're using the correct version everywhere, so I'm not totally sure why I was seeing issues on my side.

To be frank, we used the wrong CodeQL version for some amount of time. Maybe you had a CodeQL database from that? But since around a month (https://github.com/eclipse-score/communication/pull/757/changes#diff-6136fc12446089c3db7360e923203dd114b6a1466252e71667c6791c20fe6bdcL440) we should use the correct one.

Regarding your finding - I have to apologize - I think you are right. 
I will review this today or tomorrow and push a revised version

**Comment by MichaelRFairhurst** [comment]
Thanks @castler, that would plausibly explain it. I'm definitely not against merging, since these fixes are not harmful, if we find that the mystery continues. 👍

**Comment by castler** [comment]
> Thanks @castler, that would plausibly explain it. I'm definitely not against merging, since these fixes are not harmful, if we find that the mystery continues. 👍

@MichaelRFairhurst  I tried to revisit this topic from ground. I think the changes for the pure-virtual interface are valid, but I am not to sure about: https://github.com/github/codeql-coding-standards/pull/1169/changes#diff-67e20819af17d4736a107e5c7147cf6c52cd27479b579c9a96a26adc004feb39R186-R187 producing false negatives? But I also have no clue how I could test this right now.

**Comment by MichaelRFairhurst** [comment]
@castler this looks great! Approved assuming tests pass.

I agree with you that this new analysis could have new FNs. Since templates are generally undecidable `if constexpr (isPrime<someConstexprInt>()) { only_use_of_private_method(); }` I think that's totally acceptable.

It may be worth explicitly stating the new FP case in the change note, but I think it's also implied by what you have already. So I'm going to say that's your call, I know you've already invested quite a lot of time!

🍾 fantastic work!

**Comment by MichaelRFairhurst** [comment]
@castler one last note.

Unfortunately, the QL engine is producing a bad join order for your predicate `classPatternHasAnyInstantiatedMember`:

```
[2026-08-21 12:01:01] (18s) Tuple counts for UnusedLocalFunction::classPatternHasAnyInstantiatedMember/1#05a71235/1@2ef58fmr after 1.5s:
                      428538     ~6%       {2} r1 = JOIN `cached_ResolveFunction::isFunction/1#9226b83f` WITH `Declaration::Declaration.getDeclaringType/0#dispred#6d415a62` ON FIRST 1 OUTPUT Rhs.1, Lhs.0 'fn'
                      1584498569 ~6%       {2}    | JOIN WITH `Declaration::Declaration.getDeclaringType/0#dispred#6d415a62_10#join_rhs` ON FIRST 1 OUTPUT Rhs.1, Lhs.1 'fn'
                      224313     ~200%     {1}    | JOIN WITH `project#Function::Function.isConstructedFrom/1#dispred#2edd856f#2` ON FIRST 1 OUTPUT Lhs.1 'fn'
```

_(that's an intermediate result of 1.5b tuples after the first join)_

This isn't caused by anything problematic in your design at all, just a confused join orderer.

It looks like an improvement is:
```
pragma[noinline]
private predicate classHasAnyInstantiatedMember(Class c) {
  exists(Function sibling, Function siblingInstantiation |
    sibling.getDeclaringType() = c and
    siblingInstantiation.isConstructedFrom(sibling)
  )
}
```

and a corresponding `not classHasAnyInstantiatedMember(fn.getDeclaringType())`.

MichaelRFairhurst (approved):

MichaelRFairhurst (approved):

MichaelRFairhurst:
Thanks again @castler !

PR #1171, opened 2026-08-04
Fix RULE-5-10-1 false positives on compiler-predefined function identifiers

### LINKED ISSUE #1170: `RULE-5-10-1` reports false positives on compiler-predefined function identifiers (`__func__`, `__FUNCTION__`, `__PRETTY_FUNCTION__`)
### Affected rules
 - RULE-5-10-1

### Description

The query flags the compiler-predefined function identifiers `__func__`, `__FUNCTION__` and `__PRETTY_FUNCTION__` for "contains double underscores" / "starts with underscore". These are **not user-defined
identifiers** — `__func__` is mandated by the C++ standard ([dcl.fct.def.general]) and __FUNCTION__`/`__PRETTY_FUNCTION__` are GCC/Clang extensions — so RULE-5-10-1 (which is scoped to *user-defined* identifiers) should not report them.

### Example

The compiler synthesizes one such `static const char __PRETTY_FUNCTION__[] = "..."` inside every function that
references it. These references very often come from assert-style macros (e.g. an `ASSERT_MESSAGE(...)` macro that embeds `__PRETTY_FUNCTION__`), so the synthesized variable is attributed to each **macro invocation site** in user
code. The result is one false positive per macro usage.

e.g. https://github.com/eclipse-score/baselibs/blob/ebe13d853e87ce067507212ff736af14b7941f6d/score/language/futurecpp/include/score/private/assert/assert_impl.hpp#L50 

### Impact

In score/communication this causes ~79% of all RULE-5-10-1 findings (228)

castler:
## Description

RULE-5-10-1 (poorly-formed identifier) flagged the compiler-predefined
function identifiers __func__, __FUNCTION__ and __PRETTY_FUNCTION__ for
containing double underscores / leading underscores. These are not
user-defined identifiers: __func__ is mandated by the C++ standard and
__FUNCTION__/__PRETTY_FUNCTION__ are GCC/Clang extensions, synthesized
once per enclosing function. They frequently surface via assert-style
macros, producing one finding per macro invocation site (on this
codebase, 228 findings, ~79% of the rule's results).

The shared IdentifierIntroduction library already excludes
`variable.isCompilerGenerated()`, but the extractor does not mark these
predefined variables as compiler generated, so they leak through.

Exclude them in VariableDeclarationEntryIdentifier using a
name-independent structural signal: they are `static` local variables
with no definition (their only declaration entry is located at a use
site), whereas a genuine `static` local always has an in-source
definition. This deliberately avoids hard-coding the names, so a user
who really declares such an identifier (on a compiler that does not
predefine it) still has a definition and is therefore still flagged.

Fixes #1170

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - RULE-5-10-1

## Release change checklist

  - [X] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 


MichaelRFairhurst (approved):
LGTM, thank you @castler !

PR #1178, opened 2026-08-21
Fix RULE-7-0-4 false positives for non-built-in operators

### LINKED ISSUE #1177: `RULE-7-0-4`:  false positives on overloaded `operator<<`/`operator>>` in uninstantiated template bodies
### Affected rules
 - RULE-7-0-4 ( cpp/misra/inappropriate-bitwise-or-shift-operands )  

### Description

 InappropriateBitwiseOrShiftOperands.ql  reports operations that do not use the built-in bitwise or shift operators. The overwhelming majority of these are uses of the stream insertion and extraction operators, e.g.  stream << value . 
                                                                 
On our codebase ( eclipse-score/communication ) this accounts for 72 of 76 reported violations, i.e. ~95% of the results for this rule are false positives. 

Typical messages:
```                                                                                                                                                                                                         
   Shift operator '<<' requires unsigned left operand, but has type 'basic_ostream<char, char_traits<char>> &'.                                                                                          
   Shift operator '<<' requires unsigned left operand, but has type 'unknown'.                                                                                                                           
   Shift operator '<<' requires unsigned right operand, but has type 'auto &&'.                                                                                                                          
   Shift operator '<<' requires unsigned right operand, but has type 'const char[22]'.                       
```
### Example

```cpp                                                                                                                                                                                                         
template<typename ServiceIdType>
std::string ToHexString(const ServiceIdType service_id) {
std::stringstream stream;
// FALSE POSITIVE: reported twice, for the left and the right operand
stream << std::setfill('0') << std::setw(8) << std::hex << service_id;
return stream.str();
}                
```

castler:
## Description
`InappropriateBitwiseOrShiftOperands.ql` reported operations that do not use the built-in bitwise or shift operators, most notably the stream insertion and extraction operators, e.g. `stream << value`.

Resolved calls to a user provided operator are modelled as `FunctionCall`s and therefore never reach this query. The reported operations were instead the syntactic `LShiftExpr`/`RShiftExpr` nodes emitted for uninstantiated template bodies, where the operands are either unresolved (`unknown`, `auto &&`) or are the non-dependent result of an already resolved overload, such as `std::basic_ostream<char, std::char_traits<char>> &`.

The built-in bitwise and shift operators are only applicable to integral and unscoped enumeration operands, so an operation with an operand of any other type does not use them. The query now requires both operands to be of such a type before reporting a violation.

This does not introduce false negatives for dependent operands, because the template instantiations, in which the operand types are known, are still reported.

Fixes #1177

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - RULE-7-0-4

## Release change checklist

  - [X] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 

### DIFF
--- a/change_notes/2026-08-20-fix-fp-rule-7-0-4-non-numeric-operand-types.md
+++ b/change_notes/2026-08-20-fix-fp-rule-7-0-4-non-numeric-operand-types.md
@@ -1,0 +1,11 @@
+- `RULE-7-0-4` - `InappropriateBitwiseOrShiftOperands.ql`:
+  - Fixes #1177 - the rule no longer reports operands whose type is not a MISRA numeric type. The
+    operand checks used `not isUnsignedType(operandType)`, which is vacuously true for every type
+    that has no MISRA numeric type at all, such as class types and the unresolved dependent types
+    of uninstantiated template bodies. As a result the rule reported operations that do not use
+    the built-in operators, most notably the stream insertion and extraction operators. The checks
+    now use `isSignedType(operandType)` instead.
+  - Operands of character type, of a non-standard integral type, and of an unscoped enumeration
+    type without a fixed underlying type are consequently no longer reported, because none of them
+    has a MISRA numeric type. Unscoped enumerations without a fixed underlying type are covered by
+    `RULE-10-2-3`.
--- a/cpp/misra/src/rules/RULE-7-0-4/InappropriateBitwiseOrShiftOperands.ql
+++ b/cpp/misra/src/rules/RULE-7-0-4/InappropriateBitwiseOrShiftOperands.ql
@@ -63,7 +63,7 @@ where
     |
       x = op.getLeftOperand() and
       operandType = op.getLeftOperand().getExplicitlyConverted().getType() and
-      not MisraCpp23BuiltInTypes::isUnsignedType(operandType) and
+      MisraCpp23BuiltInTypes::isSignedType(operandType) and
       message =
         "Bitwise operator '" + op.getOperator() +
           "' requires unsigned numeric operands, but the left operand has type '" + operandType +
@@ -71,7 +71,7 @@ where
       or
       x = op.getRightOperand() and
       operandType = op.getRightOperand().getExplicitlyConverted().getType() and
-      not MisraCpp23BuiltInTypes::isUnsignedType(operandType) and
+      MisraCpp23BuiltInTypes::isSignedType(operandType) and
       message =
         "Bitwise operator '" + op.getOperator() +
           "' requires unsigned numeric operands, but the right operand has type '" + operandType +
@@ -82,7 +82,7 @@ where
     exists(ComplementExpr comp, Type opType |
       x = comp.getOperand() and
       opType = comp.getOperand().getExplicitlyConverted().getType() and
-      not MisraCpp23BuiltInTypes::isUnsignedType(opType) and
+      MisraCpp23BuiltInTypes::isSignedType(opType) and
       message =
         "Bit complement operator '~' requires unsigned operand, but has type '" + opType + "'."
     )
@@ -91,7 +91,7 @@ where
     exists(BinaryShiftOpOrAssignOp shift, Type leftType |
       x = shift.getLeftOperand() and
       leftType = shift.getLeftOperand().getExplicitlyConverted().getType() and
-      not MisraCpp23BuiltInTypes::isUnsignedType(leftType) and
+      MisraCpp23BuiltInTypes::isSignedType(leftType) and
       not isSignedConstantLeftShiftException(shift) and
       message =
         "Shift operator '" + shift.getOperator() +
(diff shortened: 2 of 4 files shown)


MichaelRFairhurst:
Thank you @castler! I remember seeing a lot of these, but I had filed a deviation and then forgot. Thank you for the fixes!

Great analysis of the FP, nicely done, very subtle and good catch!

In general, we often use `not thing.isFromUninstantiatedTemplate(_)`, and I think that would make sense for this. Technically this may mean we don't report warnings in uninstantiated template code where we otherwise could have...but uninstantiated template code analysis is generally tricky to get right, as you've seen here.

Adding `not x.isFromUninstantiatedTemplate(_)` seems to fix most FP cases. The remaining FPs from here I see in my databases are:
- `decltype(declval<std::ostream &>() << std::declval<T const &>())` for some reason still is a BinaryShiftExpr
- `unsigned __int128` shifts
- unscoped enums without explicit types e.g. `enum e { ... }`
- char32_t

Your version solves those decltype FPs, which is great. Note for unscoped enums, the standard says to treat them as their explicit underlying type if specified. For unscoped enums with no explicit type they should be ignored and are covered by rule 10.2.3.

For `isBuiltInOperandType`, I think we should prefer to use `expr instanceof MisraCpp23BuiltInTypes::MisraBuiltInType`, which should handle all the various cases well related to enums, typedefs, decltypes, references, specifiers, etc.

This makes me wonder what our query missed, since we're already using that module ourselves... It looks like the general mistake we made in this query is using `not ...::isUnsignedType(type)` instead of `...::isSignedType(type)` (without the `not`). For something like a user defined class, `...::is[Un]SignedType(classType)` is false, and therefore `not ...::is[Un]SignedType(classType)` is always true, and that's a wider net than we want to cast.

So the query could then look something like:

```
from Expr x, string message
where
  not isExcluded(x, ...) and
  not x.isFromUninstantiatedTemplate(_) and
  x instanceof MisraCpp23BuiltinTypes::MisraBuiltInType and
```

This would only leaves FPs for:
- char32_t
- unsigned __int128

_The char32_t FP is caused by a bug in `MisraCpp23BuiltInTypes::NumericTypes::getSignedness()` -- we should have also added `getSignedness()` to `CharacterType`._

Alternatively, instead of `x instanceof MisraCpp23BuiltinType::MisraBuiltInType`, you could update everything to:

```
   ...
   MisraCpp23BiltInTypes::isSignedType(opType) and
   ...
```

This would fix the `char32_t` case by excluding them from analysis...which is sort of incidentally correct because the `char_x_t`s are always unsigned.

If my analysis seems right to you, and I haven't missed anything, then feel free to solve with whatever approach resonates with you!

castler:
> Thank you @castler! I remember seeing a lot of these, but I had filed a deviation and then forgot. Thank you for the fixes!
> 
> Great analysis of the FP, nicely done, very subtle and good catch!
> 
> In general, we often use `not thing.isFromUninstantiatedTemplate(_)`, and I think that would make sense for this. Technically this may mean we don't report warnings in uninstantiated template code where we otherwise could have...but uninstantiated template code analysis is generally tricky to get right, as you've seen here.
> 
> Adding `not x.isFromUninstantiatedTemplate(_)` seems to fix most FP cases. The remaining FPs from here I see in my databases are:
> 
> * `decltype(declval<std::ostream &>() << std::declval<T const &>())` for some reason still is a BinaryShiftExpr
> * `unsigned __int128` shifts
> * unscoped enums without explicit types e.g. `enum e { ... }`
> * char32_t
> 
> Your version solves those decltype FPs, which is great. Note for unscoped enums, the standard says to treat them as their explicit underlying type if specified. For unscoped enums with no explicit type they should be ignored and are covered by rule 10.2.3.
> 
> For `isBuiltInOperandType`, I think we should prefer to use `expr instanceof MisraCpp23BuiltInTypes::MisraBuiltInType`, which should handle all the various cases well related to enums, typedefs, decltypes, references, specifiers, etc.
> 
> This makes me wonder what our query missed, since we're already using that module ourselves... It looks like the general mistake we made in this query is using `not ...::isUnsignedType(type)` instead of `...::isSignedType(type)` (without the `not`). For something like a user defined class, `...::is[Un]SignedType(classType)` is false, and therefore `not ...::is[Un]SignedType(classType)` is always true, and that's a wider net than we want to cast.
> 
> So the query could then look something like:
> 
> ```
> from Expr x, string message
> where
>   not isExcluded(x, ...) and
>   not x.isFromUninstantiatedTemplate(_) and
>   x instanceof MisraCpp23BuiltinTypes::MisraBuiltInType and
> ```
> 
> This would only leaves FPs for:
> 
> * char32_t
> * unsigned __int128
> 
> _The char32_t FP is caused by a bug in `MisraCpp23BuiltInTypes::NumericTypes::getSignedness()` -- we should have also added `getSignedness()` to `CharacterType`._
> 
> Alternatively, instead of `x instanceof MisraCpp23BuiltinType::MisraBuiltInType`, you could update everything to:
> 
> ```
>    ...
>    MisraCpp23BiltInTypes::isSignedType(opType) and
>    ...
> ```
> 
> This would fix the `char32_t` case by excluding them from analysis...which is sort of incidentally correct because the `char_x_t`s are always unsigned.
> 
> If my analysis seems right to you, and I haven't missed anything, then feel free to solve with whatever approach resonates with you!

Great feedback - you are right that `isSignedType(type)` will solve that way more elegant. Then we even do not need to check for the uninitiated template, as the regression tests have shown.

I tried to adjust accordingly.

MichaelRFairhurst (approved):

PR #1180, opened 2026-08-24
RULE-6-9-1: fix false positives for redeclarations that use the same type alias

castler:
@MichaelRFairhurst I want to apologize for this. I did not really understood the underlying reason for the false positives so I just throwed copilot at it which generated this PR. To be honest, I do still not really understand whats going on - but I wanted to raise this anyhow, since we have quite some findings with it.

## Description

`RULE-6-9-1` ("The same type aliases shall be used in all declarations of the
same entity") reports a large number of false positives on real code. Every one
of the reported entities actually uses *identical* alias spellings in all its
declarations, e.g. a function prototype in a header and its definition in a
`.cpp` both returning `element_fq_id&`, `LoggingCallback`, `Result<...>`, etc.

### Root cause

The query fires on:

```ql
t.getATypeNameUse() = decl1 and
not t.getATypeNameUse() = decl2
```

`TypedefType.getATypeNameUse()` is documented as a **conservative / incomplete**
relation ("not necessarily *all* type name uses"; it omits uses on prototypes
and around template instantiations). Worse, in whole-program extraction it is
**inconsistent across translation units**: the *same* header line produces
multiple `DeclarationEntry` objects (one per including TU) and
`getATypeNameUse()` associates the alias with some copies but not others.

Direct evidence (return-by-`const HandleType&` accessor):

```
proxy_base.h:118   isDecl              <- entry from TU A: getATypeNameUse = NO
proxy_base.h:118   isDecl,nameUseYES   <- entry from TU B: getATypeNameUse = YES  (same line!)
proxy_base.cpp:40  isDef               <- definition:     getATypeNameUse = NO
```

The query pairs the `YES` entry (`decl1`) with a `NO` entry (`decl2`) and
reports a divergence, even though both declarations spell the alias identically.

### Fix

Add three guards to the `where` clause, backed by three helper predicates:

1. **`sameSourceLocation(decl1, decl2)`** — two entries at the exact same file /
   line / column are the same source declaration seen from different TUs, not
   two redeclarations that could disagree. Exclude them.
2. **template-instantiation exclusion** — synthesised instantiation entries
   duplicate the template's entries without recording type-name uses.
3. **`typeAliasMentionedIn(t, decl2)`** — before reporting that `decl2` fails to
   use `t`, confirm it via `TypeMention`, which records *every* syntactic type
   mention with its location. Matching is done by **qualified name** so that a
   mention of a generic alias template (`Result`) is recognised as the same
   alias as the instantiated `getATypeNameUse()` result (`Result<X>`). For
   function definitions the mention search range is extended to the start of the
   function body to catch trailing return types (`auto f() -> T`).

### Validation

- **Real codebase** (`eclipse-score/communication`, `//score/message_passing` +
  `//score/mw/com`, whole-program CodeQL database, coding-standards pack
  `2.62.0` baseline): **276 → 30 findings (89% reduction)**. All 246 eliminated
  findings were manually verified as false positives (identical alias spelling
  confirmed by reading the source at both declaration sites); the 30 remaining
  findings are genuine spelling mismatches (e.g. a class member alias used in
  the header vs. the underlying namespace-scope alias spelled out explicitly in
  the out-of-line definition).
- **Upstream unit test:** `TypeAliasesDeclaration.qlref` still **PASSES** — the
  genuine `INT`/`Index`-vs-`int`/`i` divergences are still reported, so no true
  positives are lost.

### Known limitation

The residual ~30 findings (on our codebase) are return-by-`const Alias&`
accessors and some template members where **neither** `getATypeNameUse()` nor
`TypeMention` emits any mention of the alias on the relevant line, so no
reliable signal exists to suppress them without risking false negatives. These
are all genuine violations in our codebase, so this is expected, not a gap.

### Why no qltest regression test

The false positive is an **emergent property of real (Bazel/clang, one
compiler invocation per translation unit) whole-program extraction**: the
cross-TU `getATypeNameUse()` asymmetry does not arise in a single TU (the
relation is symmetric there), and it does **not** reproduce under `codeql
test run`'s single-pass extraction, even when deliberately mimicking the
multi-TU shape (shared header + 2/3 `.cpp` TUs, template instantiations,
trailing-return-type prototype/definition pairs, default-argument uses,
class-scope vs. fully-qualified alias spellings, `codeql database create
--command` with explicit separate `g++`/`clang++` invocations per TU — 10+
structural variations tried, none diverged). This is the same reason the
existing per-rule test never caught the issue in the first place.

What **does** reliably reproduce it is the project's own Bazel-traced
extraction, which is available and cheap to run:

```bash
# in a checkout of eclipse-score/communication
bazel run //quality/static_analysis:codeql_lint -- \
  --phase create-database --database-path /tmp/scoped_repro \
  --target //score/message_passing/log            # ~15s, 3 real TUs

codeql query run cpp/misra/src/rules/RULE-6-9-1/TypeAliasesDeclaration.ql \
  --database=/tmp/scoped_repro
# unfixed query:  1 finding  (score/message_passing/log/logging_callback.h:
#                 `LoggingCallback GetCerrLogger();` vs. its .cpp definition,
#                 identical alias spelling both sides)
# fixed query:    0 findings
```

The fix is therefore validated against this minimal, reproducible real-world
target (1 → 0) in addition to the full-scope numbers above (276 → 30), rather
than a synthetic `.expected` file, since the bug cannot be expressed in a
minimal synthetic TU under the qltest harness.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - RULE-6-9-1

## Release change checklist

  - [X] Yes
  - [ ] No

🚨🚨🚨
  - [ ] Confirmed

  - [ ] Confirmed

### Author

### Reviewer 

### DIFF
--- a/cpp/misra/src/rules/RULE-6-9-1/TypeAliasesDeclaration.ql
+++ b/cpp/misra/src/rules/RULE-6-9-1/TypeAliasesDeclaration.ql
@@ -17,14 +17,91 @@
 import cpp
 import codingstandards.cpp.misra
 
+/**
+ * Holds if `decl1` and `decl2` refer to the same source location (same file,
+ * line and column).
+ */
+predicate sameSourceLocation(DeclarationEntry decl1, DeclarationEntry decl2) {
+  decl1.getLocation().getFile() = decl2.getLocation().getFile() and
+  decl1.getLocation().getStartLine() = decl2.getLocation().getStartLine() and
+  decl1.getLocation().getStartColumn() = decl2.getLocation().getStartColumn()
+}
+
+/**
+ * Gets a line, in the file of declaration entry `decl`, that is part of the
+ * source range in which the type used for the declared entity appears.
+ *
+ * This is the range spanned by the declaration entry itself, extended - for a
+ * function definition - up to the start of the function body. The extension is
+ * required because a function definition may use a trailing return type
+ * (`auto f() -> T`), which appears after the function name and hence outside the
+ * declaration entry's own (name-based) location.
+ */
+predicate declTypeUseLine(DeclarationEntry decl, File file, int line) {
+  file = decl.getLocation().getFile() and
+  (
+    line in [decl.getLocation().getStartLine() .. decl.getLocation().getEndLine()]
+    or
+    exists(FunctionDeclarationEntry fde |
+      fde = decl and
+      fde.getBlock().getLocation().getFile() = file and
+      line in [fde.getLocation().getStartLine() .. fde.getBlock().getLocation().getStartLine()]
+    )
+  )
+}
+
+/**
+ * Holds if the type alias `t` is mentioned within the source lines spanned by
+ * the declaration entry `decl`.
+ *
+ * `TypedefType.getATypeNameUse()` is documented to return a conservative
+ * (incomplete) set of type name uses - in particular it omits uses on
+ * prototypes and around template instantiations. As a result it frequently
+ * fails to associate an alias with a redeclaration that genuinely uses it
+ * (e.g. a function prototype in a header whose definition lives in a `.cpp`),
+ * which produces false positives for this rule. `TypeMention` records every
+ * syntactic mention of a type together with its location, so we use it to
+ * confirm whether `decl` really does use the alias `t`.
+ */
+predicate typeAliasMentionedIn(TypedefType t, DeclarationEntry decl) {
+  exists(TypeMention tm |
+    // Match on the qualified name rather than object identity: a mention of a
+    // type alias template (e.g. `Result`) resolves to the generic alias, while
+    // `getATypeNameUse()` reports the instantiated alias (e.g. `Result<X>`).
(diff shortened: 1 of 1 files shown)


MichaelRFairhurst:
@castler I will take more time to take a look at this. I'm about to go on vacation for the next ~two weeks so I will review it when I'm back.

One note is that location-based querying:
```
predicate sameSourceLocation(DeclarationEntry decl1, DeclarationEntry decl2) {
  decl1.getLocation().getFile() = decl2.getLocation().getFile() and
  decl1.getLocation().getStartLine() = decl2.getLocation().getStartLine() and
  decl1.getLocation().getStartColumn() = decl2.getLocation().getStartColumn()
}
```

Can very often introduce bad joins (there are a lot of elements that share start lines, end lines, column offsets, and files). I'm going to run performance tests on it in the meantime, and I can help with any performance regressions that come up (optimizing QL is a little unintuitive vs other languages).

Thank you for all your submissions!

PR #1181, opened 2026-08-24
RULE-5-10-1: fix false positives for locals/parameters in permitted std specializations (e.g. std::hash)

castler:
## Description

`RULE-5-10-1` ("User-defined identifiers shall have an appropriate form") reports
`Identifier 'X' is defined in reserved namespace.` for ordinary, well-formed,
snake_case local variables and function parameters — e.g. `value`, `result`,
`view`, `identifier`, `instance_identifier`, `skeleton_instance_identifier` —
whenever they are declared inside the body of an explicit template
specialization that C++ explicitly permits users to add to namespace `std`,
most commonly `std::hash<UserType>::operator()`:

```cpp
namespace std
{
template <>
struct hash<score::mw::com::impl::ServiceIdentifierType>
{
    size_t operator()(const score::mw::com::impl::ServiceIdentifierType& value) const noexcept
    //                                                                     ^^^^^ flagged: "Identifier 'value' is defined in reserved namespace."
    {
        return std::hash<std::string_view>{}(value.ToHashString());
    }
};
}  // namespace std
```

`[namespace.std]`/`[hash.requirements]` explicitly permit a program to add a
template specialization for a standard library template (like `std::hash`) to
namespace `std`, provided the specialization depends on a user-defined type.
`value` here is not a new member of namespace `std` in the sense the rule is
trying to prevent — it is an ordinary parameter name scoped to the function
body, which merely happens to be lexically nested inside `namespace std { ... }`
because that is where the standard requires the specialization to be written.

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - RULE-5-10-1

## Release change checklist

  - [X] Yes
  - [ ] No

🚨🚨🚨
  - [x] Confirmed

  - [x] Confirmed

### Author

### Reviewer 

### DIFF
--- a/change_notes/2026-08-24-fix-fp-rule-5-10-1-hash-specialization-reserved-namespace.md
+++ b/change_notes/2026-08-24-fix-fp-rule-5-10-1-hash-specialization-reserved-namespace.md
@@ -1,0 +1,5 @@
+- `RULE-5-10-1` - `PoorlyFormedIdentifier.ql`:
+  - Fixed false positives where a local variable or function parameter was reported as
+    "defined in reserved namespace" merely because its enclosing function is the body of
+    an explicit template specialization that C++ permits users to add to namespace `std`
+    (for example, `std::hash<UserType>::operator()`'s parameter and local names).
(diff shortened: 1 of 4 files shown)


MichaelRFairhurst (approved):
Great find, thank you for the fix!

PR #1188, opened 2026-08-28
Improve CERT DCL30-C stack return analysis

### LINKED ISSUE #495: `DCL30-C`: Replace implementation with cpp/return-stack-allocated-memory
### Affected rules
 - `DCL30-C`

### Description

There have recently been improvements to the C++ standard query for returning stack allocated memory:
https://github.com/github/codeql/blob/4de19b3ec97d4379ef5988bda591d749e1cd0f99/cpp/ql/src/Likely%20Bugs/Memory%20Management/ReturnStackAllocatedMemory.ql#L7

We should adopt this new and improved query which improves both false positive and false negative rates.

JeewoongKim:
## Description

Fixes #495.

This updates the CERT C `DCL30-C` function-return query to use the stack-return analysis
from `cpp/return-stack-allocated-memory`.

The previous implementation used the legacy local data-flow API and missed some stack-derived
pointer returns. (e.g., `char a[2] ; return a + 1;`)

The changes include:
- switch the query to the IR-based `MustFlow` analysis
- handle field addresses and pointer offsets
- update the query to a path query
- add regression tests for pointer-offset and field-address returns

The function-return query now focuses on return values. Stack-address escapes through output
parameters are still covered by the separate `AppropriateStorageDurationsStackAdressEscape` query.

Tested locally:
- `codeql test run --show-extractor-output c/cert/test/rules/DCL30-C`
- `python scripts/validate-rule-package.py c/Declarations8`

## Change request type

 - [ ] Internal documentation
 - [ ] External documentation

 - [ ] No rules added
     - _rule number here_
     - DCL30-C

## Release change checklist

  - [x] Yes
  - [ ] No

### Author

### DIFF
--- a/c/cert/src/rules/DCL30-C/AppropriateStorageDurationsFunctionReturn.md
+++ b/c/cert/src/rules/DCL30-C/AppropriateStorageDurationsFunctionReturn.md
@@ -183,7 +183,7 @@ DCL30-C = Union( CWE-562, list) where list =
 
 ## Implementation notes
 
-The rule checks specifically for pointers to objects with automatic storage duration that are returned by functions or assigned to function output parameters.
+The rule checks specifically for pointers to objects with automatic storage duration that are returned by functions.
 
 ## References
 
--- a/c/cert/src/rules/DCL30-C/AppropriateStorageDurationsFunctionReturn.ql
+++ b/c/cert/src/rules/DCL30-C/AppropriateStorageDurationsFunctionReturn.ql
@@ -3,7 +3,7 @@
  * @name DCL30-C: Declare objects with appropriate storage durations
  * @description When pointers to local variables are returned by a function it can lead to referring
  *              to objects outside of their lifetime, which is undefined behaviour.
- * @kind problem
+ * @kind path-problem
  * @precision high
  * @problem.severity error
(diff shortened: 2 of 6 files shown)


MichaelRFairhurst:
Hi @JeewoongKim ! Thank you for sending!

My first thought is that I'd rather solve this while still using the C Objects API. It makes sense that `c + 1` is a subobject of `c`, and we have some subobject support already where that _could_ belong.

However, in general, this project would probably benefit from using the ir more often, so from that standpoint I really like your approach.

I'd like to triage the differences this shows in practice before deciding the next direction. I'm wondering if the `c + 1` example was the only motivating case or if there were other cases you wanted to expand coverage of.

Thank you again!

JeewoongKim:
Hi @MichaelRFairhurst,

The concrete false negative I verified was the `a + offset` case.

I also wanted to cover field-address returns such as `&s.x`, since the core 
`ReturnStackAllocatedMemory` query referenced by #495 handles both pointer offsets and field addresses.

So `a + offset` was the motivating false negative, while field-address returns were the additional
coverage I had in mind. The other added cases were mainly boundary checks for the IR-based implementation

